PR #1, opened 2020-02-26
Fix dependency on systemd

safl:
When libsystemd is not installed on the system, then the pkg-config
check would still be true when systemd itself is on the system. Also,
when the library is unavailable then the hard-coded "-lsystemd" would
cause the build to fail. The following changes fix that.

* configure: changed check from "systemd" to the library "libsystemd"

* examples/Makefile: Replaced hard-coded "-lsystemd" with "${LDFLAGS}"
  as defined by "config-host.mak" in examples/Makefile

* src/Makefile: added include of "config-host.mak", removed hard-coded
  "-lsystemd", "-lsystemd" will trickle in from LDFLAGS

Signed-off-by: Simon A. F. Lund <simon.lund@samsung.com>

## Test

On my laptop (DebianBuster), it:

* Now builds with and without libsystemd
* ./test/test ran without error both with and without libsystemd

### DIFF
--- a/configure
+++ b/configure
@@ -186,7 +186,7 @@ print_config "libuuid" "${libuuid}"
 ##########################################
 # check for SystemD
 systemd="no"
-pkg-config --exists systemd --atleast-version=232
+pkg-config --exists libsystemd --atleast-version=232
 if [ $? -eq 0 ]; then
   systemd="yes"
 fi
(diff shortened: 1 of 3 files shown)


keithbusch:
Thanks for the catch!

PR #2, opened 2021-03-03
ioctl: Fix NVME_LOG_CDW10_NUMDL_MASK

tbzatek:
Used in nvme_get_log(), caused nvme_get_log_error() failures.


tbzatek:
Haven't checked other related constants, this made calls to `nvme_get_log_error()` fail with higher `nr_entries` numbers.

PR #3, opened 2021-03-04
Docs strings updates

tbzatek:
Most strings are taken from the official NVMe specification document: https://nvmexpress.org/developers/nvme-specification/. Not sure what is the licensing and terms to reuse the strings.

Some annotation references like the `@enum nvme_xxx` may need fixing, as well as the hex constants annotations.


keithbusch:
Thank you!

PR #13, opened 2021-06-19
SWIG Python bindings

hreinecke:
Here's an attempt to implement python bindings via SWIG. To give an idea on how it's supposed to work I've added a 'discover_loop.py' script to the 'examples' directory.
And even that needs to be improved to really to a 'loop' over all discovery log entries and connect to them, with possibly being called recursively for discovery controllers. As I said, it's just to give an idea about the bindings.

The python interface definitely could do with proper documentation, and of course packaging is very rudimentary.


keithbusch:
I have only an introductory level experience with SWIG, so it's I'm not feeling qualified to evaluate those particular parts. I like what you have, and the parts I am more familiar with look good, so let's pull this in.

PR #15, opened 2021-07-08
Minor fixes for a couple of compilation failures

jk-ozlabs:
This series includes a couple of compilation failures on my platform; the first for a link-order issue, the second for building c++ objects with plain `cc`. 

I've taken the approach of minimal change here; we may not need the full `$(LIBS)` on the `.so` link, but that's what the current build rules specify. Happy to update this if necessary.


hreinecke (approved):
Looks good.

hreinecke (approved):
D'oh. Of course.

PR #21, opened 2021-07-23
Move python bindings to a separate directory (pynvme)

martin-belanger:
In this pull request we have:

- The Python bindings code (SWIG etc.) was moved from the `src` directory toa new `pynvme` directory and the Makefiles were adjusted accordingly. 
- A new README.md file was added to the `pynvme` directory to explain how to use the Python bindings.
- A bug was fixed in the SWIG file (`nvme.i`)


keithbusch:
Seems okay to me. I'll apply this through the command line so that we don't have the unnecessary merge commit.

PR #22, opened 2021-07-30
Add support for meson build system

martin-belanger:
Hi @keithbusch 

As we discussed earlier this week, this is the pull request that adds support for the meson build system. As you will see, I did not modify any of the existing Makefiles. This simply adds meson in parallel to the existing Makefiles. 

I added text to the README file to explain how to use meson to build libnvme. I also gave a lot of examples showing how one can change the build configuration from the command line w/o having to modify any files.

Please review at your convenience. If you have any questions, please don't hesitate to contact me. I can also give you a live demo over Zoom if you'd like.

Regards,

Martin Belanger
Technical Staff / Engineering Technologist
Dell Inc.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -11,7 +11,7 @@ all_targets += telemetry-listen display-tree display-columnar discover-loop
 
 all: $(all_targets)
 
-%: %.c
+%: %.c ../src/libnvme.a
 	$(QUIET_CC)$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $< -lnvme $(LIBS)
 
 clean:
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -70,14 +70,22 @@ nvme_host_t nvme_default_host(nvme_root_t r)
 static int nvme_scan_topology(struct nvme_root *r, nvme_scan_filter_t f)
 {
 	struct dirent **subsys;
-	int i, ret;
+	int i, num_subsys, ret;
 
-	ret = nvme_scan_subsystems(&subsys);
-	if (ret < 0)
-		return ret;
+	num_subsys = nvme_scan_subsystems(&subsys);
+	if (num_subsys < 0) {
+		nvme_msg(LOG_DEBUG, "failed to scan subsystems: %s\n",
+			 strerror(errno));
+		return num_subsys;
+	}
 
-	for (i = 0; i < ret; i++)
-		nvme_scan_subsystem(r, subsys[i]->d_name, f);
+	for (i = 0; i < num_subsys; i++) {
+		ret = nvme_scan_subsystem(r, subsys[i]->d_name, f);
+		if (ret < 0) {
+			nvme_msg(LOG_DEBUG, "failed to scan subsystem %s: %s\n",
+				 subsys[i]->d_name, strerror(errno));
+		}
+	}
 
 	nvme_free_dirents(subsys, i);
 	return 0;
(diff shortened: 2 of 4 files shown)


birkelund:
Hi @martin-belanger 

For the CCAN configurator - I know your pain.

I have another project where I use CCAN stuff, and the way I did this in meson is to use a custom target in `ccan/meson.build`:

```meson
subdir('tools/configurator')

ccan_config_h = custom_target('ccan_config_h',
  output: 'config.h',
  command: [ccan_configurator],
  capture: true,
  build_by_default: true,
)

ccan_sources = [
  'ccan/err/err.c',
  'ccan/list/list.c',
  'ccan/pr_log/pr_log.c',
  'ccan/str/str.c',
  'ccan/time/time.c',
]

ccan_lib = static_library('ccan',
  [ccan_config_h, ccan_sources],
  include_directories : ccan_inc,
)
```

In `ccan/tools/configurator/meson.build`:

```meson
ccan_configurator = executable('configurator',
  files('configurator.c'),
  c_args: ['-Wno-missing-field-initializers'],
)
```

In the top-level `meson.build` I have

```meson
ccan_inc = include_directories('ccan')
subdir('ccan')
```

Then, in `src/meson.build` I just reference this

```meson
my_lib = library('my_lib', [ccan_config_h, my_sources],
  include_directories: [my_lib_inc, ccan_inc],
  dependencies: ...,
  link_with: [ccan_lib],
)
```

birkelund:
Oh sorry, I missed that that is exactly how you do it - I only saw your attempt at replacing it... Nevermind the noise then :)

martin-belanger:
No problem @birkelund . As you saw, I tried to resolve the CCAN with an all-meson approach, but I gave up and simply used a custom target. I may come back to it later to see if I can eliminate the custom target completely. The thing is that we are not using all that the CCAN code is testing for. So it could be simplified.

meson.build
```diff
@@ -0,0 +1,216 @@
...
+project(
+    'libnvme',
+    'c', 'cpp',
+    version: '0.1',
+    license: 'LGPLv2+',
+    default_options: [
+        'buildtype=release',
+        'prefix=/usr',
+    ]
+)
```
verschwindibussen, line 70: Do we need a minimum version for meson here?

meson.build
```diff
@@ -0,0 +1,216 @@
...
+configure_file(
+    output: 'config-host.h',
+    configuration: conf
+)
+
+################################################################################
+NAME     = meson.project_name()
+SPECFILE = '@0@.spec'.format(NAME)
+
+substs = configuration_data()
```
verschwindibussen, line 189: Please use the pkgconfig module from meson.

meson_options.txt
```diff
@@ -0,0 +1,9 @@
+# -*- mode: meson -*-
+
+option('pkgconfiglibdir', type : 'string', value : '', description : 'directory for standard pkg-config files')
+
+option('systemd', type : 'combo', choices : ['auto', 'true', 'false'], description : 'libsystemd support')
```
verschwindibussen, line 5: Not completely sure about the combo type. Do we really want the auto?

pynvme/nvme.i
```diff
@@ -377,7 +377,7 @@ struct nvme_ns {
 
 %extend nvme_host {
   nvme_host(struct nvme_root *r, const char *hostnqn = NULL,
-	    const char *hostid = NULL) {
+            const char *hostid = NULL) {
```
verschwindibussen, line 380: Looks unrelated, please remove.

meson.build
```diff
@@ -0,0 +1,216 @@
...
+substs.set('NAME',       NAME)
+substs.set('VERSION',    run_command(awk, '/Version:/ { print $2 }', SPECFILE).stdout().strip())
```
verschwindibussen, line 191: Could we get rid of awk here. Reading this from the specfile looks a bit awkward.

verschwindibussen (commented):
Generally looks good. A few small cleanup suggestions.

meson_options.txt
martin-belanger: This is to automatically use these libraries if they are already installed. Otherwise, these features are automatically turned off.

verschwindibussen: We just discussed this here at ALPSS (we = Keith, Hannes and me). So we decided to make libuuid, json-c mandatory and get rig of the systemd dependency by implementing the uuid lookup directly in the library.

meson.build
verschwindibussen: Yes, we should avoid having two places defining the version. Hmm, not sure if we should remove the spec file.

PR #23, opened 2021-07-30
Add identify method to Python bindings

martin-belanger:
This adds the method `identify()` to the `nvme.ctrl` object in the `nvme.py` module.
The identify data is returned to the python caller as a dictionary.


martin-belanger:
Don't need this anymore

PR #25, opened 2021-08-06
Support for cross-compiling libnvme

jk-ozlabs:
[ this is more a review request than a pull request; I'm happy to rework in response to feedback. ]

This series adds a bunch of little cleanups to enable cross-compilation of libnvme. There are a couple of assumptions and hard-coded tools/paths that we can unify to standard-ish Makefile conventions.

The other option would be to go full autoconf+automake, but I assume there's reasons for not using those, so I'll stick with this approach for now. Let me know if you'd prefer otherwise.

This is based on my `fixes` branch - the first three patches are general bugfixes; those are covered by https://github.com/linux-nvme/libnvme/pull/15 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ examples/display-columnar
 examples/telemetry-listen
 examples/discover-loop
 
-ccan/config.h
 ccan/tools/configurator/configurator
 
 doc/_build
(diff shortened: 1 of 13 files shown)


martin-belanger:
Hi @jk-ozlabs. It's funny we had a similar idea. I already submitted a pull request (link below) to add support for the [meson build system](https://mesonbuild.com/).

https://github.com/linux-nvme/libnvme/pull/22

PR #30, opened 2021-08-19
Fix memory leaks and double free

martin-belanger:
Fixed double free (which causes things to crash) and fixed a few memory leaks.


hreinecke (commented):
Yep, looks good.

PR #34, opened 2021-09-15
nvme-cli: nvme gen-hostnqn use partition UUID on IBM POWER

wxiong38:
IBM POWER systems expose a platform created UUID in the device
tree. Use that as the UUID in the host NQN

Signed-Off-by: Wen Xiong <wenxiong@linux.vnet.ibm.com>
Reviewed-by: Brian King <brking@linux.vnet.ibm.com>

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -590,6 +590,27 @@ out_free_log:
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


keithbusch:
Looks fine, but I wonder how many platform specific paths we may end up with.

PR #35, opened 2021-09-17
CSI ignored upon ns create

MatiasBjorling:
The CSI value does not get passed ipon anmespace create after the switch to
libnvme.

Add the necessary hooks for nvme-cli.

Signed-off-by: Matias Bjørling <matias.bjorling@wdc.com>


MatiasBjorling:
Keith, would you like it as it is within the current patch set, or would you like me to modify the current functions to just take the csi directly?

keithbusch:
For some reason i'm getting a "404" page not found when I try to click your patch. Let me try to do it from the command line and see what this looks like. I'll be able to get to that later this afternoon.

MatiasBjorling:
Oh joy, heh. I can see them. Must be github or something.

PR #43, opened 2021-10-07
[meson-v2] - Add support for meson build system

martin-belanger:
@verschwindibussen , @hreinecke , @keithbusch 
I closed the [original meson pull request](https://github.com/linux-nvme/libnvme/pull/22) and I'm submitting this new one (meson-v2) instead. This new pull request addresses the following items that Daniel identified in the original pull request:
- Specify the minimum version required for meson (`meson_version`). I found that the minimum version of meson needed to build libnvme is 0.46.1 (released May 16th 2018). 
- Use `pkgconfig` meson module to generate the `libnvme.pc` file.
- Make `libuuid` and `json-c` mandatory.
- Remove dependency to `systemd`. Until the code is changed to no longer use systemd, I will leave systemd as an optional dependency in the `meson.build` file. I can remove it later once it's no longer required.
- Remove `nvme.i` from the pull request (never should have been there in the first place). 
- I no longer retrieve the meson project name and version from the spec file using `awk`. This removes the dependency on `awk`. Instead, the project name and version are defined in the top-level `meson.build` file and the spec file gets generated by meson. Note that this generated spec file specifies meson as the build system (i.e. `%meson`, `%meson_build`, etc.). See https://docs.fedoraproject.org/en-US/packaging-guidelines/Meson/ for details.


martin-belanger:
@jk-ozlabs , I need to check with @keithbusch whether the `ccan` configurator can be removed in favor of a pre-defined config.

igaw:
The header files are not correctly installed.

image

Another thing which popped up

image

jk-ozlabs:
> > Can I request that we skip the ccan configurator step, and provide a pre-defined config there? That'll make cross-compiling possible.
> 
> Sounds reasonable to me. My thinking was getting the normal meson workflow working first (including the mesonizing nvme-cli) and then figure out how to cleanly support cross-compiling. But don't let me stop you from sending PRs.

Super, thanks. I'm happy to send patches for this if no-one's already working on it.

PR #51, opened 2021-10-08
Rework scan logic and introduce libnvme versioning

hreinecke:
This patchset reworks the scanning logic to iterate over controllers, and not subsystems. Iterating over subsystems has the drawback that the host is only specified by controller attributes, so trying to enumerate controllers from subsystems may refer to the wrong host.

To accomodate for the change in behaviour this patchset also introduces libnvme symbol versioning, starting with version 1.0.1 and updating the version to 1.1.0 with this patchset.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -23,8 +23,8 @@ SED ?= sed
 INSTALL ?= install
 
 soname=$(NAME).so.1
-minor=0
-micro=2
+minor=1
+micro=0
 libname=$(soname).$(minor).$(micro)
 all_targets += $(NAME).a
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -24,7 +24,7 @@ INSTALL ?= install
 
 soname=$(NAME).so.1
 minor=0
-micro=1
+micro=2
 libname=$(soname).$(minor).$(micro)
 all_targets += $(NAME).a
 
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,4 +1,4 @@
-{
+LIBNVME_1.0 {
 	global:
 		__nvme_get_log_page;
 		__nvme_msg;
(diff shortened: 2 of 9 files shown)


src/Makefile
```diff
@@ -24,7 +24,7 @@ INSTALL ?= install
 
 soname=$(NAME).so.1
 minor=0
-micro=1
+micro=2
```
igaw: I don't think we have to update the version number yet, as we haven't released the shared library yet.

src/libnvme.map
```diff
@@ -7,7 +7,6 @@ LIBNVME_1.0 {
 		nvme_attach_ns;
 		nvme_compare;
 		nvme_copy;
-		nvme_create_ctrl;
```
igaw, line 10: Obviously, we need also the corresponding patch to nvme-cli

src/Makefile
igaw: I haven't any experience with shared library versioning. I suppose there is some documentation available how it supposed to be done.

src/libnvme.map
igaw: Well, this breaks building nvme-cli using the shared library. I think we should really align such changes.

hreinecke: Sadly I haven't figured out how to mark a function as 'not present' in newer shared library versions.
It's easy to add new functions, but one cannot easily remove them...

igaw: The only idea I have is to list all available symbols all the time, e.g. 

LIBVNME_1.0 {
  nvme_foo;
  nvme_bar;
}

LIBNVME_1.1 {
  nvme_foo;
}

igaw:
Let's close this PR as it is pretty outdated. I've opened a bug for the linker symbol versioning.

PR #55, opened 2021-10-11
fabrics: Do not overflow when reading systemd_uuid

igaw:
gcc complains with:

../src/nvme/fabrics.c:774:15: warning: ‘read’ writing 512 bytes into a region of size 37 overflows the destination [-Wstringop-overflow=]
  774 |         len = read(f, system_uuid, 512);
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~
../src/nvme/fabrics.c: In function ‘nvmf_hostnqn_generate’:
../src/nvme/fabrics.c:863:14: note: destination object ‘uuid_str’ of size 37
  863 |         char uuid_str[37]; /* e.g. 1b4e28ba-2fa1-11d2-883f-0016d3cca427 + \0 */
      |              ^~~~~~~~

Let's limit the read to 37 as we know that's the max size for a UUID.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -40,6 +40,7 @@
 #include "private.h"
 
 #define NVMF_HOSTID_SIZE	37
+#define UUID_SIZE		37  /* 1b4e28ba-2fa1-11d2-883f-0016d3cca427 + \0 */
 
 const char *nvmf_dev = "/dev/nvme-fabrics";
 const char *nvmf_hostnqn_file = "/etc/nvme/hostnqn";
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -771,7 +771,10 @@ static int uuid_from_device_tree(char *system_uuid)
 	f = open(PATH_UUID_IBM, O_RDONLY);
 	if (f < 0)
 		return -ENXIO;
-	len = read(f, system_uuid, 512);
+
+	/* uuid is max 37 chars long,
+	 * e.g. 1b4e28ba-2fa1-11d2-883f-0016d3cca427 + \0 */
+	len = read(f, system_uuid, 37);
```
hreinecke, line 777: We need to forcibly insert a terminating 'NULL' character after the read; we have no idea _what_ we read, so there is no guarantee that the string is properly terminated.

hreinecke (changes requested):
Otherwise looks good.

PR #57, opened 2021-10-11
Remove Makefiles and replace with a wrapper for meson

martin-belanger:
Hi @igaw, @hreinecke, @keithbusch,

This pull request removes the traditional `Makefiles` and replaces them with a single top-level `Makefile` that invokes `meson` to build `libnvme`. In other words, this is just a `meson` wrapper for people that are not familiar with `meson`.

The new `Makefile` supports the following:
```
make         - Build the code
make install - Build and install
make test    - Build and run tests
make clean   - Remove build artifacts but keep configuration
make purge   - Remove build artifacts and the configuration
```
Note that I removed the following targets: `cscope`, `tag-archive`, `create-archive`, `archive`, and `srpm`. I wasn't sure whether to keep them. Please let me know if I should add some or all of them back to the new `Makefile`.


igaw:
I think we are at the point where we could switch to meson. 
@martin-belanger  could you refresh this PR?

@hreinecke, @keithbusch are you okay with switching to meson now? Or do you want to keep the Makefile build system. In this case I would recommend to remove meson. Having two build systems is causing already a bit of troubles.

keithbusch:
Yes, I'm okay with purely meson build.

martin-belanger:
@igaw Yes, I can refresh this pull request. Will do it later today or tomorrow.

martin-belanger:
I'm closing this pull request and will submit a new one.

PR #58, opened 2021-10-11
[Python]: Add set_persistent() and C definitions

martin-belanger:
These are changes to the Python module only. 

1. The C definitions (`enum`, `#define`) are now available in the Python module.
2. The method `set_persistent()` is now available in the Python `nvme_ctrl` class.
3. The `persistent` property is also added to the `nvme_ctrl` class.


pynvme/nvme.i
```diff
@@ -18,6 +18,19 @@
...
+ * when we extend "struct nvme_ctrl" (see "%extend nvme_ctrl" below).
+ * Instead, we define the method as nvme_ctrl::__set_persistent__() to avoid the
+ * conflict. Fortunately, we can work around the problem with SWIG's %rename()
+ * command so that in the generated Python code the method will be named
+ * set_persistent() and not __set_persistent__() as shown here:
+ *
+ *   def set_persistent(self, persistent: "bool") -> "void":
+ *       return _nvme.ctrl_set_persistent(self, persistent)
+ */
+%rename(set_persistent)  __set_persistent__;
```
hreinecke, line 32: I'd rather change the name of the function, such that it doesn't clash with the SWIG generated names.
Easier, and one doesn't has to deal with the ugly rename construct.

pynvme/nvme.i
```diff
@@ -605,3 +623,15 @@ struct nvme_ns {
...
+// We want to swig all the #define and enum from types.h, but none of the structs.
+%{
+#include "nvme/types.h"
+%}
+#define __attribute__(x)
+%rename($ignore, %$isclass) "";     // ignore all classes/structs
+%rename($ignore, %$isfunction) "";  // ignore all functions
+%rename($ignore, %$isunion) "";     // ignore all unions
+%rename($ignore, %$isvariable ) ""; // ignore all variables
+
```
hreinecke, line 636: Whee. The things you can do with SWIG ...
Thanks for doing this.

hreinecke (changes requested):

pynvme/nvme.i
martin-belanger: Ok. I renamed the function to `persistent_set()`

PR #60, opened 2021-10-11
[Python]: Add init() method to class nvme_ctrl.

martin-belanger:
Make C function `nvme_init_ctrl()` available to Python as `nvme_ctrl.init()`. This allows Python programs to reuse an existing (persistent) nvme device.

### DIFF
--- a/src/nvme/tree.h
+++ b/src/nvme/tree.h
@@ -854,6 +854,16 @@ nvme_ctrl_t nvme_scan_ctrl(nvme_root_t r, const char *name);
 void nvme_rescan_ctrl(nvme_ctrl_t c);
 
 /**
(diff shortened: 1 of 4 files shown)


src/nvme/tree.h
```diff
@@ -59,7 +59,7 @@ typedef bool (*nvme_scan_filter_t)(nvme_subsystem_t);
  * nvme_first_host() -
  * @r:
  *
- * Return:
```
igaw, line 62: Could you do a seperate patch for the whitespace fixes? Kind of hides the nvme_init_ctrl() move. Thanks!

martin-belanger: Will do. I forgot that my editor was set to delete trailing spaces on save. ;-)

PR #74, opened 2021-10-19
meson, src/Makefile: fix ccan config.h

birkelund:
In the make-based setup, the `ccan/ccan-config.h` generated by the CCAN configurator is never read by the CCAN sources which expect `config.h`. Instead, because of `-I..` in `src/Makefile`, the CCAN sources picks up the base `config.h` (as generated by `configure`). All kinds of wrong since `config.h` should actually have been `config-host.h` (which is why it isn't cleaned up by `make clean` btw).

Fix the compiler args to not include the base directory and generate the CCAN config file as `ccan/config.h`. Rename `config.h` to `config-host.h`.

Secondly, for the meson setup, there is something fishy going on with CCAN as well. It works for `src/`, but if the examples or tests try to use CCAN, it throws up because the defines in the combined `config.h` doesn't seem to be picked up correctly. Reverting to the CCAN generator based setup (meson config already present in `ccan/meson.build`) fixes this.

### DIFF
--- a/ccan/meson.build
+++ b/ccan/meson.build
@@ -11,7 +11,7 @@ configurator = executable(
     c_args: ['-D_GNU_SOURCE'],
 )
 
-config_h = custom_target(
+ccan_config_h = custom_target(
     'config.h',
     output:  'config.h',
     capture: true,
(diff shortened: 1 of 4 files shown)


birkelund:
#77 much better :+1:

PR #75, opened 2021-10-20
Minor header definition fixes

jk-ozlabs:
These changes add a few small fixes for nvme headers.

The `private.h` header uses a couple of types that aren't included directly.

The `types.h` header defines a type with an unusual abbreviation of "heal" for "health", which is inconsistent with the subsystem health type. Although this is a public-API header, libnvme doesn't currently expose any function which uses that type.

As always, comments & queries most welcome!

### DIFF
--- a/src/nvme/private.h
+++ b/src/nvme/private.h
@@ -9,6 +9,15 @@
 #ifndef _LIBNVME_PRIVATE_H
 #define _LIBNVME_PRIVATE_H
 
+#include <ccan/list/list.h>
+
+#include "fabrics.h"
+
+#ifdef CONFIG_LIBUUID
+#include <uuid/uuid.h>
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
> > The `types.h` header defines a type with an unusual abbreviation of "heal" for "health", which is inconsistent with the subsystem health type. Although this is a public-API header, libnvme doesn't currently expose any function which uses that type.
> 
> Looks like a misspelling to me.

Yeah, me too. We could add a compat define if necessary:

```
    #define nvme_mi_ctrl_heal_status nvme_mi_ctrl_health_status
```

but it doesn't sound like anything is currently relying on the `_heal_` name anyway - there's no other MI implementation that I know of...

PR #76, opened 2021-10-20
Initial NVMe-MI protocol support

jk-ozlabs:
This series implements the NVMe Management Interface (NVMe-MI), allowing management of NVMe devices through an out-of-band MCTP channel, using the new `AF_MCTP` sockets support in Linux.

These introduce a separate shared library - `libnvme-mi.so`, as the library's consumers would typically be different from those using `libnvme.so`. I'm open to feedback on the best structure for integrating the MI layer though.

This series is on top of the `fixes` PR (#75).

With these changes, and an `AF_MCTP`-enabled kernel, I can interact with NVMe hardware over the out-of-band SMBus channel:

```
# examples/mi-mctp 1 9 info
NVMe MI subsys info:
 num ports: 2
 major ver: 1
 minor ver: 0
NVMe MI port info:
  port 0
    type PCIe[1]
    MCTP MTU: 0
    MEB size: 0
    PCIe max payload: 0x4000
    PCIe link speeds: 0x00
    PCIe current speed: 0x00
    PCIe max link width: 0x00
    PCIe neg link width: 0x00
    PCIe port: 0x00
  port 1
    type SMBus[2]
    MCTP MTU: 120
    MEB size: 0
    SMBus address: 0xa6
    VPD access freq: 0x03
    MCTP address: 0x3a
    MCTP access freq: 0x03
    NVMe basic management: enabled
NVMe MI subsys health:
 subsystem status:  0x30
 smart warnings:    0xff
 composite temp:    46
 drive life used:   1%
 controller status: 0x0001
```

Comments/queries/etc most welcome.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -32,3 +32,10 @@ executable(
     dependencies: libnvme_dep,
     include_directories: [incdir, internal_incdir]
 )
+
+executable(
+    'mi-mctp',
+    ['mi-mctp.c'],
+    dependencies: libnvme_mi_dep,
+    include_directories: [incdir, internal_incdir]
+)
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -115,8 +115,56 @@ int do_info(nvme_mi_ep_t ep)
 	return 0;
 }
 
+static int show_ctrl(nvme_mi_ep_t ep, uint16_t ctrl_id)
+{
+	struct nvme_mi_read_ctrl_info ctrl;
+	int rc;
+
+	rc = nvme_mi_mi_read_mi_data_ctrl(ep, ctrl_id, &ctrl);
+	if (rc)
+		return rc;
+
+	printf("  Controller id: %d\n", ctrl_id);
+	printf("    port id: %d\n", ctrl.portid);
+	if (ctrl.prii & 0x1) {
+		uint16_t bdfn = le16_to_cpu(ctrl.pri);
+		printf("    PCIe routing valid\n");
+		printf("    PCIe bus: 0x%02x\n", bdfn >> 8);
+		printf("    PCIe dev: 0x%02x\n", bdfn >> 3 & 0x1f);
+		printf("    PCIe fn : 0x%02x\n", bdfn & 0x7);
+	} else {
+		printf("    PCIe routing invalid\n");
+	}
+	printf("    PCI vendor: %04x\n", le16_to_cpu(ctrl.vid));
+	printf("    PCI device: %04x\n", le16_to_cpu(ctrl.did));
+	printf("    PCI subsys vendor: %04x\n", le16_to_cpu(ctrl.ssvid));
+	printf("    PCI subsys device: %04x\n", le16_to_cpu(ctrl.ssvid));
+
+	return 0;
+}
+
+static int do_controllers(nvme_mi_ep_t ep)
+{
+	struct nvme_ctrl_list ctrl_list;
+	int rc, i;
+
+	rc = nvme_mi_mi_read_mi_data_ctrl_list(ep, 0, &ctrl_list);
+	if (rc) {
+		warnx("Can't perform Controller List operation");
+		return rc;
+	}
+
+	printf("NVMe controller list:\n");
+	for (i = 0; i < le16_to_cpu(ctrl_list.num); i++) {
+		uint16_t id = le16_to_cpu(ctrl_list.identifier[i]);
+		show_ctrl(ep, id);
+	}
+	return 0;
+}
+
 enum action {
 	ACTION_INFO,
+	ACTION_CONTROLLERS,
 };
 
 int main(int argc, char **argv)
@@ -132,7 +180,8 @@ int main(int argc, char **argv)
 			"usage: %s <net> <eid> [action] [action args]\n",
 			argv[0]);
 		fprintf(stderr, "where action is:\n"
-			"  info\n");
+			"  info\n"
+			"  controllers\n");
 		return EXIT_FAILURE;
 	}
 
@@ -150,6 +199,8 @@ int main(int argc, char **argv)
 
 		if (!strcmp(action_str, "info")) {
 			action = ACTION_INFO;
+		} else if (!strcmp(action_str, "controllers")) {
+			action = ACTION_CONTROLLERS;
 		} else {
 			fprintf(stderr, "invalid action '%s'\n", action_str);
 			return EXIT_FAILURE;
@@ -167,6 +218,9 @@ int main(int argc, char **argv)
 	switch (action) {
 	case ACTION_INFO:
 		rc = do_info(ep);
+		break;
+	case ACTION_CONTROLLERS:
+		rc = do_controllers(ep);
 		break;
 	}
 
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/**
+ * This file is part of libnvme.
+ * Copyright (c) 2021 Code Construct Pty Ltd.
+ *
+ * Authors: Jeremy Kerr <jk@codeconstruct.com.au>
+ */
+
+/**
+ * mi-mctp: open a MI connection over MCTP, and query controller info
+ */
+
+#include <err.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <libnvme-mi.h>
+
+#include <ccan/array_size/array_size.h>
+#include <ccan/endian/endian.h>
+
+static void show_port_pcie(struct nvme_mi_read_port_info *port)
+{
+	printf("    PCIe max payload: 0x%x\n", 0x80 << port->pcie.mps);
+	printf("    PCIe link speeds: 0x%02x\n", port->pcie.sls);
+	printf("    PCIe current speed: 0x%02x\n", port->pcie.cls);
+	printf("    PCIe max link width: 0x%02x\n", port->pcie.mlw);
+	printf("    PCIe neg link width: 0x%02x\n", port->pcie.nlw);
+	printf("    PCIe port: 0x%02x\n", port->pcie.pn);
+}
+
+static void show_port_smbus(struct nvme_mi_read_port_info *port)
+{
+	printf("    SMBus address: 0x%02x\n", port->smb.vpd_addr);
+	printf("    VPD access freq: 0x%02x\n", port->smb.mvpd_freq);
+	printf("    MCTP address: 0x%02x\n", port->smb.mme_addr);
+	printf("    MCTP access freq: 0x%02x\n", port->smb.mme_freq);
+	printf("    NVMe basic management: %s\n",
+	       (port->smb.nvmebm & 0x1) ? "enabled" : "disabled");
+}
+
+static struct {
+	int typeid;
+	const char *name;
+	void (*fn)(struct nvme_mi_read_port_info *);
+} port_types[] = {
+	{ 0x00, "inactive", NULL },
+	{ 0x01, "PCIe", show_port_pcie },
+	{ 0x02, "SMBus", show_port_smbus },
+};
+
+static int show_port(nvme_mi_ep_t ep, int portid)
+{
+	void (*show_fn)(struct nvme_mi_read_port_info *);
+	struct nvme_mi_read_port_info port;
+	const char *typestr;
+	int rc;
+
+	rc = nvme_mi_mi_read_mi_data_port(ep, portid, &port);
+	if (rc)
+		return rc;
+
+	if (port.portt < ARRAY_SIZE(port_types)) {
+		show_fn = port_types[port.portt].fn;
+		typestr = port_types[port.portt].name;
+	} else {
+		show_fn = NULL;
+		typestr = "INVALID";
+	}
+
+	printf("  port %d\n", portid);
+	printf("    type %s[%d]\n", typestr, port.portt);
+	printf("    MCTP MTU: %d\n", port.mmctptus);
+	printf("    MEB size: %d\n", port.meb);
+
+	if (show_fn)
+		show_fn(&port);
+
+	return 0;
+}
+
+int main(int argc, char **argv)
+{
+	struct nvme_mi_nvm_ss_health_status ss_health;
+	struct nvme_mi_read_nvm_ss_info ss_info;
+	nvme_root_t root;
+	nvme_mi_ep_t ep;
+	uint8_t eid;
+	int net;
+	int rc;
+	int i;
+
+	if (argc != 3) {
+		fprintf(stderr, "usage: %s <net> <eid>\n", argv[0]);
+		return EXIT_FAILURE;
+	}
+
+	net = atoi(argv[1]);
+	eid = atoi(argv[2]) & 0xff;
+
+	root = nvme_mi_create_root(stderr, DEFAULT_LOGLEVEL);
+	if (!root)
+		err(EXIT_FAILURE, "can't create NVMe root");
+
+	ep = nvme_mi_open_mctp(root, net, eid);
+	if (!ep)
+		err(EXIT_FAILURE, "can't open MCTP endpoint %d:%d", net, eid);
+
+	rc = nvme_mi_mi_read_mi_data_subsys(ep, &ss_info);
+	if (rc)
+		err(EXIT_FAILURE, "can't perform Read MI Data operation");
+
+	printf("NVMe MI subsys info:\n");
+	printf(" num ports: %d\n", ss_info.nump + 1);
+	printf(" major ver: %d\n", ss_info.mjr);
+	printf(" minor ver: %d\n", ss_info.mnr);
+
+	printf("NVMe MI port info:\n");
+	for (i = 0; i <= ss_info.nump; i++)
+		show_port(ep, i);
+
+	rc = nvme_mi_mi_subsystem_health_status_poll(ep, true, &ss_health);
+	if (rc)
+		err(EXIT_FAILURE, "can't perform Health Status Poll operation");
+
+	printf("NVMe MI subsys health:\n");
+	printf(" subsystem status:  0x%x\n", ss_health.nss);
+	printf(" smart warnings:    0x%x\n", ss_health.sw);
+	printf(" composite temp:    %d\n", ss_health.ctemp);
+	printf(" drive life used:   %d%%\n", ss_health.pdlu);
+	printf(" controller status: 0x%04x\n", le16_to_cpu(ss_health.pdlu));
+
+	nvme_mi_close(ep);
+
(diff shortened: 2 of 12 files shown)


birkelund:
Ah. I was searching lore for it, but couldn't find anything. Thanks, I'm super keep on trying it out :)

I like the separation by having this as a separate library and I think it helps my case for splitting up libnvme into a spec core library and "interface" libraries (i.e. `libnvme-ioctl`, `libnvme-vfio`, `libnvme-mctp`).

examples/meson.build
```diff
@@ -23,3 +23,8 @@ executable(
     link_with: libnvme,
     include_directories: incdir)
 
+executable(
+    'mi-mctp',
+    ['mi-mctp.c'],
+    link_with: libnvme_mi_static,
```
igaw, line 29: This looks wrong. I think you should just use libnvme_mi here, the user should be able to choose how the library is build/used.

src/libnvme-mi.map
```diff
@@ -0,0 +1,11 @@
+{
+	global:
+		nvme_mi_init_ctrl;
+		nvme_mi_close;
+		nvme_mi_mi_read_mi_data_subsys;
+		nvme_mi_mi_read_mi_data_port;
```
igaw, line 6: The naming scheme might be consistent. As German speaker I have to lough just heargin mimi :)
I leave these details to Keith and Hannes though :)

src/meson.build
```diff
@@ -56,6 +68,38 @@ libnvme_dep = declare_dependency(
...
+    version: libnvme_version,
+    soversion: '1',
+    link_args: ['-Wl,--version-script=' + mi_version_script_arg],
+    dependencies: mi_deps,
+    link_depends: mi_mapfile,
+    include_directories: incdir,
+    install: true,
+)
+
+libnvme_mi_static = static_library(
```
igaw, line 83: Please adopt the same strategy how to build libraries as we have in mainline. Klaus fixed up my mess I did with the meson.build file. See commit:  c874f6697578 ("meson: use library() and rely on user settings")

src/nvme/mi-mctp.c
```diff
@@ -0,0 +1,212 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2020 Code Construct Pty Ltd
```
igaw, line 4: I suppose you want to set the Copyright date to 2021.

examples/mi-mctp.c
```diff
@@ -79,31 +80,17 @@ static int show_port(nvme_mi_ep_t ep, int portid)
...
-	if (!ep)
-		err(EXIT_FAILURE, "can't open MCTP endpoint %d:%d", net, eid);
+	int i, rc;
 
 	rc = nvme_mi_mi_read_mi_data_subsys(ep, &ss_info);
-	if (rc)
-		err(EXIT_FAILURE, "can't perform Read MI Data operation");
+	if (rc) {
+		warn("can't perform Read MI Data operation");
+		return -1;
```
igaw, line 95: return EXIT_FAILURE

I just mentioning it once, there are couple of more places later in the series which do the same.

igaw (changes requested):
I don't see any technical reason not to merge this. But I'd like to hear some feedback from Keith or Hannes. Furthermore, I know Klaus is waiting to get this merged for his libnvme-linux PR :)

birkelund:
> I don't see any technical reason not to merge this. But I'd like to hear some feedback from Keith or Hannes.

Thumbs up for this as well. I have interest in NVMe MI as well and will be following this closely and helping out :)

jk-ozlabs:
Thanks for the updates folks! I have some small changes planned, and will also coordinate with the review above.

Just on my side: I've since added more on the Admin Channel functionality. However, I'd rather *this* initial PR be the core support (along with some minimal initial command support, as a PoC). The further command support I'll push as a separate branch, and we can look at that as a separate PR later.

I'll update this PR with a new push soon.

tasleson:
Unless I'm mistaken, the NVMe base specification for 2.0 mentions management interface with fabric/TCP for transports in addition to those for SMBus/I2C/PCIe.  Looking at this PR, I'm curious how that functionality would be integrated?  I suppose new function(s) to create the additional transport specific endpoint(s) would be the main requirement?

jk-ozlabs:
> I suppose new function(s) to create the additional transport specific endpoint(s) would be the main requirement?

Yep, exactly. It's my intention that future transports would still provide the same `nvme_mi_ep_t` API, and just have a transport-specific constructor (say, `nvme_mi_open_tcp`).

This would mean that the external MI API is all transport-agnostic - aside from the constructor, of course!

In the current internal API, we separate the transport from the endpoint:

```c
struct nvme_mi_transport {
	const char *name;
	bool mic_enabled;
	int (*submit)(struct nvme_mi_ep *ep,
		      struct nvme_mi_req *req,
		      struct nvme_mi_resp *resp);
	void (*close)(struct nvme_mi_ep *ep);
};

struct nvme_mi_ep {
	const struct nvme_mi_transport *transport;
	void *transport_data;
};
```
Where the `struct nvme_mi_transport` is provided by each possible transport (mctp, tcp, pci, etc), and in this case the `->submit` op allows generic MI code to perform the transport-specific transaction. This is all internal API though, so we can adjust to suit future transport implementations.

examples/meson.build
jk-ozlabs: Ack, will fix.

examples/mi-mctp.c
jk-ozlabs: Even though this isn't an exit status?

src/meson.build
jk-ozlabs: Ah, I'd based this on the earlier meson defs, which had both. I'll make this consistent with the current `meson.build`

src/nvme/mi-mctp.c
jk-ozlabs: 👍

examples/mi-mctp.c
igaw: You are right. Although returning a ESOMETING would be better IMO in this particular case.

src/libnvme-mi.map
jk-ozlabs: Yeah, this is for a structured namespace, really. The base "prefix" here is `nvme_mi`, the MI commands are given an additional `_mi`, the admin functions use `_admin`.

If there's a better naming scheme, I'm ok to work with that.

examples/mi-mctp.c
jk-ozlabs: ok, I'll figure out an appropriate value for these.

src/libnvme-mi.map
igaw: Let's keep it consistent and I can giggle everytime I read it :)

jk-ozlabs:
`mi` branch updated from review here, plus a couple of other fixes. Let me know if you have more comments!

src/libnvme-mi.map
jk-ozlabs: ok, I'm happy to provide both consistency and amusement. :laughing:

src/libnvme-mi.map
```diff
@@ -0,0 +1,11 @@
+{
```
igaw, line 1: Probably a good idea would be to add the symbol version from day 1 on. But this is something we can do after a merge.

src/nvme/mi-mctp.c
```diff
@@ -0,0 +1,212 @@
...
+#include <sys/types.h>
+#include <sys/uio.h>
+
+#include <ccan/endian/endian.h>
+
+#include "private.h"
+#include "log.h"
+#include "mi.h"
+
+#if 1
```
igaw, line 24: Left over from debugging?

src/nvme/private.h
```diff
@@ -131,4 +132,45 @@ void json_read_config(nvme_root_t r, const char *config_file);
...
+
+struct nvme_mi_resp {
+	struct nvme_mi_msg_hdr *hdr;
+	size_t hdr_len;
+	void *data;
+	size_t data_len;
+	__u32 mic;
+};
+
+
```
igaw, line 154: One new line too much

igaw (commented):
I looked over how the library is added, build integration. Looks reasonable to me.

Hannes told me he wants to dig into the spec first to before giving feedback.

src/nvme/mi-mctp.c
jk-ozlabs: Not really - more a marker for where we make this conditional on the new MCTP definitions being present in the usual toolchain headers.

Once we get things like `AF_MCTP`, `struct sockaddr_mtctp` etc present in those, we'll add an actual condition here.

If you prefer though, I can put that conditional in now - we're at a stage where we know what defs may be present in the kernel headers from actual releases.

src/libnvme-mi.map
jk-ozlabs: OK, happy to take your guidance on this.

src/nvme/private.h
jk-ozlabs: ack, will remove.

src/nvme/mi-mctp.c
igaw: I just stumbled over it when quickly looking through the code. It looks odd without an explanation. I'd say if you can
add a proper conditional now, go for it :)

jk-ozlabs: OK, can do!

jk-ozlabs:
Updated from @igaw's comments; we now check for a `linux/mctp.h`, and conditionally use definitions there if available.

jk-ozlabs:
Are there still change requests pending for this? It all seems to be resolved, but let me know if not.

igaw:
We are working on getting the initial release in shape. This is something for v1.1.

jk-ozlabs:
OK, no problem - just making sure you're not waiting on anything on my end. Keep me posted if that changes though!

examples/mi-mctp.c
```diff
@@ -162,9 +164,66 @@ static int do_controllers(nvme_mi_ep_t ep)
...
+
+	tmp = atoi(argv[1]);
+	if (tmp < 0 || tmp > 0xffff) {
+		fprintf(stderr, "invalid controller ID\n");
+		return -1;
+	}
+
+	ctrl_id = tmp & 0xffff;
+
+	ctrl = nvme_mi_init_ctrl(ep, tmp);
```
drakedog2008, line 199: possible memleak. ctrl is not free() at last

src/nvme/mi.h
```diff
@@ -0,0 +1,96 @@
...
+
+/* An endpoint may expose multiple controllers */
+struct nvme_mi_ctrl;
+typedef struct nvme_mi_ctrl *nvme_mi_ctrl_t;
+
+/* Transport-specific endpoint initialisation. Once an endpoint is created,
+ * the rest of the API is transport-independent. */
+nvme_mi_ep_t nvme_mi_open_mctp(unsigned int netid, uint8_t eid);
+
+nvme_mi_ctrl_t nvme_mi_init_ctrl(nvme_mi_ep_t ep, __u16 ctrl_id);
```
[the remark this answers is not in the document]
jk-ozlabs: Good catch - I'll add suitable destructor functions for controller and endpoints.

jk-ozlabs:
OK! I've updated for the API changes from the 1.1 release, which have resulted in some cascading changes for the libnvme-mi API.

Since the `nvme_root_t` is now fairly core to the API (eg, `nvme_msg()` requires a `nvme_root_t`), we now have an analogue in the MI code: `nvme_mi_create_root()`, which performs almost the same as `nvme_create_root()`, but prevents clashes with the latter. Like the base libnvme API, the root must be created before calling other API functions.

This does bring up a fairly important design point though: we could either have two separate `_create_root` functions (which currently do the same thing), or we could look at unifying those. If we do unify, we'd need to work out what should happen with applications that want to link to both libnvme *and* libnvme-mi, as a duplicated `_create_root` function will cause a namespace collision. One alternative would be to integrate all of the MI functions into the base libnvme, and drop the concept of a separate `libnvme-mi` library.

[I do like having a separate libnvme-mi, but I'm not too attached to that approach]

examples/mi-mctp.c
jk-ozlabs: This is now handled through `nvme_mi_close_ctrl`.

jk-ozlabs:
@iagw: I've just clicked "re-request review", as you had some comments that I think we've addressed. Could you let me know if those are still pending?

igaw:
Will do when I am back from my holiday in a week.

jk-ozlabs:
Just pushed a small update to deal with the new `libuuid` dependency semantics.

igaw:
Sorry for the delay on the review. I'll look into it next week.

src/nvme/mi.c
```diff
@@ -0,0 +1,260 @@
...
+#include <ccan/endian/endian.h>
+
+#include "log.h"
+#include "mi.h"
+#include "private.h"
+
+/* MI-equivalent of nvme_create_root, but avoids clashing symbol names
+ * when linking against both libnvme and libnvme-mi.
+ */
+nvme_root_t nvme_mi_create_root(FILE *fp, int log_level)
```
igaw, line 22: The question is if we should try to move common infrastructure and generic code out of `libnvme`? If I understood it correctly libnvme-mi can be used completely standalone too? I'd like to avoid having these function being duplicated. IIRC, there was already interests in splitting libnvme up into a generic part and a Linux specific one.

src/nvme/mi.c
```diff
@@ -0,0 +1,260 @@
...
+	struct nvme_mi_ctrl *ctrl;
+
+	ctrl = malloc(sizeof(*ctrl));
+	if (!ctrl)
+		return NULL;
+
+	ctrl->ep = ep;
+	ctrl->id = ctrl_id;
+
+
```
igaw, line 63: one empty line too much

igaw (commented):
This all looks in good shape. Just a couple of nitpicks and the question if we should split up libnvme into a generic part. Thoughs?

And is this supposed to happen:
```bash
./test-mi           
crc mismatch
```

src/libnvme-mi.map
```diff
@@ -0,0 +1,14 @@
...
+	global:
```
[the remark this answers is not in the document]
jk-ozlabs: Ack, will fix.

src/meson.build
```diff
@@ -60,6 +73,25 @@ libnvme_dep = declare_dependency(
...
+libnvme_mi = library(
+    'nvme-mi', # produces libnvme-mi.so
+    mi_sources,
+    version: library_version,
+    link_args: ['-Wl,--version-script=' + mi_version_script_arg],
+    dependencies: mi_deps,
+    link_depends: mi_mapfile,
+    include_directories: [incdir, internal_incdir],
+    install: true,
+)
```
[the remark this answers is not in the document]
jk-ozlabs: Ack, will fix.

igaw:
>The question if we should split up libnvme into a generic part. Thoughs?

libnvme-mi also depens on Linux's `AF_MCTP` socket interface. If we split libnvme into a generic part and a Linux specific, should the there also be a libnvme-mi-linux part? Just brainstorming here. Not sure if this practical or desirable at all.

src/nvme/mi.c
jk-ozlabs: I'm OK with a small amount of duplication here; the additional complexity of splitting libnvme doesn't seem to be a net win for this kind of thing. Having the MI interfaces as a "side-band" to the main interface seems OK for now.

Yes, libnvme-mi can be use completely standalone, without any local NVMe devices. The dependence on the kernel is just to create an `AF_MCTP` socket, but that could also be abstracted too (by replacing the MCTP backend with something else, as mentioned in https://github.com/linux-nvme/libnvme/pull/76#issuecomment-977326083 )

I do see one opportunity for unification though: we could be to make libnvme *completely* backend-independent; this would mean that the MI and ioctl-direct interfaces would use the same API, just a different constructor for the `nvme_root_t` - backed either by ioctls or MCTP (or something else for fabrics?); However, this would be a lot of churn, as there's lot of places in the current libnvme code that assume that everything is behind a file descriptor.  Overall, I don't think that's worth doing, particularly at this stage - I think we should give the MI interface some exposure to real applications to mature first.

Good point though - I'd definitely like to revisit this later, regardless of the decision now!

jk-ozlabs:
> If we split libnvme into a generic part and a Linux specific, should there also be a libnvme-mi-linux part?

I might have somewhat covered this above, but we can sort-of abstract this with the configurable transport for the MI backend - the MCTP-specific bits are all in mi-mctp.c, which provides the method of actually sending and receiving the MI messages.

However, I'm a little averse to creating too much of an abstraction when we only have one implementation to work with at present.

jk-ozlabs:
@igaw thanks for the review, I appreciate it! I'll do some fix ups from your comments, and push a new series.

> And is this supposed to happen:
> 
> ```shell
> ./test-mi           
> crc mismatch
> ```

One of the tests creates a message with an invalid crc, so that message is expected, as a warning form the libnvme-mi code. As long as the exit status of the process is OK:

```shell
[jk@rico libnvme]$ ./obj/test/test-mi; echo $?
crc mismatch
0
```
\- but I can always remove the warning if you prefer.

jk-ozlabs:
New series pushed, with updates as suggested by @igaw

src/nvme/mi.c
igaw: I hear you. I also don't want to over engineering it. If it is just these few function you need for `libnvme-mi` than we might just go along with it, especially if it is a standalone library. Let's see if someone else has some thoughts on this.

igaw:
> One of the tests creates a message with an invalid crc, so that message is expected, as a warning form the libnvme-mi code. As long as the exit status of the process is OK:

Ah, maybe add a 'test passed' or 'test failed' and we could hook it up into our testing CI steps.

igaw:
> However, I'm a little averse to creating too much of an abstraction when we only have one implementation to work with at present.

Yeah, maybe should just go with it and see how things turn out and plan for a libnvme 2.0 in year or two for a big refactoring.

igaw:
@hreinecke @keithbusch I plan to merge this soon. Please speak up if don't agree.

jk-ozlabs:
Added https://github.com/linux-nvme/libnvme/pull/76/commits/f87fa1c93af12234a1cb86003a1135b0022e7c9c to improve the test reporting, we now get:

```
[jk@pecola libnvme]$ ./obj/test/test-mi
Running test read_mi_data...  OK
Running test transport_fail...  OK
Running test invalid_crc...  OK
--- begin test output
crc mismatch
--- end test output
```

jk-ozlabs:
Awesome, thank *you* for the review! I've got a few follow-ups now for extending the feature sets, will get those posted soon.

igaw:
No problemo. I've suspected there is more to come :)

PR #77, opened 2021-10-21
meson + ccan configuration fixes

jk-ozlabs:
These changes fix a couple of issues with the ccan configuration system,
reported by @birkelund .

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -142,7 +142,7 @@ conf.set10(
     description: 'Is byteswap.h include-able?'
 )
 conf.set10(
-    'HAVE_BSWAP64',
+    'HAVE_BSWAP_64',
     cc.links(
         '''#include <byteswap.h>
             int main(void) {


birkelund:
Right, it was just a misspelled named in meson, I missed that.

Thanks!, LGTM!

igaw:
Oh, I screwed it up. When I merged PR#8 I didn't see it was based on PR#74. So PR#74 was not supposed to go in without reviewing. Sorry about that.

@jk-ozlabs what's the way forward? Should I revert PR#74 and apply this PR on top?

jk-ozlabs:
@iagw: working on a fix now - I'll see if I can get the revert done without breaking the build in-between. Will update this PR when I have something workable :)

jk-ozlabs:
OK, this has the reverts integrated with the fixes so we don't break the build in-between. The changes are no longer obvious though.

If you'd prefer the non-atomic series, which is a bit cleaner, I've also pushed that in my [ccan-fix-breaky](https://github.com/jk-ozlabs/libnvme/tree/ccan-fix-breaky) branch.

igaw:
BTW, with the recent changes how deal with ccan and the header cleanup (removing le32_to_cpu) we broke nvme-cli. I think we need to add ccan to nvme-cli as well. Care to send a PR for it, since you seem to have a way better understanding in these things than I do :)

jk-ozlabs:
Yep, I can definitely have a go at fixing that!

PR #88, opened 2021-10-29
Fix uuid_from_dmi() to work on Debian

martin-belanger:
The system UUID can be read from the `sysfs` in two different locations:

```
1) /sys/class/dmi/id/product_uuid
2) /sys/firmware/dmi/entries
```
Location 1) is present on all the systems I've tested: SUSE, Red Hat, 
Fedora, and Ubuntu. Unfortunately, location 2) does not exist on 
Debian-based systems.

Currently, `uuid_from_dmi()` only reads from location 2), which fails on Debian.

This patch adds code that reads from location 1) first, and will
only try location 2) if reading from 1) fails.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -781,7 +781,7 @@ static int uuid_from_device_tree(char *system_uuid)
 
 #define PATH_DMI_ENTRIES       "/sys/firmware/dmi/entries"
 
-static int uuid_from_dmi(char *system_uuid)
+static int uuid_from_dmi_entries(char *system_uuid)
 {
 	int f;
 	DIR *d;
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -803,7 +803,6 @@ static int uuid_from_dmi(char *system_uuid)
 		if (f < 0)
 			continue;
 		len = read(f, buf, 512);
-		len = read(f, buf, 512);
```
igaw, line 806: Strictly speaking this should be a commit of its own.

src/nvme/fabrics.c
```diff
@@ -838,6 +837,64 @@ static int uuid_from_dmi(char *system_uuid)
...
+
+			memcpy(system_uuid, line, UUID_SIZE);
+			system_uuid[UUID_SIZE-1]= '\0'; /* \n -> \0 */
+
+			/* Swap little-endian to network order per
+			   DMTF SMBIOS 3.0 Section 7.2.1 System UUID */
+			for (i = 0; swaptbl[i].from != -1; i++) {
+				tmp = system_uuid[swaptbl[i].from];
+				system_uuid[swaptbl[i].from] = system_uuid[swaptbl[i].to];
+				system_uuid[swaptbl[i].to] = tmp;
```
igaw, line 869: Nice! Could you make uuid_from_dmi_entries() also using this swap code? Maybe have it as small helper function?

igaw (changes requested):
Only nitpicks.

src/nvme/fabrics.c
martin-belanger: Will do

src/nvme/fabrics.c
martin-belanger: Ok, makes sense.

martin-belanger:
This is getting messy. Let me close and submit a fresh pull request.

PR #89, opened 2021-11-02
Remove a duplicate read() statement

martin-belanger:
Remove a duplicate read() statement. 

I checked with Hannes and he confirmed that the line was added by mistake.


igaw:
Looks good, but your commit message title needs a new line. Could you please update it? Thanks

martin-belanger:
Done

PR #92, opened 2021-11-02
Fix reading of system UUID on Debian

martin-belanger:
Fix reading of system UUID on Debian.

The system UUID can be read from the sysfs in two different locations:

```
1) /sys/class/dmi/id/product_uuid
2) /sys/firmware/dmi/entries
```

Location 1) is present on all the systems I've tested: SUSE, Red Hat,
Fedora, and Ubuntu. Unfortunately, location 2) does not exist on
Debian-based systems.

Currently, uuid_from_dmi() only reads from location 2), which fails
on Debian.

This patch adds code that reads from location 1) first, and will
only try location 2) if reading from 1) fails.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -781,7 +781,7 @@ static int uuid_from_device_tree(char *system_uuid)
 
 #define PATH_DMI_ENTRIES       "/sys/firmware/dmi/entries"
 
-static int uuid_from_dmi(char *system_uuid)
+static int uuid_from_dmi_entries(char *system_uuid)
 {
 	int f;
 	DIR *d;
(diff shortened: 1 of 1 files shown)


martin-belanger:
@igaw - I wanted to do what you suggested and move the conversion from little-endian into a separate "swap" function that both `uuid_from_dmi_entries()` and `uuid_from_product_uuid()` could use. However, the data handled by `uuid_from_dmi_entries()` is a binary `uuid` (16-byte array), whereas the data handled by `uuid_from_product_uuid()` is a string `uuid` (36-char + `NUL`). So it's not possible for both to share the same "swap" function.

I did simplify the little-endian swap algorithm a bit compared to the previous pull request.

igaw:
Okay, thanks for checking the details. Anyway looks good, but here are a few trailing spaces in the code snipped. Could fix those whitespace damages please? Thanks!

martin-belanger:
Trailing spaces removed

igaw:
I've applied the patch directly after changing the return value from -1 to -ENXIO to match with the other dmi function.

PR #93, opened 2021-11-04
Docs updates + minor fixes

tbzatek:
Mostly doc strings update, found typo in a constant name in between. Would be nice to run spell checker or something before the first public release.

Also removed the _'v'_ prefix from the library version number to align with the usual `pkg-config` file versioning practices.


igaw:
> Would be nice to run spell checker or something before the first public release.

Totally agree. If you have time to work on it please send the spelling fixes.

PR #94, opened 2021-11-05
Merge nvme cli monolithic

igaw:
This is a first rough version attempt to merge the missing patches from the nvme-cli-monolithic branch. Please see
also my notes in my PR in nvme-cli https://github.com/linux-nvme/nvme-cli/pull/1223

Ah yes, most stuff is only compile tested...


igaw:
@keithbusch  and @hreinecke please double check if the naming convention I used is okay. I had to adapt from the
original patch. So it's likely off in places.

keithbusch:
Thanks for doing this merge! Looks okay at first glance. I am traveling this week, though, I'll have time to test things out on Monday.

hreinecke (approved):
Looks good.

hreinecke (commented):
Hmm. This will change the ABI, and will cause nvme-cli to not build against this version unless patched, too.
Are we okay with that? Or do we need to provide compability wrapper?

igaw:
There was no official libnvme release yet, no? I think we can still argue that his doesn't need any backwards compatibility  (yet)...

src/libnvme.map
```diff
@@ -114,6 +114,7 @@
 		nvme_get_log_endurance_group;
 		nvme_get_log_endurance_grp_evt;
 		nvme_get_log_error;
+		nvme_get_log_fid_sup_efcs;
 		nvme_get_log_fw_slot;
```
hreinecke, line 118: Please make that 'get_log_fid_sup_effects'. 'efcs' is making no sense, and is not something one could figure out as an abbreviation.

src/nvme/ioctl.h
```diff
@@ -1004,6 +1004,18 @@ int nvme_get_log_lba_status(int fd, bool rae, __u64 offset, __u32 len,
 int nvme_get_log_endurance_grp_evt(int fd, bool rae, __u32 offset, __u32 len,
 				   void *log);
 
+/**
+ * nvme_get_log_fid_sup_efcs() -
+ * @fd:		File descriptor of nvme device
```
hreinecke, line 1032: Same here.

hreinecke (changes requested):

src/nvme/ioctl.c
```diff
@@ -1927,8 +1927,8 @@ int nvme_resv_report(int fd, __u32 nsid, bool eds, __u32 len,
 }
 
 int nvme_zns_mgmt_send(int fd, __u32 nsid, __u64 slba, bool select_all,
-		       enum nvme_zns_send_action zsa, __u32 data_len,
-		       void *data)
+		       __u32 timeout, enum nvme_zns_send_action zsa,
+		       __u32 data_len, void *data)
```
hreinecke, line 1963: Odd. Why do have zns command a timeout, and other commands haven't?
I'm pretty sure we can set a timeout on any command, so what makes these special?

src/nvme/types.h
```diff
@@ -2785,7 +2787,9 @@ struct nvme_telemetry_log {
 	__le16	dalb1;
 	__le16	dalb2;
 	__le16	dalb3;
-	__u8	rsvd14[368];
+	__u8    rsvd14[2];
+	__le32  dalb4;
+	__u8    rsvd20[362];
 	__u8	ctrlavail;
```
hreinecke, line 2809: All good, but where's the plugin?
Shouldn't we modify the commit message to refer to the changes we actually did?

hreinecke (changes requested):

src/nvme/ioctl.c
igaw: I don't know, this is from  fa97500b9a72 ("Add Timeout Parameter for Zone Management Commands")
https://github.com/linux-nvme/nvme-cli/commit/fa97500b9a72

hreinecke:
Hmm. Okay, just wanted to point out. Which means it's just the one commit message which needs to be modified. Rest is okay for merging.

src/libnvme.map
igaw: I was also unhappy with it, especially after I realized that I spelled it NVME_LOG_LID_FID_SUPPORT_EFFECTS for the define.

src/nvme/ioctl.h
igaw: oki

src/nvme/types.h
igaw: The plugin part is in the nvme-cli PR (https://github.com/linux-nvme/nvme-cli/pull/1223):

https://github.com/linux-nvme/nvme-cli/pull/1223/commits/d361c3781778e658fb1f67109847ebf53111ecd9

igaw:
New version, rebased on current master and changed following names:

```
nvme_get_log_sup_pages    -> nvme_get_log_supported_log_pages                      
nvme_get_log_fid_sup_efcs -> nvme_get_log_fid_supported_effects                    
nvme_cap_mgmt             -> nvme_capacity_mgmt       
```
This should make things more consistent with the rest of the code base

src/nvme/ioctl.c
igaw: I suppose we should add the timeout to all commands then. I'll created an issue to track this, see #97

src/nvme/types.h
igaw: Ah got it, I changed the commit message title. Indeed a bit confusing.

igaw:
If no one is vetoing this down now, I am going to apply it soon (after processing the pending PRs).

igaw:
FYI, just a rebase to recent HEAD to do a final check before merge

PR #101, opened 2021-11-15
types: fix wrong bits nvme_id_psd apw[178:176], aps[183:182]

sc108-lee:

it used to be active_work_scale on nvme-cli, so rename it as apws (active power work & scale)
https://github.com/linux-nvme/nvme-cli/blob/nvme-cli-monolithic/linux/nvme.h#L258

### DIFF
--- a/doc/libnvme.rst
+++ b/doc/libnvme.rst
@@ -6280,9 +6280,8 @@ Returns true if given offset is 64bit register, otherwise it returns false.
     __u8 ips;
     __u8 rsvd19;
     __le16 actp;
-    __u8 apw;
-    __u8 aps;
-    __u8 rsvd23[8];
+    __u8 apws;
+    __u8 rsvd23[9];
   };
 
 **Members**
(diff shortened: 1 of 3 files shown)


doc/libnvme.rst
```diff
@@ -6340,12 +6339,12 @@ Returns true if given offset is 64bit register, otherwise it returns false.
   NVM subsystem over a 10 second period in this power state with
   the workload indicated in the Active Power Workload field.
 
-``apw``
+``apws``
+  APW occupy the first 3 bits of the field
```
igaw, line 6343: Could you use something like 'Bits 2-0:' to be more precises in the wording? This aligns also to the existing bit field description.

doc/libnvme.rst
```diff
@@ -6280,9 +6280,8 @@ Returns true if given offset is 64bit register, otherwise it returns false.
     __u8 ips;
     __u8 rsvd19;
     __le16 actp;
-    __u8 apw;
-    __u8 aps;
-    __u8 rsvd23[8];
+    __u8 apws;
+    __u8 rsvd23[9];
```
igaw, line 6284: This looks good. The 2.0a base spec agrees :)

igaw (changes requested):
Looks good, just needs small documentation update.

doc/libnvme.rst
sc108-lee: Thanks for the comments, I Just modified that explanation

igaw (approved):

igaw:
Next time also send a PR to update nvme-cli accordingly. I fix it up now as it breaks my merge attempt.

BTW, while doing the fixup I noticed 

```
                print_ps_power_and_scale(ctrl->psd[i].idlp,                                                                           
                                 nvme_psd_power_scale(ctrl->psd[i].ips));                                                             
                printf(" active_power:");                                                                                             
                print_ps_power_and_scale(ctrl->psd[i].actp,                                                                           
                                 nvme_psd_power_scale(ctrl->psd[i].apws));    
```
So this assumes 'ips' has the same bit layout which doesn't seem correct the spec. And after starring at the ips definition I think this part is also wrong in libnvme.

sc108-lee:
oh... I had that modification in company pc, but since it could be different commit hash (if it merged with cherry-pick, or rebased).
so I didnt uploaded that yet.
so you mean upload that fix at the same time?

*ips does not sam.
bits
151:150 Idle Power Scale (IPS) 
149:144 Reserved

igaw:
I've fixed nvme-cli, it builds again (not in the github CI) but locally. So this part is okay. You don't have do anything.

About: IPS I am not sure if our implementation is correct. I think the nvme-cli/libnvme are not correctly defining the field. If you have time to double check this I would appreciate it.

keithbusch:
I think the code is correct, the comment is wrong though. The power scale is just the upper 2 bits, not 3. The spec just makes this field very awkward to deal with.

keithbusch:
And as tempting as it is, don't use bitfields. Those are so poorly defined by C99 (if interested: 6.7.2.1, paragraph 10) that we shouldn't rely on how a compiler might layout the struct for wire defined data.

PR #103, opened 2021-11-15
tree: add namespace chrdev name to nvme_ns

minwooim:
Namespace device node is initialized into two different nodes: block
device and character device which is a generic device used when block
device is not available due to initialization failure in the kernel.

This can be directly retrieved from the `struct nvme_ns` to print out
the generic node in somewhere like `nvme list` command in nvme-cli.

Signed-off-by: Minwoo Im <minwoo.im.dev@gmail.com>

### DIFF
--- a/src/nvme/private.h
+++ b/src/nvme/private.h
@@ -45,6 +45,7 @@ struct nvme_ns {
 	int fd;
 	__u32 nsid;
 	char *name;
+	char *generic_name;
 	char *sysfs_dir;
 
 	int lba_shift;
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -1337,6 +1337,11 @@ const char *nvme_ns_get_name(nvme_ns_t n)
 	return n->name;
 }
 
+const char *nvme_ns_get_generic_name(nvme_ns_t n)
+{
+	return n->generic_name;
+}
+
 const char *nvme_ns_get_model(nvme_ns_t n)
 {
 	return n->c ? n->c->model : n->s->model;
@@ -1541,6 +1546,20 @@ static int nvme_ns_init(struct nvme_ns *n)
 	return 0;
 }
 
+static void nvme_ns_set_chrdev_name(struct nvme_ns *n, const char *name)
+{
+	char generic_name[PATH_MAX];
+	int instance, head_instance;
+	int ret;
+
+	ret = sscanf(name, "nvme%dn%d", &instance, &head_instance);
+	if (ret != 2)
+		return;
+
+	sprintf(generic_name, "ng%dn%d", instance, head_instance);
+	n->generic_name = strdup(generic_name);
+}
+
 static nvme_ns_t nvme_ns_open(const char *name)
 {
 	struct nvme_ns *n;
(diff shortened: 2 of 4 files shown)


src/nvme/tree.c
```diff
@@ -1541,6 +1546,17 @@ static int nvme_ns_init(struct nvme_ns *n)
 	return 0;
 }
 
+static void nvme_ns_set_chrdev_name(struct nvme_ns *n, const char *name)
+{
+	char chrdev_name[PATH_MAX];
+	int instance, head_instance;
+
+	sscanf(name, "nvme%dn%d", &instance, &head_instance);
```
igaw, line 1554: Please add error checking here. sscanf might fail.

src/nvme/private.h
```diff
@@ -45,6 +45,7 @@ struct nvme_ns {
 	int fd;
 	__u32 nsid;
 	char *name;
+	char *chrdev_name;
```
igaw, line 48: I am not sure if chdrev_name (and also nvme_ns_get_chrdev_name) is a good name. The existing nvme device is also a char device. The differenece is this one is the 'generic' one. So I think the generic bit should be somehow included in the naming scheme.

src/nvme/tree.h
```diff
@@ -452,6 +452,14 @@ const char *nvme_ns_get_sysfs_dir(nvme_ns_t n);
  */
 const char *nvme_ns_get_name(nvme_ns_t n);
 
+/**
+ * nvme_ns_get_chrdev_name() -
+ * @n:
+ *
+ * Return:
```
igaw, line 459: I know so far we are lacking a lot of documentation. But let's try to add some documentation to newly introduced APIs.

src/nvme/tree.h
```diff
@@ -452,6 +452,14 @@ const char *nvme_ns_get_sysfs_dir(nvme_ns_t n);
  */
 const char *nvme_ns_get_name(nvme_ns_t n);
 
+/**
+ * nvme_ns_get_chrdev_name() -
+ * @n:
+ *
+ * Return: 
+ */
+const char *nvme_ns_get_chrdev_name(nvme_ns_t n);
```
igaw, line 461: And you need to update libnvme.map to expose the function.

igaw (changes requested):

src/nvme/tree.h
minwooim: Got it! let me fill this up! Thanks.

src/nvme/private.h
minwooim: let me have this `generic_name`.

src/nvme/tree.c
minwooim: Sure, Thanks for the review.

src/nvme/tree.h
minwooim: Missed that one.  Will do this. Thanks.

minwooim:
Updated commit. Please have a look :)

igaw:
Could you please rename nvme_ns_set_chrdev_name() to  nvme_ns_set_generic_name() too?  I saw this after I merged this PR, so please as new PR. Thanks!

minwooim:
Sorry for the miss! Thanks!

igaw:
No worries. It's just software :)

PR #105, opened 2021-11-16
TP8013 & 8014 support

hreinecke:
Hi all,

here's a patchset to support TP8013 (unique discovery controller identification) and TP8014 (discovery controller discovery).
The kernel parts are already merged.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -391,7 +391,7 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 	struct nvme_fabrics_config *cfg = nvme_ctrl_get_config(c);
 	const char *transport = nvme_ctrl_get_transport(c);
 	const char *hostnqn, *hostid;
-	bool discover = false;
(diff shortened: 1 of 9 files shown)


src/nvme/fabrics.c
```diff
@@ -417,6 +432,8 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 			 nvme_ctrl_get_trsvcid(c)) ||
 	    (hostnqn && add_argument(argstr, "hostnqn", hostnqn)) ||
 	    (hostid && add_argument(argstr, "hostid", hostid)) ||
+	    (discover && !discovery_nqn &&
```
martin-belanger, line 435: This is good. It allows operating with older kernels that don't support the new "discovery" attribute as long as we use the well-known NQN (`nqn.2014-08.org.nvmexpress.discover`). I was wondering how you would address that. :grinning:

hreinecke:
Now with proper documentation of the added function prototypes.
Just to keep @igaw happy :-)

igaw:
Thanks :) No objections from my side.

PR #110, opened 2021-11-17
fabrics: add nvmf_default_config()

hreinecke:
'struct nvme_fabrics_config' has some default values, and the output
functions key off these values to filter the values being printed.
So to avoid applications having to know these default values this
patch introduces a function nvmf_default_config() to ensure that
every application is using the same defaults.

### DIFF
--- a/src/nvme/fabrics.h
+++ b/src/nvme/fabrics.h
@@ -125,6 +125,14 @@ const char *nvmf_qptype_str(__u8 qptype);
 const char *nvmf_cms_str(__u8 cms);
 
 /**
+ * nvmf_default_config - Default values for fabrics configuration
+ * @cfg: config values to set
+ *
+ * Initializes @cfg with default values.
+ */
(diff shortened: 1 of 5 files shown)


src/nvme/fabrics.h
```diff
@@ -124,6 +124,14 @@ const char *nvmf_qptype_str(__u8 qptype);
  */
 const char *nvmf_cms_str(__u8 cms);
 
+/**
+ * nvmf_default_config - Default values for fabrics configuration
+ * @cfg: config values to set
+ *
+ * Initializes @cfg with default values.
+ */
+void nvmf_default_config(struct nvme_fabrics_config *cfg);
```
igaw, line 133: nvmf_default_config() needs to be added to libnvme.map

igaw (changes requested):
Rest looks good.

src/nvme/fabrics.h
hreinecke: Done.

PR #111, opened 2021-11-17
[RFC] Add timeout argument to all commands

igaw:
Allow to set a timeout per command. Inspired by d588b59e2539 ("ioctl:
Add Timeout Parameter for Zone Management Commands")


igaw:
Not so sure about it. Though it would make the API consistent.

BTW, while I was it, I saw several commands which lack '__u32 *results'. Seems also a bit inconsistent.

hreinecke:
Please check PR #118 ; this is what I had in mind.
And we can build upon that to gradually move the missing parameters to the calling functions if and when it's required.

igaw:
I agree my attempt to pass the timeout parameter through all call levels is excessive. Your #118 is way better. Let's close this in favor of your PR.

PR #113, opened 2021-11-17
Add discovery_ctrl_set() to Python bindings.

martin-belanger:
Add discovery_ctrl_set() to Python bindings.

Following Hannes' recent TP8013 changes, we need
to add this method to the Python bindings.


martin-belanger:
Incomplete commits. I will resubmit.

PR #114, opened 2021-11-17
Add discovery_ctrl_set() to Python bindings

martin-belanger:
Add discovery_ctrl_set() to Python bindings

Following Hannes' recent TP8013 changes, we need
to add this method to the Python bindings.

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -295,6 +295,7 @@ struct nvme_ctrl {
   %immutable serial;
   %immutable sqsize;
   %immutable persistent;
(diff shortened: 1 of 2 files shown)


hreinecke:
Can you check PR #112? I think that should already do the trick...

martin-belanger:
PR #112 does not add `nvme_ctrl_set_discovery_ctrl()` to `libnvme.map`, which is needed by `discovery_ctrl_set()` in the Python bindings.

hreinecke:
Ah, indeed, you are right. I've forgot about that.

Issue #97, opened 2021-11-10
Add timeout to all commands


igaw:
@hreinecke  pointed out that it doesn't make sense to have only a timeout argument for the ZNS commands. Let's add them to the rest of the commands as well.

See also #94

igaw:
See #111

PR #118, opened 2021-11-18
RFC: Sanitize ioctl command wrappers

hreinecke:
There is quite some inconsistency as to whether the ioctl wrapper functions return the NVMe CQE result or whether they accept a timeout argument.
So this patchsets sanitizes the calling convention by always adding a timeout and result argument.


igaw:
Yes this is way saner than my attempt.

igaw:
Fixes: #97

hreinecke:
Go for it, then.

PR #119, opened 2021-11-18
tree: rename namespace function name chrdev to generic

minwooim:
The attribute in `struct nvme_ns` is `generic_name`, not chrdev.  So
rename the function name of this attribute.

Signed-off-by: Minwoo Im <minwoo.im.dev@gmail.com>


igaw:
Thanks!

PR #121, opened 2021-11-18
Add authentication keys to JSON schema

hreinecke:
Add the authentication key definitions to the JSON schema.


igaw:
LGTM

PR #122, opened 2021-11-18
Sanitize zns command arguments

hreinecke:
With the recent up the calling convention has been sanitzed to always
have the 'timeout' argument as second-to-last, and the 'result' as the
laster argument.
Modify the zns commands accordingly.


igaw:
looks good to me

PR #123, opened 2021-11-18
nvme-cli: ctrl-loss-tmo should accept -1 as value

maurizio-lombardi:
When passing the parameter --ctrl-loss-tmo=-1
nvme-cli silently drops it and replaces its value with the
default one (600) despite the fact that -1 is valid
and accepted by the kernel (it means "off" and
that it will try to reconnect forever).

This is due to the fact that -1 is treated like "variable unset"
by nvme-cli and the add_int_argument() function ignores
all negatives arguments.

This patch fixes the bug by allowing arguments to have
negative values and removing the if condition that assumes that
ctrl_loss_tmo = -1 means that the variable has not been set.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -207,7 +207,7 @@ static int add_int_argument(char **argstr, char *tok, int arg, bool allow_zero)
 {
 	char *nstr;
 
-	if (arg < 0 || (!arg && !allow_zero))
+	if (!arg && !allow_zero)
 		return 0;
 	if (asprintf(&nstr, "%s,%s=%d", *argstr, tok, arg) < 0) {
 		errno = ENOMEM;
@@ -412,9 +412,6 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 			errno = EINVAL;
 			return -1;
 		}
-		/* Use the default ctrl loss timeout if unset */
-                if (cfg->ctrl_loss_tmo == -1)
-			cfg->ctrl_loss_tmo = NVMF_DEF_CTRL_LOSS_TMO;
 	}
 
 	/* always specify nqn as first arg - this will init the string */
@@ -473,6 +470,7 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 	     add_int_argument(argstr, "fast_io_fail_tmo",
 			      cfg->fast_io_fail_tmo, false)) ||
 	    (strcmp(transport, "loop") &&
+	     cfg->tos != -1 &&
 	     add_int_argument(argstr, "tos", cfg->tos, true)) ||
 	    add_bool_argument(argstr, "duplicate_connect",
 			      cfg->duplicate_connect) ||


src/nvme/fabrics.c
```diff
@@ -473,6 +470,7 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 	     add_int_argument(argstr, "fast_io_fail_tmo",
 			      cfg->fast_io_fail_tmo, false)) ||
 	    (strcmp(transport, "loop") &&
+	     cfg->tos != -1 &&
```
hreinecke, line 473: That is an unrelated fix, and the underlying issue (of tos not being initialized correctly) has already been handled in PR #110.

src/nvme/fabrics.c
```diff
@@ -207,7 +207,7 @@ static int add_int_argument(char **argstr, char *tok, int arg, bool allow_zero)
 {
 	char *nstr;
 
-	if (arg < 0 || (!arg && !allow_zero))
+	if (!arg && !allow_zero)
```
hreinecke, line 210: I wonder it it wouldn't be more sensible to introduce an 'add_uint_argument' function or an additional argument 'allow_negative'; negative values are invalid for most arguments, and with this change we lose the ability to check for negative values.

src/nvme/fabrics.c
```diff
@@ -412,9 +412,6 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 			errno = EINVAL;
 			return -1;
 		}
-		/* Use the default ctrl loss timeout if unset */
-                if (cfg->ctrl_loss_tmo == -1)
-			cfg->ctrl_loss_tmo = NVMF_DEF_CTRL_LOSS_TMO;
```
hreinecke, line 417: Indeed, this needs to be removed; setting default values has been fixed with PR #110, and this is a left-over from previous attempts to set correct defaults.

hreinecke (changes requested):

src/nvme/fabrics.c
maurizio-lombardi: I had to add this because tos default value is -1; before this patch, tos was filtered out by add_int_argument(), but now it's passed to the kernel and the latter returns an error.

src/nvme/fabrics.c
maurizio-lombardi: At the moment, libnvme doesn't check the validity of the arguments passed by the user, it just filters them out, silently.
example: 
# ./nvme connect -t tcp -a 10.37.153.75 -n test-nqn -s 4080  --tos=-100
this command successfully completes even if tos is invalid.

I think it would be better to pass to the kernel whatever the user wants and let the kernel decide if it should accept, ignore or return an error.

src/nvme/fabrics.c
hreinecke: But isn't the error rather that 'tos' is not supported for loop?
IE shouldn't we exclude 'tos' entirely for 'loop'?

src/nvme/fabrics.c
hreinecke: I would agree with you if the user would have a chance to figure out what went wrong when sending down the string.
ATM there is no programmatic way how the user (or application) could differentiate between invalid arguments and the connection failing due to controller errors.
And if we don't have that we have to ensure that we're not passing down invalid arguments.

src/nvme/fabrics.c
maurizio-lombardi: I'm not sure I follow you here,  "tos" is already excluded when the transport is equal to "loop", it's what that strcmp() instruction is supposed to do.

src/nvme/fabrics.c
maurizio-lombardi: Ah, right, I see the problem.
So the build_options() function has to be rewritten, because right now nvme-cli doesn't check if an argument has a valid value or not, it adds it to the string it will pass to the kernel or it filters it out without warnings...

src/nvme/fabrics.c
hreinecke: But my point was that this fix is for 'ctrl_loss_tmo' handling, not 'tos' handling.
Yet this patch introduces changes to 'tos' handling.

hreinecke:
I've now rewritten this patch in PR #135; please check if it resolves your issue.

PR #124, opened 2021-11-18
Fix ccan mem-leaks

martin-belanger:
Fix ccan mem-leaks

meson provide the option: -Db_sanitize=address.

Ex: meson .build -Db_sanitize=address && ninja -C .build

This is used to configure the build environment with the
gcc -fsanitize=address option, which is a debug tool that 
can be used to find memory problems such as leaks or 
use after free.

The ccan configurator is a build tool to help configure
the libnvme project for building on different platform. It
is only used during the build process and thus was never
written with making sure that it didn't leak memory.

When enabling meson's -Db_sanitize=address the build fails
because the ccan configurator, which is now compiled with
the -fsanitize=address option, crashes due to memory leaks.

This patch fixes the memory leaks in ccac configurator.

Note that the -Db_sanitize=address is simply a convenience
option used during debugging and should not be used for normal
builds. Another meson option that is available for debugging
is -Dbuildtype=debug, which disables optimization.

### DIFF
--- a/ccan/tools/configurator/configurator.c
+++ b/ccan/tools/configurator/configurator.c
@@ -403,22 +403,29 @@ static bool run_test(const char *cmd, struct test *test)
 		size_t len;
 		const char *deps = test->depends;
 		char *dep;
+		char *s;
 
 		/* Space-separated dependencies, could be ! for inverse. */
 		while ((len = strcspn(deps, " "))) {
 			bool positive = true;
 			if (deps[len]) {
-				dep = strdup(deps);
-				dep[len] = '\0';
+				s = strdup(deps);
+				s[len] = '\0';
 			} else {
-				dep = (char *)deps;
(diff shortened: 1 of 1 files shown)


hreinecke:
Can someone remind me: why do we need the ccan configurator when running meson?
To my knowledge, meson does _exactly_ the same thing (and more) than the ccan configurator.
Can't we move the remaining bits into meson configuration and _not_ build it when using meson?

igaw:
I had the same question. I'll check the situation. 

Besides that, the fix should go upstream to the ccan project too.

martin-belanger:
I totally agree. In fact, if you look at the history of the top-level `meson.build` file, I believe that someone had removed ccan from meson at one point and then it got added right back.

igaw:
That was probably me, trying to figure out all the build details. I think we are solved most of the bigger build issues with meson by now.

PR #125, opened 2021-11-19
meson: do not build ccan configurator

hreinecke:
meson _is_ a configuration tool, so there is no point in building
the ccan configurator. Especially as the generated config file is
never used anywhere.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -143,7 +143,6 @@ incdir = include_directories(['.', 'ccan', 'src'])
 
 ################################################################################
 sources = []
-subdir('ccan')
 subdir('src')
 subdir('libnvme')
 subdir('test')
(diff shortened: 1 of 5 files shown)


igaw:
The build failure is likely caused by my recent renaming of the header files. Those includes should be libnvme-config.h to match with my change

igaw:
This was hidden by running the configurator which created a config file, I suppose.

igaw:
BTW, you can remove the ccan/meson.build along, me thinks

igaw:
see #126

meson.build
```diff
@@ -143,7 +143,6 @@ incdir = include_directories(['.', 'ccan', 'src'])
 
 ################################################################################
 sources = []
-subdir('ccan')
```
igaw, line 146: IIRC we need this for the debug build. For this configuration the additional ccan/str/debug.c is necessary.

igaw:
Closing this one, see #126

PR #127, opened 2021-11-19
tree: Allow empty tr_addr for loop and pcie in nvme_init_ctrl()

tbzatek:
Just a tiny issue found by testing against local loop target.


hreinecke:
I found that this is not sufficient; one needs to check for empty address in nvme_ctrl_alloc(), too
(Cf PR #132).

tbzatek:
Good point, works for me, thanks!

PR #128, opened 2021-11-19
Clean up remaining ccan artifacts from meson

martin-belanger:
Clean up remaining ccan artifacts from meson.

After further investigation, I found that none of the ccan source
files used to build libnvme using meson are actually used. This
patch removes those files.


igaw:
I'd prefer to leave them in the build. The reason being they are part of the ccan library and I rather not start cherry picking individual files. Just removing them from the build is not enough, the files should be removed. 

It's easy to foresee where we add code which depends on those *.c files. 

I though the debug target depends on some of the code in the *.c files.

igaw:
As explained in my last comment, we should leave the ccan *.c files in the build, because it is used when
CCAN_LIST_DEBUG=1 and CCAN_STR_DEBUG=1 are set. This is currently missing in our meson setup.
Sending a PR for it.

Anyway this is what have for Makefile as well:

`src/Makefile:libccan_srcs := $(wildcard $(CCANDIR)ccan/*/*.c)
`

PR #135, opened 2021-11-29
Fix up handling of '-1' values

hreinecke:
Some settings like 'ctrl_loss_tmo' accept '-1' as valid values, so we need to fixup add_int_argument() to handle it correctly.
And a '-1' does not imply 'unset', so we shouldn't assume that.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -222,6 +222,22 @@ static int add_int_argument(char **argstr, char *tok, int arg, bool allow_zero)
 	return 0;
 }
 
+static int add_int_or_minus_one_argument(char **argstr, char *tok, int arg)
+{
+	char *nstr;
+
+	if (arg < -1)
+		return 0;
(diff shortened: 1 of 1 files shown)


maurizio-lombardi:
Hello, it doesn't work because you call add_int_argument() against "tos" with  allow_zero_or_minus_one set to true.
tos = -1 is then rejected by the kernel (see the log below)

Note that setting allow_zero_or_minus_one to false is probably bad too, because tos=0 afaik is a valid value.

```
[root@kvm-02-guest08 nvme-cli]# ./nvme connect -t tcp -a 10.8.1.94 -n test-nqn -s 4080 --ctrl-loss-tmo=-1
no controller found

dmesg:
[ 2411.204334] nvme_fabrics: ctrl_loss_tmo < 0 will reconnect forever
[ 2411.206412] nvme_fabrics: I/O fail on reconnect controller after 0 sec
[ 2411.209203] nvme_fabrics: Invalid type of service -1
```

hreinecke:
I've reworked the patch to add a new function "add_int_or_minus_one_argument()", which should address the concerns.

PR #136, opened 2021-11-30
Fix memory leak in SWIG module

martin-belanger:
Fix memory leak in SWIG module.

The API PyDict_SetItemString() does not steal references
when adding objects to a dictionary. Therefore, the caller
must decrement the reference count after adding objects to
the dictionary if those objects are no longer needed.


hreinecke:
Looks good.

PR #137, opened 2021-12-01
types: Fabrics doc strings updates

tbzatek:
Includes a typo correction (`enum nvmf_addr_familiy`), prefix changes (`enum nvme_trtype` -> `enum nvmf_trtype`), transition away from anonymous enums and added some more enum members according to the standard. All subtle API breaks.

Haven't found any reference to `struct nvmf_disc_log_entry.common`, leaving it without description for the time being.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3830,7 +3830,7 @@ enum nvme_ae_info_notice {
 };
 
 /**
- * enum nvme_subsys_type -
+ * enum nvme_subsys_type - Type of the NVM subsystem.
  * @NVME_NQN_DISC:		Discovery type target subsystem
  * @NVME_NQN_NVME:		NVME type target subsystem
  * @NVME_NQN_CURR:		Current Discovery type target subsystem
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -3980,31 +4014,43 @@ enum  nvme_rdma_prtype {
...
- * enum -
- * @NVMF_TCP_SECTYPE_NONE: No Security
- * @NVMF_TCP_SECTYPE_TLS:  Transport Layer Security
+ * enum nvmf_tcp_sectype - Transport Specific Address Subtype Definition for
+ * 	  NVMe/TCP Transport
+ * @NVMF_TCP_SECTYPE_NONE:  No Security
+ * @NVMF_TCP_SECTYPE_TLS:   Transport Layer Security version 1.2
+ * @NVMF_TCP_SECTYPE_TLS13: Transport Layer Security version 1.3 or a subsequent
+ * 			    version. The TLS protocol negotiates the version and
+ * 			    ciper suite for each TCP connection.
```
martin-belanger, line 4033: typo: `ciper` -> `cipher`

igaw:
nvme-cli builds fine with this change.

PR #138, opened 2021-12-01
Misc. fabrics fixes

tbzatek:
Two more tiny fixes. Not sure what was the original idea behind `nvmf_hostnqn_generate()` but I believe adding extra newline should be a caller responsibility.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -958,7 +958,7 @@ char *nvmf_hostnqn_generate()
 	if (ret < 0)
 		return NULL;
 
-	if (asprintf(&hostnqn, "nqn.2014-08.org.nvmexpress:uuid:%s\n", uuid_str) < 0)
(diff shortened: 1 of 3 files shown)


martin-belanger:
To answer your question.

`nvmf_hostnqn_generate()` was created to support nvme-cli's "`nvme gen-hostnqn`" command. This may be used by distros to create the file `/etc/nvme/hostnqn` when installing `nvme-cli`. For example, Debian performs the following commands in the `nvme-cli.postinst` script:
```
if [ "$1" = "configure" ]; then
    if [ ! -s /etc/nvme/hostnqn ]; then
        nvme gen-hostnqn > /etc/nvme/hostnqn
    fi
    if [ ! -s /etc/nvme/hostid ]; then
        uuidgen > /etc/nvme/hostid
    fi
fi
```

I agree with you that the caller should be the one to add the newline character, if needed.

igaw:
I hope we have rooted out all the warts with gen-hostnqn by now :)

PR #139, opened 2021-12-01
libnvme: Add new events support in PEL

wxiong38:
Add two new events support in header file.

Signed-off-by: Wen Xiong <wenxiong@linux.ibm.com>

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3044,6 +3044,8 @@ enum nvme_persistent_event_types {
     NVME_PEL_FORMAT_COMPLETION_EVENT	= 0x08,
     NVME_PEL_SANITIZE_START_EVENT	= 0x09,
     NVME_PEL_SANITIZE_COMPLETION_EVENT	= 0x0a,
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -3044,6 +3044,8 @@ enum nvme_persistent_event_types {
     NVME_PEL_FORMAT_COMPLETION_EVENT	= 0x08,
     NVME_PEL_SANITIZE_START_EVENT	= 0x09,
     NVME_PEL_SANITIZE_COMPLETION_EVENT	= 0x0a,
+    NVME_PEL_SET_FEATURE_EVENT		= 0x0b,
+    NVME_PEL_TELEMETRY_CRT		= 0x0c,
```
igaw, line 3048: Don't we need a struct definition for the data entry (Figure 242)?

igaw (changes requested):

igaw:
Anyway, I've went ahead and applied it via command line. I took the liberty to fix the author in the commit.

PR #140, opened 2021-12-08
MAINT: Fix size_t type name build error

ikegami-t:
This fix is for the pull request change #1266.

Signed-off-by: Tokunori Ikegami <ikegami.t@gmail.com>

### DIFF
--- a/src/nvme/linux.h
+++ b/src/nvme/linux.h
@@ -10,6 +10,7 @@
 #define _LIBNVME_LINUX_H
 
 #include "types.h"
+#include <stddef.h>
 
 /**
  * nvme_fw_download_seq() -


src/nvme/linux.h
```diff
@@ -10,6 +10,7 @@
 #define _LIBNVME_LINUX_H
 
 #include "types.h"
+#include <stddef.h>
```
igaw, line 13: Please move the standard includes before our project headers. Thanks.

igaw (changes requested):
It would be better if you could explain in your commit message what's broken. Just
referencing a github PR (not even in the same project) is pretty useless when looking
at the git history.

src/nvme/linux.h
ikegami-t: > Please move the standard includes before our project headers. Thanks.

Since it seems the CI build error below suggests as the PR change order.
  <https://github.com/linux-nvme/nvme-cli/runs/4458586468?check_suite_focus=true#step:4:152>
```
libnvme/src/nvme/linux.h:13:1: note: ‘size_t’ is defined in header ‘<stddef.h>’; did you forget to ‘#include <stddef.h>’?
   12 | #include "types.h"
  +++ |+#include <stddef.h>
   13 | 
```

Also it is desribed the includes order as below.
  <https://llvm.org/docs/CodingStandards.html#include-style>
1. Main Module Header
2. Local/Private Headers
3. LLVM project/subproject headers (clang/..., lldb/..., llvm/..., etc)
4. System #includes

The clang-format tool enable the order fomatting as default.
  Note: The includes order option: IndentCaseLabels can be disabled by the clang-format configuration file.

igaw: I don't argue against what clang says on this topic. The project doesn't follow this rule at this point. I just don't
want avoid mixed styles.

This means if you want to change this, please fixup all include orders (but not with
the reformat tooling hammer method please).

ikegami-t: > I don't argue against what clang says on this topic. The project doesn't follow this rule at this point. I just don't want avoid mixed styles.
> 
> This means if you want to change this, please fixup all include orders (but not with the reformat tooling hammer method please).

Thanks for your coments. Noted them. The changes for the following PR to resolve the CI build error but the PR has been closed. So let me close this PR also.
https://github.com/linux-nvme/nvme-cli/pull/1266

PR #143, opened 2021-12-10
zns: Add support for zone random write area (zrwa)

sc108-lee:
NVMe - TP 4076 Zoned Random Write Area 2021.08.23 - Ratified

Co-authored-by: Klaus Jensen <k.jensen@samsung.com>


src/nvme/ioctl.c
```diff
@@ -2034,13 +2036,14 @@ int nvme_resv_report(int fd, __u32 nsid, bool eds, __u32 len,
 }
 
 int nvme_zns_mgmt_send(int fd, __u32 nsid, __u64 slba,
-		       enum nvme_zns_send_action zsa,
-		       bool select_all, __u32 data_len,
+		       enum nvme_zns_send_action zsa, bool select_all,
+			   __u8 zsaso, __u32 data_len,
```
igaw, line 2040: whitepace damage in front of zsaso

src/nvme/ioctl.h
```diff
@@ -2714,6 +2714,7 @@ int nvme_resv_report(int fd, __u32 nsid, bool eds, __u32 len,
  * @slba:	Starting logical block address
  * @zsa:	Zone send action
  * @select_all:	Select all flag
+ * @zsaso:  Zone Send Action Specific Option
```
igaw, line 2717: The description is not aligned.

src/nvme/types.h
```diff
@@ -2077,6 +2077,10 @@ struct nvme_zns_lbafe {
  * @mor: Maximum Open Resources
  * @rrl: Reset Recommended Limit
  * @frl: Finish Recommended Limit
+ * @numzrwa: Number of ZRWA Resources
+ * @zrwafg: ZRWA Flush Granularity
+ * @zrwasz: ZRWA Size
+ * @zrwacap: ZRWA Capability
```
igaw, line 2083: Please also align these description. While at it, you could also align the rest of the documentation :)

src/nvme/util.c
```diff
@@ -243,6 +243,8 @@ static const char * const nvm_status[] = {
 	[NVME_SC_INVALID_PI]			 = "Invalid Protection Information: The command's Protection Information Field settings are invalid for the namespace's Protection Information format",
 	[NVME_SC_READ_ONLY]			 = "Attempted Write to Read Only Range: The LBA range specified contains read-only blocks",
 	[NVME_SC_CMD_SIZE_LIMIT_EXCEEDED]	 = "Command Size Limit Exceeded",
+	[NVME_SC_ZNS_INVALID_OP_REQUEST] = "Invalid Zone Operation Request: The operation requested is invalid",
```
igaw, line 246: Please align the string here as well.

igaw (changes requested):
just whitespace nitpicking. the rest looks good (checked it with the spec).

src/nvme/ioctl.c
sc108-lee: Thanks for the details, I didnt realize that above line using both(space & tab).
fixed with spaces

src/nvme/ioctl.h
sc108-lee: Thanks. few of docs they using spaces, so I thought this use same.

sc108-lee:
@igaw all aligned :)
Thanks for the checking.
code indentation is mixed with spaces and tabs.
it makes confusion,
so wouldn't it be better to use cpplint or something that defines the rule?

src/nvme/ioctl.h
igaw: It seems most of the doc is somehow aligned. That's why I think it makes sense to have here as well.

igaw:
> @igaw all aligned :) Thanks for the checking. code indentation is mixed with spaces and tabs. it makes confusion, so wouldn't it be better to use cpplint or something that defines the rule?

The project uses the Linux coding style, which is a bit hard to capture in a tool and relies on the author to have good taste. Whatever that means. As far I am not aware that there is a tool out there which can get it right. 

I am hesitant to go the route like in https://github.com/linux-nvme/nvme-cli/pull/1266. The resulting code looks awful in my opinion.

PR #144, opened 2021-12-10
ioctl: Add Offset Type to get_log_page()

igaw:
Port of the "add get log page 2.0 spec fields" patch.


igaw:
I'll add a enum for the OT as we have for CSI. This should make reading the argument list for nvme_get_log() easier.

igaw:
After a discussion with @hreinecke, I changed the type of 'ot' to bool.

PR #145, opened 2021-12-13
libnvme: Add NVMe MI Features: Controller Metadata (0x7E) and Namespa…

ikegami-t:
…ce Metadata (0x7F).

Splitted between nvme-cli and libnvme.

Commit b8a403b ("Add NVMe MI Features: Controller Metadata (0x7E) and Namespace Metadata (0x7F).")
Commit 2ae875d ("fix whitespace damage")
Signed-off-by: Nate Roiger <nate.roiger@hpe.com>
Signed-off-by: Tokunori Ikegami <ikegami.t@gmail.com>

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3547,6 +3547,55 @@ enum nvme_apst_entry {
 };
 
 /**
+ * struct nvme_host_metadata_element_desc -
+ * @type:
+ * @rev:
+ * @len:
+ * @val:
+ */
+struct nvme_host_metadata_element_desc {
+	__u8			type;	/* Element Type */
+	__u8			rev;	/* Element Revision */
+	__u16			len;	/* Element Length */
+	__u8			val[0];	/* Element Value (UTF-8) */
+};
+
+/**
+ * struct nvme_host_metadata -
+ * @ndesc:
+ * @rsvd1:
+ * @descs:
+ */
+struct nvme_host_metadata {
+	__u8						ndesc;
+	__u8						rsvd1;
+	union {
+		struct nvme_host_metadata_element_desc descs[0];
+		__u8 descs_buf[4094];
+	};
+};
+
+/**
+ * enum nvme_mi_ctrl_metadata_type -
+ */
+enum nvme_mi_ctrl_metadata_type {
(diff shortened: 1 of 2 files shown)


src/nvme/types.h
```diff
@@ -5305,6 +5330,9 @@ enum nvme_features_id {
 	NVME_FEAT_FID_RESV_MASK					= 0x82,
 	NVME_FEAT_FID_RESV_PERSIST				= 0x83,
 	NVME_FEAT_FID_WRITE_PROTECT				= 0x84,
+
+	NVME_MI_FEAT_CTRL_METADATA	= 0x7E,
+	NVME_MI_FEAT_NS_METADATA	= 0x7F,
```
igaw, line 5335: These should follow the name scheme from above and also placed accordingly their value, that is before NVME_FEAT_FID_SW_PROGRESS.

src/nvme/types.h
```diff
@@ -3546,6 +3546,31 @@ enum nvme_apst_entry {
 	NVME_APST_ENTRY_ITPT_MASK = 0xffffff,
 };
 
+struct nvme_mi_host_metadata_element_desc {
```
igaw, line 3549: Not so sure about the 'mi' in the name here. It doesn't show up in the spec.

src/nvme/util.c
```diff
@@ -434,6 +434,10 @@ int nvme_get_feature_length(int fid, __u32 cdw11, __u32 *len)
 	case NVME_FEAT_FID_WRITE_PROTECT:
 		*len = 0;
 		break;
+	case NVME_MI_FEAT_CTRL_METADATA:
+	case NVME_MI_FEAT_NS_METADATA:
+		*len = 4096;
```
igaw, line 439: After looking at the rest of the function, it looks like the idea is to define the struct to the right size and just do a sizeof(struct nvme_mi_host_metadata) here.

igaw (changes requested):
Thanks for working on these missing commits.

ikegami-t:
> Thanks for working on these missing commits.

Thanks for your reviewing. I have just fixed the chenged for the comments and force-pushed them into both libnvme and nvme-cli so please review again.

igaw:
Thanks for helping out. I've applied the version in #146 was based on your work here.

PR #147, opened 2021-12-16
build: Add subpackages contents to .gitignore

sc108-lee:
we just need to maintain .wrap files for dependencies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ config.log
 cscope.*
 
 .build
+
+subprojects
+!subprojects/*.wrap


sc108-lee:
@igaw I saw your commits for nvme-cli like below
https://github.com/linux-nvme/nvme-cli/commit/7a0217ffa27756b94f0c18bd1baf3ab33164ed92

you are going to add more dependencies as wrap file, so how about to change .gitignore as this PR?
so we don't need to care when new subproject comes

igaw:
Ah nice, didn't know about the '!' feature. Yes let's do this.

igaw:
BTW, I've picked your patch for my nvme-cli PR.

sc108-lee:
cool :)
thanks for picking it up

PR #149, opened 2021-12-16
Use command argument structure to avoid overly long argument lists

hreinecke:
Quite some functions like nvme_get_features(), nvme_set_features(), or nvme_get_log() have quite long argument lists, making it very error prone due to mixed up arguments. It also make future extension quite unwieldy as one have to modify the argument list and all derived functions.
To avoid all of that this patch introduces a function-specific argument structure with an 'args_size' as the first field, to make it easy to add new parameters to those functions and to allow for bounds-checking on the passed-in structure.
With that the function should be backwards compatible, and updating the shared library will be easy here.

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -1413,34 +1413,46 @@ int nvme_set_features_iocs_profile(int fd, __u8 iocsi, bool save)
 				   save, NULL);
 }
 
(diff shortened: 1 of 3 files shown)


hreinecke:
Has been superseded by PR#150.

keithbusch:
Some of these commands do indeed take quite a few parameters. I agree a struct may be easier.

PR #150, opened 2021-12-17
Rework the libnvme ioctl API

hreinecke:
The ioctl functions take a lot of arguments, making it very error prone as one has to remember the order of those arguments.
This patchset reworks the API to always use an argument structure to avoid this problem.
In using a structure we can also implement backwards compability as new arguments can simply be added to the end of the structure, and we can do bounds-checking in the called function to ensure we're not overrunning the structure. Older programs will continue to run as the calling sequence doesn't change.


src/nvme/ioctl.h
```diff
@@ -386,17 +386,54 @@ int nvme_get_nsid(int fd, __u32 *nsid);
...
  * Return: The nvme command status if a response was received (see
  * &enum nvme_status_field) or -1 with errno set otherwise.
  */
-int nvme_identify(int fd, enum nvme_identify_cns cns, __u32 nsid,
-		  __u16 cntid, __u16 nvmsetid, __u16 domid,
-		  __u8 uuidx, __u8 csi, void *data, __u32 timeout,
-		  __u32 *result);
+int nvme_identify(struct nvme_identify_args *args);
+
+static int nvme_identify_cns_nsid(int fd, enum nvme_identify_cns cns,
```
igaw, line 417: Isn't 'static inline' needed here as well?

src/nvme/types.h
```diff
@@ -2773,6 +2773,11 @@ struct nvme_self_test_log {
 	struct nvme_st_result	result[NVME_LOG_ST_MAX_RESULTS];
 } __attribute__((packed));
 
+enum nvme_cmd_get_log_telemetry_host_lsp {
+	NVME_LOG_TELEM_HOST_LSP_RETAIN			= 0,
+	NVME_LOG_TELEM_HOST_LSP_CREATE			= 1,
+};
```
igaw, line 2779: Is this on purpose here? Looks like a left over from something else

igaw:
I assume you decided to move to 'static inline' function to reduce the number of symbols we need to export via the library? I like this idea. I am trying to think of an way to reduce a bit the full struct definition repetition, though not sure if there is a good way.

src/nvme/types.h
hreinecke: Oh, no, that's on purpose.
Originally these definitions were internally in srv/nvme/ioctl.c, which won't work once I've pulled the declaration into src/nvme/ioctl.h . And it arguably should have been in srv/nvme/types.h in the first place.

hreinecke:
Well, there were several ideas here.
The one is to provide templates on how the library should be used, and also to make it easier for future improvements like adding new arguments and stuff.
And the other is that in moving things into ioctl.h we give users a chance to simply copy the definitions to implement their own functions and/or allow them to code additional arguments.

hreinecke:
And in general I've tried to keep only the core functions in the library (ie those functions converting the argument structures into the passthrough commands themselves). All the wrapper functions (which are just converting command-line arguments into the argument structure) should be kept in the header file to clarifying they _are_ wrapper functions.
Not that I've followed that to the letter, but it's the idea anyway.

src/nvme/types.h
igaw: ah okay, didn't spot the removal lines. the web ui stuff isn't good for this.

igaw:
So what's the status on this? I personally would like to go this way. @keithbusch any thoughts on this?

keithbusch:
Yeah, I'm okay with this. The number of parameters are indeed getting a bit much, and most uses don't even care about them, so this looks like the right way to go.

igaw:
Alright, let me refresh this series in this case. I don't know if Hannes has time for it. And since things depends on this one, it's better to get it applied it sooner than later :)

igaw:
While updating nvme-cli I run into this style question. Which is the better style. Obviously one works only with C99:

```
-               err = nvme_get_features(fd, NVME_NSID_ALL,
-                                       NVME_FEAT_FID_HOST_BEHAVIOR, 0, 0, 0,
-                                       len, buf, NVME_DEFAULT_IOCTL_TIMEOUT,
-                                       &result);
+               args.args_size  = sizeof(args);
+               args.fd         = fd;
+               args.fid        = NVME_FEAT_FID_HOST_BEHAVIOR;
+               args.nsid       = NVME_NSID_ALL;
+               args.sel        = 0;
+               args.cdw11      = 0;
+               args.uuidx      = 0;
+               args.data_len   = len;
+               args.data       = buf;
+               args.timeout    = NVME_DEFAULT_IOCTL_TIMEOUT;
+               args.result     = &result;
+               err = nvme_get_features(&args);
                if (err > 0) {
                        nvme_show_status(err);
                } else if (err < 0) {
@@ -3244,10 +3255,20 @@ static int get_feature_id(int fd, struct feat_cfg *cfg, void **buf,
                memset(*buf, 0, cfg->data_len);
        }
 
-       return nvme_get_features(fd, cfg->feature_id, cfg->namespace_id,
-                                cfg->sel, cfg->cdw11, cfg->uuid_index,
-                                cfg->data_len, *buf,
-                                NVME_DEFAULT_IOCTL_TIMEOUT, result);
+       struct nvme_get_features_args args = {
+               .args_size      = sizeof(args),
+               .fd             = fd,
+               .fid            = cfg->feature_id,
+               .nsid           = cfg->namespace_id,
+               .sel            = cfg->sel,
+               .cdw11          = cfg->cdw11,
+               .uuidx          = cfg->uuid_index,
+               .data_len       = cfg->data_len,
+               .data           = *buf,
+               .timeout        = NVME_DEFAULT_IOCTL_TIMEOUT,
+               .result         = result
+       };
+       return nvme_get_features(&args);
 }

```

keithbusch:
I prefer your 2nd example, definitely. It implicitly zeroes any fields you didn't set, which I think is what we'd want.

tasleson:
I believe it would be good to wrap any user visible structures with `#pragma pack`, to allow applications to use things like `-fpack-struct=1` if they are so inclined to do so.

keithbusch:
Shouldn't we use 'reserved' fields if there are any holes in the struct to force all alignments to be the same no matter how you compiled it?

src/nvme/ioctl.c
```diff
@@ -594,48 +594,61 @@ int nvme_nvm_identify_ctrl(int fd, struct nvme_id_ctrl_nvm *id)
...
+		.lsp = NVME_LOG_LSP_NONE,
+		.lsi = NVME_LOG_LSI_NONE,
+		.rae = rae,
+		.uuidx = NVME_UUID_NONE,
+		.len = len,
+		.log = log,
+		.timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
+		.result = NULL,
+		.csi = NVME_CSI_NVM,
+		.ot = false,
```
igaw: I think I would move the timeout and result to the end of the struct to make it more consistent with the rest.

(Since I am already working on it I'll do it)

igaw:
I've just rebased the last version from Hannes  and added a missing export. No changes in the patches so far.

src/nvme/ioctl.c
igaw: Thinking on it a bit more, maybe it would make more sense to have the common ones at the beginning of the struct. This would be 'args_size', fd', 'timeout and 'result'.

tasleson:
> Shouldn't we use 'reserved' fields if there are any holes in the struct to force all alignments to be the same no matter how you compiled it?

That's another way to approach it.  The structures appear to be in the same order as they were in the argument list.  In some cases this isn't the ideal order to "fill in the holes".  The users will run into an error path at runtime when the structure size is incorrect if they have compiled with different structure packing, but I think we should try to avoid it.

PR #151, opened 2021-12-20
linux: Add nvme_get_logical_block_size()

igaw:
Add helper to retrieve logical block size using nvme_identify_ns().


igaw:
@hreinecke  is the name of the function okay? I just ported it but somehow I am not completely convinced of the name. To be fair it would match the name of the sysfs entry.

igaw:
Just a rebase, checking if build still is fine

PR #153, opened 2021-12-21
fabrics: Declare loop index at the beginning of the block

igaw:
The gcc in Centos 6.10 complains with

nvme/fabrics.c:909: error: ‘for’ loop initial declarations are only allowed in C99 mode
nvme/fabrics.c:909: note: use option -std=c99 or -std=gnu99 to compileyour code

Let's declare the index at the beginning of the basic block.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/1293


brendon:
Thanks @igaw! :)

PR #155, opened 2021-12-21
Replace make by meson

martin-belanger:
Replace make by meson. This involves replacing all Makefiles 
by meson.build files. Also the top-level configure script now 
simply invokes meson. The top-level Makefile is just a wrapper 
for meson. It supports the following operations:

```
make
make test
make install
make clean    # Remove build artifacts but keep configuration
make purge    # Remove build artifacts and configuration
```


martin-belanger:
Hi @igaw and @keithbusch,
Resubmitting the changes to replace `make` by `meson`. I replaced the contents of the top-level `configure` script, which now simply invokes `meson` to configure the project. Similarly, I replaced the contents of the top-level `Makefile` to invoke `meson` or `ninja` depending on the operation performed. The operations supported by this new `Makefile` are:
```
make
make test
make install
make clean    # Remove build artifacts but keep configuration
make purge    # Remove build artifacts and configuration
```
I wasn't sure if you wanted to keep the following operations. I kept them in the top-level `Makefile`, but please let me know if I should remove them.
```
make cscope
make tag-archive
make create-archive
make archive
make srpm
```
The top-level `configure` script will automatically be invoked when running any of the `make` operations. However, `configure` can also be invoke prior to running `make` if someone wants to tweak some of the build parameters. For example, if someone doesn't want to build the `Python` bindings, they could simply do as follows:
```
./configure -Dpython=false
make
```

Makefile
```diff
@@ -70,8 +49,8 @@ tag-archive:
 	@git tag $(TAG)
 
 create-archive:
-	@git archive --prefix=$(NAME)-$(VERSION)/ -o $(NAME)-$(VERSION).tar.gz $(TAG)
-	@echo "The final archive is ./$(NAME)-$(VERSION).tar.gz."
+	@git archive --prefix=$(TAG)/ -o $(TAG).tar.gz $(TAG)
+	@echo "The final archive is ./$(TAG).tar.gz."
```
eli-schwartz, line 53: This is a duplicate of your above addition:

```make

install test dist: ${BUILD-DIR}
	cd ${BUILD-DIR} && meson $@
```

igaw:
'make test' fails for Python on my system. 'meson .build && ninja -C .build test' works fine.

```
$ make test
cd .build && meson test
ninja: Entering directory `/home/wagi/work/libnvme-upstream/.build'
[16/16] Linking target test/main-test
1/7 main                               OK              0.01s
2/7 main                               OK              0.01s
3/7 cpp                                OK              0.01s
4/7 register                           OK              0.01s
5/7 zns                                OK              0.01s
6/7 [Python] import libnvme            FAIL            0.02s   exit status 1
>>> MALLOC_PERTURB_=0 PYTHONPATH=/home/wagi/work/libnvme-upstream/.build/libnvme/.. /usr/bin/python3 -c 'from libnvme import nvme'

7/7 [Python] create ctrl object        FAIL            0.03s   exit status 1
>>> MALLOC_PERTURB_=0 PYTHONPATH=/home/wagi/work/libnvme-upstream/.build/libnvme/.. /usr/bin/python3 /home/wagi/work/libnvme-upstream/.build/../libnvme/tests/create-ctrl-obj.py
```

igaw:
> I wasn't sure if you wanted to keep the following operations. I kept them in the top-level `Makefile`, but please let me know if I should remove them.
> 
> ```
> make cscope
> make tag-archive
> make create-archive
> make archive
> make srpm
> ```

I'd say we should only keep them if those targets really work properly. 'make archive' creates an empty tar file at this point.

> The top-level `configure` script will automatically be invoked when running any of the `make` operations other than `clean` and `purge`. However, `configure` can also be invoke prior to running `make` if someone wants to tweak some of the build parameters. For example, if someone doesn't want to build the `Python` bindings, they could simply do as follows:
> 
> ```
> ./configure -Dpython=false
> make
> ```

I think it doesn't make sense to add a configure script which dosen't mimic the exact version of the script. We either wrap all the old configure options to meson or we do not add the configure script at all. Looking at the configure script this would be --prefix, --includir, --libddir, ... --disable-uuid, --disable-json. At this point we don't have a --disable-python. Given this I'd suggest we don't add the configure script. We want the users to start using meson natively. The make wrapper is for those of us who have hard time changing their muscle memory :)

martin-belanger:
Ok. That makes sense. I will eliminate the configure script.

martin-belanger:
> 'make test' fails for Python on my system. 'meson .build && ninja -C .build test' works fine.
> 

That's interesting. Both methods work on my system. I have changed the `Makefile` to use `ninja -C .build test` instead of `meson test`.

martin-belanger:
Hi @igaw - As you suggested, I removed the `configure` script. I also removed these targets from the Makefile: "`cscope`", "`tag-archive`", "`create-archive`", "`archive`", and "`srpm`". I'm not sure if any of these are still needed and I honestly don't know what the intention was for having them in the Makefile. If any of these are still needed, then someone with the proper knowledge (not me) will need to add them.

keithbusch:
Those are just convenient build targets. They're probably not necessary and were probably copied from other open source projects.

PR #156, opened 2021-12-22
util: Define inline function as static inline

igaw:
The gcc in Centos 6.10 complains with

  src/nvme/util.h:120: multiple definition of `nvme_feature_decode_arbitration'

'inline' alone does not guarantee that the function gets inlined. The
compiler is still free to decide, thus in this case it doesn't inline
and that's why we having multiple implementation with the same
name (in different compile units). 'static inline' forces the inline
which was the indent of this code.

Fixes: [#1293](https://github.com/linux-nvme/nvme-cli/issues/1293)


brendon:
Thanks @igaw, you missed this one I think:

```
inline void nvme_feature_decode_host_identifier(__u32 value, bool *exhid)
```

igaw:
Thanks for reviewing. Double checked it this time too :)

PR #158, opened 2021-12-23
linux: Add support to generate DH-HMAC-CHAP keys

igaw:
Initially the DH-HMAC-CHAP key generation code was part of the
nvme-cli repository and depended on OpenSSL 1. Move the key generation
code into the library. While at it also add support for OpenSSL 3.

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -68,6 +68,7 @@ LIBNVME_1_0 {
 		nvme_fw_commit;
 		nvme_fw_download;
 		nvme_fw_download_seq;
+		nvme_gen_dhchap_key;
 		nvme_get_ana_log_len;
 		nvme_get_attr;
 		nvme_get_ctrl_attr;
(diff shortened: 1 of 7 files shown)


igaw:
I can't get nvme-cli linking this as subproject. The linker is not adding libuuid, json-c nor openssl when tries to link. Something is not correctly setup with meson.

igaw:
I think the problem is caused by libnvme getting the wrong 'config.h' as input. Not the the one from the subprojects instead the main one.

igaw:
Alright figured it out. Moving the config.h to libnvme is the right thing to do, but obviously, ccan was not happy. The trick is to use internal lookup path. Now things are working :)

src/libnvme.map
```diff
@@ -68,6 +68,7 @@ LIBNVME_1_0 {
 		nvme_fw_commit;
 		nvme_fw_download;
 		nvme_fw_download_seq;
+		nvme_gen_dhchap_key;
```
johnmeneghini, line 71: Perfect.

igaw:
Just a rebase; checking if build is still good.

PR #160, opened 2021-12-29
log: Make nvme_log_message thread-local

tbzatek:
Also change it to a fixed-length buffer to avoid memory deallocation
burden on a thread exit.

----

This is my attempt to make `nvme_log_message` thread safe/aware by making the log message string thread-local. For simplicity I've changed it from dynamically allocated string to a fixed-length buffer with arbitrary size 512 (feel free to change it!) to avoid the need for explicit deallocation.

Not much tested so far, wanted to make this API change as soon before any public release.

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -98,7 +98,7 @@ static int discover_err = 0;
   if (connect_err == 1) {
     SWIG_exception(SWIG_AttributeError, "Existing controller connection");
   } else if (connect_err) {
-    if (nvme_log_message)
+    if (nvme_log_message[0])
       SWIG_exception(SWIG_RuntimeError, nvme_log_message);
     else
       SWIG_exception(SWIG_RuntimeError, "Connect failed");
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -172,6 +172,8 @@ LIBNVME_1_0 {
 		nvme_lockdown;
 		nvme_log_level;
 		nvme_log_message;
+		nvme_log_pid;
+		nvme_log_timestamp;
 		nvme_lookup_host;
 		nvme_lookup_subsystem;
 		nvme_namespace_attach_ctrls;
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -35,7 +35,7 @@
 int nvme_log_level = DEFAULT_LOGLEVEL;
 bool nvme_log_timestamp;
 bool nvme_log_pid;
-char *nvme_log_message = NULL;
+__thread char nvme_log_message[NVME_LOG_MESSAGE_LEN] = {'\0'};
 
 void __attribute__((format(printf, 3, 4)))
 __nvme_msg(int lvl, const char *func, const char *format, ...)
@@ -83,9 +83,8 @@ __nvme_msg(int lvl, const char *func, const char *format, ...)
 		message = NULL;
 	va_end(ap);
 
-	if (nvme_log_message)
-		free(nvme_log_message);
-	nvme_log_message = strdup(message);
+	strncpy(nvme_log_message, message, NVME_LOG_MESSAGE_LEN - 1);
+	nvme_log_message[NVME_LOG_MESSAGE_LEN - 1] = '\0';
 
 	if (lvl <= nvme_log_level)
 		fprintf(stderr, "%s%s", header ? header : "<error>",
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -33,8 +33,8 @@
 #endif
 
 int nvme_log_level = DEFAULT_LOGLEVEL;
-bool nvme_log_timestamp;
-bool nvme_log_pid;
+bool nvme_log_timestamp = false;
+bool nvme_log_pid = false;
 __thread char nvme_log_message[NVME_LOG_MESSAGE_LEN] = {'\0'};
 
 void __attribute__((format(printf, 3, 4)))
--- a/src/nvme/log.h
+++ b/src/nvme/log.h
@@ -21,10 +21,12 @@
 #define __nvme_log_func NULL
 #endif
 
+#define NVME_LOG_MESSAGE_LEN 512
+
 extern int nvme_log_level;
 extern bool nvme_log_timestamp;
 extern bool nvme_log_pid;
-extern char *nvme_log_message;
+extern __thread char nvme_log_message[NVME_LOG_MESSAGE_LEN];
 
 void __attribute__((format(printf, 3, 4)))
 __nvme_msg(int lvl, const char *func, const char *format, ...);


igaw:
Can you elaborate on your use case? Or is this just for our Python binding?

Anyway, I think we should reconsider to export nvme_log_message. If it is just for the Python binding, we could 
add some sort of callback API for the log messages. Exposing a buffer like this is just a bad idea IMO.

tbzatek:
No, no Python considered at this point. Let me explain in broader view:

As a general purpose library a consumer would like to get detailed context in case an error happens. Something more than the generic `errno` codes that are not always fully aligned and self-explanatory. An arbitrary error message that could potentially be presented to user in upper layers.

This is mostly related to fabrics as ioctls have their status codes with corresponding `nvme_status_to_string()` utility function to get a human readable message.

Also since the library functions might be used from multiple threads (not considering potential issues at this point), a single global error message buffer would not be sufficient. Declaring thread safety or awareness would be also nice to have before any public release.

The `nvme_msg()` also logs on `stderr` with the default `nvme_log_level = LOG_NOTICE` settings, something that may work for `nvme-cli` but not when used from another library or daemon. That's just a detail however and subject to further changes.

So yeah, there are multiple options available, hiding the message buffer behind an accessor function or setting up a custom logging callback might work as well. Exposing a bare buffer was just the simplest way.

igaw:
I think we are on the same page here. Ideally the library doesn't need any internal global variables/states. Usually, this is just asking for troubles. I haven't checked if there is anything else. 

Anyway, I am not convinced that going the TLS way is the right way here. The library depends then a compiler and libc which support it (older distros?) and the modestly unused per thread memory allocation. But it's pretty simple and would work. 

I was pondering on my idea also for a while. My initial idea was to introduce some kind of register function for an arbitrary number of callback function. But this makes live harder than it needs to be. One single callback should be enough. The synchronization has to be handled by the callback function.

tbzatek:
In the line of current library minimalism, the logging callback might be the easiest way. Perhaps something like this?

```c
typedef void (*nvme_log_func) (int level, const char *func, const char *msg);
void nvme_register_log_callback(nvme_log_func log_func);
```
with an existing internal `nvme_msg()` macro taking care of formatting the message.

Then `nvme-cli` would simply print `msg` on `stderr`, perhaps adding a timestamp or a PID. Multithreaded consumers would save the message somewhere in TLS for example. There won't be any context or user pointer passed back with the callback, for simplicity.

tasleson:
> In the line of current library minimalism, the logging callback might be the easiest way. Perhaps something like this?

Callbacks can make things difficult for other languages to use the library.  Having something on the heap with a function in the library to free it or having the API user provide a buffer and it's length to log into might make it easier for other languages to leverage.

igaw:
@tasleson seems you have more experience with this kind of problems. If you have time to work on this, I would really appreciate it!

tasleson:
@igaw
> @tasleson seems you have more experience with this kind of problems. If you have time to work on this, I would really appreciate it!

I'll work with @tbzatek on this.

tasleson:
I spent some time looking at this yesterday.  At the moment I actually think what Tomas has proposed is the best approach I can see without making sweeping changes to the library architecture.   I believe `__thread` is supported by gnu and clang toolsets.  We could #ifdef the `__thread` to allow those that don't have support to not require it, but also be limited in functionality too.

Having the ability to register a callback for logging won't be straight forward for some languages, but that wouldn't necessarily be required.

I'm assuming the kernel IOCTL interface has all the required locking to work with threads/multi-process user space?

igaw:
Alright, I was a bit hesitant about the `__thread` as I would like to avoid any hard dependency on this feature. IIRC alternative libc implementation might not support it (but maybe my memory is completely outdated here.) So if we can make this an optional thing I am not against it. Especially if we don't have any better idea how to handle it.

hreinecke:
Wouldn't it be easier if we add a function to supply the buffer from the caller?

igaw:
> Wouldn't it be easier if we add a function to supply the buffer from the caller?

But how does this help with concurrent updates?

tasleson:
> > Wouldn't it be easier if we add a function to supply the buffer from the caller?
> 
> But how does this help with concurrent updates?

Perhaps a better question is, should we even try to make the library thread safe?  I believe some of the operations require a sequence of commands to complete.  Perhaps the entire discussion is moot, and the statement of only one process should be talking to the driver at any one time should simply be documented, if this is indeed the case?

What is the expected functionality between userspace/kernel with regards to concurrent operations?

igaw:
Except the logging related variable, there is only one other global variable  `static struct nvme_host *default_host;  ` in tree.c. I'd say the library is thread safe. It is making the ioctl calls more simpler to use by providing a C function wrappers. There is not a lot of magic in it.

And as far I can tell the userspace/kernel boundary should also be safe for concurrent interaction too (though it might not really make sense ). If not we need to fix it

The logging part is a bit of an odd ball in this regard.

hreinecke:
And really, the log message thingie is _just_ to provide a better error description in case of 'connect' fails when called from the python module.
Essentially me being too lazy figuring out the various return codes.
So I guess the better option here is to clarify the various return codes from nvmf_add_ctrl(), then we could just look at the return code / errno in the python wrapper and format the appropriate error message _there_ without having to use the log message buffer.
(Or figure out how stderr can be routed to the python SWIG interface.)

In either case, I doubt we need to spend lots and lots of time here.

igaw:
You mean something like this?

```
 %exception nvme_ctrl::connect {
+  int save_err = dup(fileno(stderr));
+  fflush(stderr);
+  dup2(save_err, fileno(stderr));
+  close(fileno(stderr));
+
   connect_err = 0;
   $action  /* $action sets connect_err to non-zero value on failure */
   if (connect_err == 1) {
     SWIG_exception(SWIG_AttributeError, "Existing controller connection");
   } else if (connect_err) {
-    if (nvme_log_message)
-      SWIG_exception(SWIG_RuntimeError, nvme_log_message);
-    else
+    static char msg[1024];
+
+    read(save_err, msg, 1024);
+    msg[1023] = '\0';
+    SWIG_exception(SWIG_RuntimeError, msg);
+  } else {
       SWIG_exception(SWIG_RuntimeError, "Connect failed");
   }
+  dup2(save_err, fileno(stderr));
+  close(save_err);
 }
```

Anyway, I agree, there is not much todo here. If we stick with the current solution I would say it make sense to prefix the symbol `__nvme_log_buffer` and don't care if we going to remove or change it.

hreinecke:
> You mean something like this?
> 
> ```
>  %exception nvme_ctrl::connect {
> +  int save_err = dup(fileno(stderr));
> +  fflush(stderr);
> +  dup2(save_err, fileno(stderr));
> +  close(fileno(stderr));
> +
>    connect_err = 0;
>    $action  /* $action sets connect_err to non-zero value on failure */
>    if (connect_err == 1) {
>      SWIG_exception(SWIG_AttributeError, "Existing controller connection");
>    } else if (connect_err) {
> -    if (nvme_log_message)
> -      SWIG_exception(SWIG_RuntimeError, nvme_log_message);
> -    else
> +    static char msg[1024];
> +
> +    read(save_err, msg, 1024);
> +    msg[1023] = '\0';
> +    SWIG_exception(SWIG_RuntimeError, msg);
> +  } else {
>        SWIG_exception(SWIG_RuntimeError, "Connect failed");
>    }
> +  dup2(save_err, fileno(stderr));
> +  close(save_err);
>  }
> ```
> 

Nah, I'd rather thought to audit nvmf_add_ctrl(), and define which error codes are set under which circumstances.
Then we have a clear set of error codes, and can add a switch in the python wrapper to translate the error codes into strings.
That would avoid us having to use the message log buffer at all.

igaw:
Yeah, the dub stuff is just a big hack. Let me try to find the error codes.

igaw:
As expected there are several ambiguous mappings, especially EINVAL. I don't think this is going to work either unless we introduce our own error codes for configuration errors.

hostname2traddr:
- error codew from getaddrinfo
  EAI_ADDRFAMILY
  EAI_AGAIN
  ...
  "failed to resolve host %s info\n", c->traddr);

- inet_ntop
  EAFNOSUPPORT
  ENOSPC
  -> "failed to get traddr for %s\n", c->traddr);

  EINVAL
  -> "failed to get traddr for %s\n", c->traddr);

build_option:
- EINVAL
  -> "need a transport (-t) argument\n");
  -> "need a address (-a) argument\n");

__nvmf_add_ctrl:
- error codes from open
  EACCES
  EEXIST
  EINTR
  EINVAL
  EISDIR
  ELOOP
  ...
  -> "Failed to write to %s: %s\n",

- error codes from write
  EAGAIN
  EBADF
  EFBIG
  ...
  -> "Failed to read from %s: %s\n",

- EINVAL
  -> "Failed to parse ctrl info for \"%s\"\n", argstr

hreinecke:
I'd rather do a more sweeping conversion:

diff --git a/src/nvme/fabrics.c b/src/nvme/fabrics.c
index 7645e40..df66c04 100644
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -363,6 +363,7 @@ static int hostname2traddr(nvme_ctrl_t c)
        ret = getaddrinfo(c->traddr, NULL, &hints, &host_info);
        if (ret) {
                nvme_msg(LOG_ERR, "failed to resolve host %s info\n", c->traddr);
+               errno = EHOSTUNREACH;
                return ret;
        }
 
@@ -380,13 +381,15 @@ static int hostname2traddr(nvme_ctrl_t c)
        default:
                nvme_msg(LOG_ERR, "unrecognized address family (%d) %s\n",
                        host_info->ai_family, c->traddr);
-               ret = -EINVAL;
+               errno = EINVAL;
+               ret = -1;
                goto free_addrinfo;
        }
 
        if (!p) {
                nvme_msg(LOG_ERR, "failed to get traddr for %s\n", c->traddr);
-               ret = -errno;
+               errno = EADDRNOTAVAIL
+               ret = -1;
                goto free_addrinfo;
        }
        c->traddr = strdup(addrstr);

hreinecke:
Sigh. The paste function obviously does not support markup.

hreinecke:
Or, even simpler, just ignore the individual error codes from hostaddr2traddr(), and always fold it into EHOSTUNREACH.
And fold the errors from 'build_options' onto EINVAL.

That leaves only __nvme_add_ctrl().

hreinecke:
Yes. But name them 'ENVME_CONNECT_XXX' to clarify that it's _just_ about the connect call.

igaw:
Right, sounds good. Let me render this into a proper PR.

igaw:
Let's close this PR. We are removing nvme_log_buffer.

tasleson:
@hreinecke FYI: use triple backtics with `patch` as the syntax highlighter, e.g.

```patch
diff --git a/src/libnvme.map b/src/libnvme.map
index f70bf2c..c4df7c4 100644
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -169,6 +169,7 @@ LIBNVME_1_0 {
                nvme_init_ctrl_list;
                nvme_init_dsm_range;
                nvme_init_id_ns;
+               nvme_io;
                nvme_io_passthru64;
                nvme_io_passthru;
                nvme_lockdown;
```

tbzatek:
Works for me, thanks!

Issue #72, opened 2021-10-19
Missing dependencies


leongross:
# Issue: 
When building `libnvme` the build fails due to the missing `pyhon3` development headers.
```
$ meson .build
The Meson build system
Version: 0.54.2
Source dir: /home/ec2-user/libnvme
Build dir: /home/ec2-user/libnvme/.build
Build type: native build
Project name: libnvme
Project version: v0.1
C compiler for the host machine: cc (gcc 7.5.0 "cc (SUSE Linux) 7.5.0")
C linker for the host machine: cc ld.bfd 2.35.1.20201123-7
C++ compiler for the host machine: c++ (gcc 7.5.0 "c++ (SUSE Linux) 7.5.0")
C++ linker for the host machine: c++ ld.bfd 2.35.1.20201123-7
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Run-time dependency uuid found: YES 2.33.1
Run-time dependency json-c found: YES 0.15
Checking if "__builtin_type_compatible_p" compiles: YES
Checking if "typeof" compiles: YES
Checking if "byteswap.h" compiles: YES
Checking if "bswap64" links: YES
Checking if "statement-expr" compiles: YES
Checking if "isblank" links: YES
Configuring config.h using configuration
Configuring libnvme.spec using configuration
Program python3 found: YES (/usr/bin/python3)
Program swig found: YES (/usr/bin/swig)
Dependency python found: YES (sysconfig)
Build targets in project: 10

Found ninja-1.10.0 at /usr/bin/ninja
$ cd .build/
$ ninja
[18/27] Compiling C object 'pynvme/37eb337@@_nvme.cpython-36m-x86_64-linux-gnu@sha/meson-generated_.._nvme_wrap.c.o'
FAILED: pynvme/37eb337@@_nvme.cpython-36m-x86_64-linux-gnu@sha/meson-generated_.._nvme_wrap.c.o
[...]

PR #163, opened 2022-01-07
build: Detect if Python.h header is present

igaw:
Make sure the devel package is also installed not just the the Python
interpreter.

Fixes: #72

### DIFF
--- a/libnvme/meson.build
+++ b/libnvme/meson.build
@@ -15,7 +15,8 @@ if want_python != 'false'
     python3 = import('python').find_installation('python3')
     py3_dep = python3.dependency(required: want_python == 'true')
     swig = find_program('swig', required: want_python == 'true')
-    have_python_support = py3_dep.found() and swig.found()
+    header_found = cc.has_header('Python.h', dependencies: py3_dep)
+    have_python_support = py3_dep.found() and swig.found() and header_found
 else
     have_python_support = false
 endif


igaw:
@eli-schwartz sorry to ask such a simple question but I was not able to find anything in the meson documentation on this topic.

Is this the way how we are supposed to check if 'Python.h' is installed on the system? One thing which bothers me, do we actually know it's the matching Python.h version with the Python interpreter found in the path.

igaw:
Thanks @eli-schwartz. Will change to `cc.has_header`.

The question on which Python interpreter is used popped up in #72. If I understood it correctly, during the setup step Python 3.6 (with devel package installed) is detected. But during the build Python 3.9 is used which doesn't have the Python.h header file.  Hmm, that would also mean my header test here wouldn't be enough I suppose.

igaw:
On a second thought, if in #72 Python 3.6 is found (so my previous claim is wrong about two different python version invovled) but there is nothing checking if the header files are present this PR would indeed be the right fix.

eli-schwartz:
> during the setup step Python 3.6 (with devel package installed) is detected. But during the build Python 3.9 is used

I don't understand this. If you have multiple versions of python installed on your system (one of which has a devel package installed, and which you want to build against) then meson will always consistently use whichever python it found in find_installation. It may be the wrong python, hence no devel headers, but it should ***NEVER*** detect one during `meson setup` and then write out `ninja` rules which use the other.

If the wrong python is detected, this can be overridden using, for example, a `--native-file foo.ini` with a binaries table specifying the path to the python interpreter (which is then used by `find_program('python3')` as well as `find_installation('python3')`).

igaw:
> > during the setup step Python 3.6 (with devel package installed) is detected. But during the build Python 3.9 is used
> 
> I don't understand this. If you have multiple versions of python installed on your system (one of which has a devel package installed, and which you want to build against) then meson will always consistently use whichever python it found in find_installation. It may be the wrong python, hence no devel headers, but it should _**NEVER**_ detect one during `meson setup` and then write out `ninja` rules which use the other.

Sorry, my bad. I got confused again because I was still thinking in the Makefile based world :)
So all should be good.

> If the wrong python is detected, this can be overridden using, for example, a `--native-file foo.ini` with a binaries table specifying the path to the python interpreter (which is then used by `find_program('python3')` as well as `find_installation('python3')`).

Nice! Thanks for explaining.

eli-schwartz:
No problem, happy to help. :)

> sorry to ask such a simple question

It's not necessarily such a simple question. ;)

Anyway, I'm actually quite willing to have people ping me for meson questions and/or ask for my style review on meson changes (no promises about project specific details, but I can definitely provide advice on what meson is capable of).

Issue #165, opened 2022-01-09
Build documentation error


alexmchou:
Hi,
I used "meson .build -Dman=true" then "cd .build, ninja", but got following error messages:
[1/1] Generating man with a custom command.
FAILED: doc/libnvme.1 
/usr/bin/sphinx-build -b man /home/zxm/libnvme/doc /home/zxm/libnvme/.build/doc
Running Sphinx v1.8.5
building [mo]: targets for 0 po files that are out of date
building [man]: all manpages
updating environment: 2 added, 0 changed, 0 removed
reading sources... [ 50%] index
reading sources... [100%] libnvme

/home/zxm/libnvme/doc/libnvme.rst:3906: WARNING: Unexpected indentation.
/home/zxm/libnvme/doc/libnvme.rst:6585: WARNING: Unexpected indentation.
/home/zxm/libnvme/doc/libnvme.rst:6583: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/zxm/libnvme/doc/libnvme.rst:6586: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/zxm/libnvme/doc/libnvme.rst:10192: WARNING: Inline strong start-string without end-string.
/home/zxm/libnvme/doc/libnvme.rst:10652: WARNING: Unexpected indentation.
/home/zxm/libnvme/doc/libnvme.rst:11132: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/zxm/libnvme/doc/libnvme.rst:11153: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/zxm/libnvme/doc/libnvme.rst:11153: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/zxm/libnvme/doc/libnvme.rst:11180: WARNING: Inline interpreted text or phrase reference start-string without end-string.

Sphinx error:
master file /home/zxm/libnvme/doc/contents.rst not found
ninja: build stopped: subcommand failed.

igaw:
Does this also happen when you do 

```
meson .build -Dman=true
ninja -C .build
```

We had the same problem for 'test'.  Let me try to figure out what's going on.

BTW, if you want to update the in tree sources check the update-docs.sh script.

martin-belanger:
[...]

PR #167, opened 2022-01-10
Fix error building the man pages

martin-belanger:
Fix for issue https://github.com/linux-nvme/libnvme/issues/165

I'm not familiar with Sphinx. By looking up the issue on Google, I found that by default Sphinx looks for a file named `contents.rst`. To specify a different name, in our case `libnvme.rst`, we must define it in `conf.py` as:
```
master_doc = 'libnvme'
```
This seems to fix the issue reported in https://github.com/linux-nvme/libnvme/issues/165.


igaw:
Thanks!

PR #170, opened 2022-01-13
Fill holes in ioctl's argument structs

igaw:
Fixes #169

PR #171, opened 2022-01-13
Fill holes in ioctl's argument structs (alternative version)

### LINKED ISSUE #169: Fill holes in ioctl's argument structs
See discussion in #150

igaw:
@hreinecke suggested to rearrange the members instead adding reserved members. 
This version also adds `__attribute__((__packed__))`.

See also #170 

Fixes #169

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -1203,7 +1203,7 @@ int nvme_ns_attach(struct nvme_ns_attach_args *args)
 
 	if (args->args_size < sizeof(*args))
 		return -EINVAL;
-	return nvme_submit_admin_passthru(args->fd, &cmd, NULL);
+	return nvme_submit_admin_passthru(args->fd, &cmd, args->result);
 }
 
 int nvme_fw_download(struct nvme_fw_download_args *args)
--- a/src/nvme/ioctl.h
+++ b/src/nvme/ioctl.h
@@ -1845,34 +1845,33 @@ static inline int nvme_get_log_persistent_event(int fd,
 /**
  * nvme_set_features_args - Arguments for the NVMe Admin Set Feature command
  * @fd:		File descriptor of nvme device
- * @fid:	Feature identifier
+ * @result:	The command completion result from CQE dword0
+ * @timeout:	Timeout in ms
  * @nsid:	Namespace ID, if applicable
  * @cdw11:	Value to set the feature to
  * @cdw12:	Feature specific command dword12 field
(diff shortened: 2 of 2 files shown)


igaw:
That was a bit of work. I am toast.

Unfortunately, there is a serious drawback in this version. The problem is that `__attribute__((packed))` means align(1). The compiler starts to throw warnings for `&args.member` which is not a byte type. And yes we do have a bunch of them. So when using ` __attribute__((packed, aligned(__alignof__(__u64)))) ` we can't rely on the `args_size = sizeof(arg)` trick anymore.

hreinecke:
I think it's okay; there's some style suggestion as to whether 'sizeof(__u32 *)' for the alignment is the best to describe a pointer size, but that can be fixed up later.

Reviewed-by: Hannes Reinecke <hare@suse.de>

igaw:
Extending a arg struct is possible with the args_size trick but only when we add a new member to an arg struct and the sizeof(arg) changes. For example:

```
struct nvme_resv_report_args {
        int                        args_size;            /*     0     4 */
        int                        fd;                   /*     4     4 */
        __u32 *                    result;               /*     8     8 */
        __u32                      nsid;                 /*    20     4 */
        struct nvme_resv_status *  report;               /*    24     8 */
        __u32                      len;                  /*    32     4 */
        _Bool                      eds;                  /*    36     1 */

        /* size: 40, cachelines: 1, members: 8 */
        /* padding: 3 */
        /* last cacheline: 40 bytes */
} __attribute__((__aligned__(8)));
```

The sizeof operator will say `sizeof(struct nvme_resv_report_args) = 40` and let's assume we want to add a new bool,
we need to pad the remaining 3 bytes:

```
struct nvme_resv_report_args {
        int                        args_size;            /*     0     4 */
        int                        fd;                   /*     4     4 */
        __u32 *                    result;               /*     8     8 */
        __u32                      nsid;                 /*    20     4 */
        struct nvme_resv_status *  report;               /*    24     8 */
        __u32                      len;                  /*    32     4 */
        _Bool                      eds;                  /*    36     1 */
        __u8                       rsvd37[3];            /*    37     3 */
        _Bool                      fancy;                /*    40     1 */

        /* size: 48, cachelines: 1, members: 10 */
        /* padding: 7 */
        /* last cacheline: 48 bytes */
} __attribute__((__aligned__(8)));
```
Any new member will increase the existing struct by at least 8 bytes.

I don't have a real problem with this solution but I just wanted to highlight it and that we all agree this is fine.

PR #173, opened 2022-01-14
log: Introduce libnvme error codes

igaw:
Add libnvme specific error codes which map to a specific error
message. The system error codes in errno are too overloaded to figure
out what is going wrong.

With this we can remove the nvme_log_message buffer. This avoids
having a global library buffer.

Fixes: #160

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -361,7 +361,7 @@ static int hostname2traddr(nvme_ctrl_t c)
 	ret = getaddrinfo(c->traddr, NULL, &hints, &host_info);
 	if (ret) {
 		nvme_msg(LOG_ERR, "failed to resolve host %s info\n", c->traddr);
-		return ret;
+		return -ENVME_CONNECT_RESOLVE;
 	}
 
 	switch (host_info->ai_family) {
@@ -378,13 +378,13 @@ static int hostname2traddr(nvme_ctrl_t c)
 	default:
 		nvme_msg(LOG_ERR, "unrecognized address family (%d) %s\n",
 			host_info->ai_family, c->traddr);
-		ret = -EINVAL;
+		ret = -ENVME_CONNECT_ADDRFAM;
 		goto free_addrinfo;
 	}
 
 	if (!p) {
 		nvme_msg(LOG_ERR, "failed to get traddr for %s\n", c->traddr);
-		ret = -errno;
+		ret = -ENVME_CONNECT_TRADDR;
 		goto free_addrinfo;
 	}
 	c->traddr = strdup(addrstr);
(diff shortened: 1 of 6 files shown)


src/nvme/fabrics.c
```diff
@@ -531,7 +531,7 @@ static int __nvmf_add_ctrl(const char *argstr)
 	}
 
 	nvme_msg(LOG_ERR, "Failed to parse ctrl info for \"%s\"\n", argstr);
-	errno = EINVAL;
+	errno = ENVME_CONNECT_PARSE;
```
hreinecke, line 534: This should be the return value, not the errno, to be consistent with the previous changes.

src/nvme/tree.c
```diff
@@ -1156,7 +1156,7 @@ int nvme_init_ctrl(nvme_host_t h, nvme_ctrl_t c, int instance)
 	if (strcmp(c->transport, "loop")) {
 		c->address = nvme_get_attr(path, "address");
 		if (!c->address) {
-			errno = ENXIO;
+			errno = ENVME_CONNECT_INVAL_TR;
```
hreinecke, line 1159: All the other functions called from nvmf_add_ctrl() would return the error code directly; nvme_init_ctrl() is the only one returning -1 and setting errno.
We need to be consistent here; suggestion is to have all functions from nvmf_add_ctrl() return -1 and setting the errno.

src/nvme/tree.c
```diff
@@ -1166,13 +1166,13 @@ int nvme_init_ctrl(nvme_host_t h, nvme_ctrl_t c, int instance)
 	if (!subsys_name) {
 		nvme_msg(LOG_ERR, "Failed to lookup subsystem name for %s\n",
 			 c->name);
-		errno = ENXIO;
+		errno = ENVME_CONNECT_LOOKUP_SUBSYS_NAME;
```
hreinecke, line 1169: Same here. We need to be consistent wrt errno and return values.

src/nvme/tree.c
```diff
@@ -1166,13 +1166,13 @@ int nvme_init_ctrl(nvme_host_t h, nvme_ctrl_t c, int instance)
...
 		ret = -1;
 		goto out_free_name;
 	}
 	s = nvme_lookup_subsystem(h, subsys_name, c->subsysnqn);
 	if (!s) {
-		errno = ENXIO;
+		errno = ENVME_CONNECT_LOOKUP_SUBSYS;
```
hreinecke, line 1175: And here, too.

hreinecke (changes requested):

src/nvme/fabrics.c
igaw: Yeah, I think the function is a bit inconsistent with the error handling.

src/nvme/tree.c
igaw: Yes, makes sense.

src/nvme/fabrics.c
igaw: Updated accordingly. Internal function (static) return the error code direct. The exported function map the error code to errno and return -1 in the error case.

hreinecke:
So, this looks good. Should be ready to merge.

Reviewed-by: Hannes Reinecke <hare@suse.de.

PR #177, opened 2022-01-16
Implement nvme_errno_to_string()

hreinecke:
Add a function nvme_errno_to_string() to provide a string describing
the error from nvme_add_ctrl().

### DIFF
--- a/src/nvme/util.c
+++ b/src/nvme/util.c
@@ -476,3 +476,25 @@ int nvme_get_directive_receive_length(enum nvme_directive_dtype dtype,
 		return -EINVAL;
 	}
 }
+
+static const char * const libnvme_status[] = {
+	[ENVME_CONNECT_RESOLVE] = "failed to resolve host",
+	[ENVME_CONNECT_ADDRFAM] = "unrecognized address family",
+	[ENVME_CONNECT_TRADDR] = "failed to get transport address",
+	[ENVME_CONNECT_TARG] = "no transport specified",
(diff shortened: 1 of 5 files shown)


igaw:
Looks good. Thanks!

src/nvme/util.c
```diff
@@ -476,3 +476,27 @@ int nvme_get_directive_receive_length(enum nvme_directive_dtype dtype,
 		return -EINVAL;
 	}
 }
+
+static const char * const libnvme_status[] = {
+	[ENVME_CONNECT_RESOLVE] = "failed to resolve host",
+	[ENVME_CONNECT_ADDRFAM] = "unrecognized address family",
+	[ENVME_CONNECT_TRADDR] = "failed to get traddr",
+	[ENVME_CONNECT_TARG] = "need a transport (-t) argument",
+	[ENVME_CONNECT_AARG] = "need a address (-a) argument\n",
```
tbzatek, line 485: For a pure library use, no commandline is necessarily involved and the `(-t)` and `(-a)` switches might be misleading. These are the strings that will potentially be presented to end user in a GUI.

tbzatek (commented):
Thanks for this addition, it comes handy.

src/nvme/util.c
igaw: Good point. And maybe it should be rephrased to "missing transport argument" etc. I've just copied the strings from the nmve_log() entries. Was a bit lazy :)

hreinecke: Fixed now.

PR #178, opened 2022-01-16
Docs strings updates (#5)

tbzatek:
A couple of minor updates. I've also changed the `NVMF_DISC_EFLAGS_` defines into an `enum` for consistency. Shouldn't cause any significant compatibility issues.


hreinecke:
Thanks for doing this.

Issue #174, opened 2022-01-14
Python tests fail in debug build


igaw:
When the library is configured as debug build the Python test fail due to missing str_strstr() function from the ccan library.

martin-belanger:
I'll take a look

martin-belanger:
@igaw - I figured it out.

When in debug mode, `CCAN_STR_DEBUG` is defined. This causes `#ifdef CCAN_STR_DEBUG` to evaluate to true and the standard POSIX APIs such as `strstr()` to be replaced by `str_strstr()` (see file `ccan/str/str.h`). I'm not sure why we go through all this trouble. Maybe this comment in `ccan/str/str.h` gives us a clue:
```
/* With GNU magic, we can make const-respecting standard string functions. */
```
In any case, when `CCAN_STR_DEBUG` is defined, any application that includes `ccan/str/str.h` will now be instructed to call `str_strstr()` instead of `strstr()`. And it will expect to find `str_strstr()` in `libnvme.so`. However, `str_strstr()` is not exposed in `libnvme.so`'s global symbols. It needs to be added to `libnvme.map`. The question is: _should debug wrapper functions such as `str_strstr()` be added to `libnvme.map`_?

The other thing we could do is to keep the standard POSIX APIs (i.e. `strstr()`, `strchr()`, `strrchr()`, etc.) and never replace them by wrapper functions.

martin-belanger:
BTW, while working on this I found some `meson.build` clean-ups that I would like to submit. If that's ok?

igaw:
Ah that's the reason. The str debug feature of ccan was already helpful in one of my debug session, so I'd like to keep it somehow. I wonder if it would be possible to add the str_strstr during debug builds to the export list.

This brings up the topic if we want to introduce a EXPORT() macro which manages the export symbols and drop the libnvme.map approach. Another point is that we need to annotate the header files anyway so that we can 
deprecated function etc. But that is something for another issue I suppose :)

Sure, send the clean-ups.

eli-schwartz:
If it's in a public header, you must export it. If it's in a private header, you should not export it. :p 

Maybe for libnvme-internal useful utility functions, you'd prefer to create an e.g. `libcommon = static_library('common', sources, install: false)` and have both the libnvme library and any executables link to it -- then that static library would contain functions which aren't exported, but are used across both shared libraries and executables.

igaw:
Fixed with #185. Thanks a lot.

PR #179, opened 2022-01-16
Consolidate ccan sources into src/meson.build

martin-belanger:
The ccan directory is just used as a source directory containing
files that get built along with the other libnvme sources. There
is no point in having a meson.build in the ccan directory just to 
list the sources to be added to src/meson.build. In fact, that just 
makes it harder to see what is actually getting build. This commit 
removes the ccan/meson.build file and changes the definition of 
the "sources" in src/meson.build to include the ccan files.


martin-belanger:
Hi @igaw - This is the clean up stuff I mentioned in https://github.com/linux-nvme/libnvme/issues/174

src/meson.build
```diff
@@ -14,19 +16,26 @@ sources += [
     'nvme/log.c',
     'nvme/tree.c',
     'nvme/util.c',
+    join_paths(source_dir, '../ccan/ccan/list/list.c'),
+    join_paths(source_dir, '../ccan/ccan/str/debug.c'),
+    join_paths(source_dir, '../ccan/ccan/str/str.c'),
```
eli-schwartz, line 21: You don't really need to join these paths to the source dir.

What is the difference between `$PWD/../ccan/ccan/list/list.c` and `../ccan/ccan/list/list.c`? Meson processes them as files either way, and resolves them relative to current_source_dir.

There is a better reason to use it for the version-script. Meson doesn't know that that link argument is a file, so it cannot relativize it.

igaw:
I don't know what you have against ccan/meson.build. Anyway Eli had a good point with building a private static ccan library which we just add to libnvme and the python binding library. Let's rather to this.

martin-belanger:
Agreed. A private static ccan library would certainly be better than the current passing around of the variable `sources` between 3 separate meson.build files.

PR #180, opened 2022-01-17
meson: ssl version checking allow version >=1.1.0

sc108-lee:
for lower version of ssl, config will be below
#define CONFIG_OPENSSL
#undef CONFIG_OPENSSL_1
#undef CONFIG_OPENSSL_3

so function nvme_gen_dhchap_key is not defined in c file

#158 @igaw , I assume that you just allow version >=1.1.0, based on your meson changes

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -45,6 +45,7 @@ conf.set('CONFIG_JSONC', json_c_dep.found(), description: 'Is json-c required?')
 
 # Check for OpenSSL availability
 openssl_dep = dependency('openssl',
+                         version: '>=1.1.0',
                          required: get_option('openssl'),
                          fallback : ['openssl', 'openssl_dep'])
 if openssl_dep.found()


sc108-lee:
checking both of allowed version >=1.1.0 & <2.0.0 or >=3.0.0
if is not exist, fallback

src/nvme/linux.c
```diff
@@ -443,23 +443,6 @@ char *nvme_get_path_attr(nvme_path_t p, const char *attr)
 	return nvme_get_attr(nvme_path_get_sysfs_dir(p), attr);
 }
 
-#ifndef CONFIG_OPENSSL
-int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
```
igaw, line 447: The dummy implementation if the library is not compiled with OpenSSL should stay IMO.

eli-schwartz:
Why not just do one `dependency('openssl', version: '>=1.1.0', ...)`

igaw:
Unfortunately, the 3.x API is not backwards compatible. We have two implementation for generating the hash, one using 1.1.x API and one using the 3.x API.

eli-schwartz:
That has absolutely nothing to do with which version you check for.

Doing:
- two unconditional dependency() lookups one after the other, followed by setting some configuration_data() based on matching the openssl_dep.version() attribute,

can just as easily be done with:
-  ***one*** unconditional dependency () lookup followed by setting some configuration_data() based on matching the openssl_dep.version() attribute.

src/nvme/linux.c
sc108-lee: oh I should revert that if fallback fails

sc108-lee:
@eli-schwartz 
original meson says, 2.x is not support for it.
so logic is [ find 3.x -> find 1.x -> fallback 1.1.1 ]

eli-schwartz:
OpenSSL went from 1.1.x directly to 3.x

sc108-lee:
@eli-schwartz ohh.. is it?
that means i just need to add version checking of >=1.1.0
thanks for the checking, I thought ssl-fips-2.~ means 2.x version of it.

eli-schwartz:
I don't think that the OpenSSL FIPS module provided pkg-config files for libssl/libcrypto/openssl, so for the purposes of this dependency lookup it is not relevant.

igaw:
@eli-schwartz I tried to explain why there are CONFIG_OPENSSL_1 and CONFIG_OPENSSL_3 :)

PR #185, opened 2022-01-18
Define ccan as a static library

martin-belanger:
Build ccan as a static library, which gets linked to `libnvme.so`.

NOTE: since ccan depends on `config.h`, I moved the generation of `config.h` 
to the top-level `meson.build` file. It actually makes more sense to generate
`config.h` in the top-level `meson.build` file since all `subdir` projects depend 
on that file.

Refer to the discussion on pull request: https://github.com/linux-nvme/libnvme/pull/179

### DIFF
--- a/ccan/meson.build
+++ b/ccan/meson.build
@@ -6,13 +6,21 @@
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
 
-sources += files([
+sources = [
     'ccan/list/list.c',
     'ccan/str/debug.c',
     'ccan/str/str.c',
-])
+]
 
 if get_option('buildtype') == 'debug'
     add_project_arguments('-DCCAN_LIST_DEBUG=1',  language : ['c', 'cpp'])
     add_project_arguments('-DCCAN_STR_DEBUG=1',  language : ['c', 'cpp'])
 endif
+
+libccan = static_library(
+    'ccan',
+    sources,
+    install: false,
+    include_directories: [incdir, internal_incdir],
+    dependencies: config_dep,
+)
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -25,4 +25,3 @@ executable(
     link_with: libnvme,
     include_directories: [incdir, internal_incdir]
 )
-
--- a/internal/meson.build
+++ b/internal/meson.build
@@ -1,0 +1,21 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# This file is part of libnvme.
+# Copyright (c) 2022 Dell Inc.
+#
+# Authors: Martin Belanger <Martin.Belanger@dell.com>
+#
+# Because libnvme is used as a subproject of nvme-cli,
+# the config.h file must be generated as a private file
+# hidden from nvme-cli.
+
+config_h = configure_file(
+    output: 'config.h',
+    configuration: conf
+)
+
+internal_incdir = include_directories('.')
+
+config_dep = declare_dependency(
+  include_directories : internal_incdir,
+  sources: config_h)
--- a/libnvme/meson.build
+++ b/libnvme/meson.build
@@ -5,11 +5,6 @@
 #
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
-configure_file(
-    output: 'config.h',
-    configuration: conf
-)
-
 want_python = get_option('python')
 if want_python != 'false'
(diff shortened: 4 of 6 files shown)


igaw:
No, you can't move the config file. It needs to be hidden from nvme-cli for example.

martin-belanger:
Ok will move it into its own directory because it needs to be created before any of the other components.

igaw:
FWIW, there is a bit of documentation on this topic. https://mesonbuild.com/Wrap-best-practices-and-tips.html#do-not-put-configh-in-external-search-path

I suggest you rename 'internal_incdir' to 'private_incdir' to match with the directory name.

And there is a `link_with: libccan` missing in libnvme/meson.build

ccan/meson.build
```diff
@@ -6,13 +6,21 @@
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
 
-sources += files([
+sources = [
+    config_h,
```
igaw, line 10: This looks wrong. Why do you need to add the header to the source list?

martin-belanger: I've seen header files added to sources in other projects (i.e. systemd), but after re-reading the meson documentation there is a better to do this. I will send a fix.

With regards to adding `link_with: libccan` to `libnvme/meson.build`, I don't think that's needed because `libccan.a` is linked to `libnvme.so`. Therefore, the ccan code is part of `libnvme.so`. In `libnvme/meson.build` we link with `libnvme.so` (as shown below), therefore we link the ccan code along with `libnvme.so`.

```
    pynvme_clib = python3.extension_module(
        '_nvme',
        pymod_swig[1],
        dependencies : py3_dep,
        include_directories: [incdir, internal_incdir],
        link_with: libnvme,
        install: true,
        subdir: 'libnvme',
    )
```

martin-belanger:
@igaw - I addressed your comments. Note that instead of renaming `internal_incdir` to `private_incdir` to match the directory name, I renamed the directory `private` to `internal`.

ccan/meson.build
igaw: You need to add the static library to python library. This very idea of the static library started because of the missing ccan symbols in the python library (see #174).

```
7/7 [Python] create ctrl object        FAIL            0.04s   exit status 1
>>> PYTHONPATH=/home/wagi/work/libnvme-upstream/.build/libnvme/.. MALLOC_PERTURB_=0 /usr/bin/python3 /home/wagi/work/libnvme-upstream/.build/../libnvme/tests/create-ctrl-obj.py
―――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――
stderr:
Traceback (most recent call last):
  File "/home/wagi/work/libnvme-upstream/.build/../libnvme/tests/create-ctrl-obj.py", line 4, in <module>
    from libnvme import nvme
  File "/home/wagi/work/libnvme-upstream/.build/libnvme/nvme.py", line 13, in <module>
    from . import _nvme
ImportError: /home/wagi/work/libnvme-upstream/.build/libnvme/_nvme.cpython-38-x86_64-linux-gnu.so: undefined symbol: str_strstr
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
```

with 
```patch
@@ -31,7 +31,7 @@ if have_python_support
         pymod_swig[1],
         dependencies : py3_dep,
         include_directories: [incdir, internal_incdir],
-        link_with: libnvme,
+        link_with: [libnvme, libccan],
         install: true,
         subdir: 'libnvme',
     )

```

```
[0/1] Running all tests.
1/7 main                               OK              0.01s
2/7 main                               OK              0.01s
3/7 cpp                                OK              0.01s
4/7 register                           OK              0.01s
5/7 zns                                OK              0.01s
6/7 [Python] import libnvme            OK              0.03s
7/7 [Python] create ctrl object        OK              0.03s
```

martin-belanger: @igaw - Addressed your latest comments. I understand now why you wanted `libccan` explicitly added to `libnvme/meson.build`. It's to solve the missing symbol when `buildtype=debug`. By explicitly linking `libccan` we're basically making all ccan symbols public.

igaw: The symbols from libccan should not be exported. For libnvme we maintain the libnvme.map. But not for the python library. This needs to be addressed. Ideally in a more usable way than libnvme.map.

Issue #181, opened 2022-01-17
build warnings on 32 bit architectures


igaw:
Got some buidl warnings for 32 bit architectures.

https://build.opensuse.org/build/home:wagi:branches:Base:System/openSUSE_Factory/i586/libnvme/_log

```
[   45s] In file included from ../src/nvme/fabrics.c:36:
[   45s] ../src/nvme/fabrics.c: In function ‘nvmf_get_discovery_log’:
[   45s] ../src/nvme/fabrics.c:781:37: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 5 has type ‘uint64_t’ {aka ‘long long unsigned int’} [-Wformat=]
[   45s]   781 |                 nvme_msg(LOG_DEBUG, "%s: discover genctr %lu, retry\n",
[   45s]       |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[   45s]   782 |                          name, genctr);
[   45s]       |                                ~~~~~~
[   45s]       |                                |
[   45s]       |                                uint64_t {aka long long unsigned int}
[   45s] ../src/nvme/log.h:36:36: note: in definition of macro ‘nvme_msg’
[   45s]    36 |                                    format, ##__VA_ARGS__);              \
[   45s]       |                                    ^~~~~~
[   45s] ../src/nvme/fabrics.c:781:60: note: format string is defined here
[   45s]   781 |                 nvme_msg(LOG_DEBUG, "%s: discover genctr %lu, retry\n",
[   45s]       |                                                          ~~^
[   45s]       |                                                            |
[   45s]       |                                                            long unsigned int
[   45s]       |                                                          %llu
[   45s] In file included from ../src/nvme/fabrics.c:36:
[   45s] ../src/nvme/fabrics.c:798:36: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 5 has type ‘uint64_t’ {aka ‘long long unsigned int’} [-Wformat=]
[   45s]   798 |                 nvme_msg(LOG_INFO, "%s: could only fetch %lu of %lu records\n",
[   45s]       |                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[   45s]   799 |                          name, numrec, le64_to_cpu(log->numrec));
[   45s]       |                                ~~~~~~
[   45s]       |                                |
[   45s]       |                                uint64_t {aka long long unsigned int}
[...]

PR #186, opened 2022-01-19
src: Use PRIu64 format specifier for uint64 types

igaw:
When compiling for 32bit architectures the compiler is complaining
about the wrong format specifier for uint64. Use the the PRIu64 format
specifier in this case which is supported from 32bit and 64bit.

Fixes: #181


hreinecke (approved):
Okay.

Issue #187, opened 2022-01-19
Ability to have error messages go to different file descriptor or user supplied callback


tasleson:
I think it would be good to have a library user provide a file descriptor for logging or a call back as suggested here: https://github.com/linux-nvme/libnvme/pull/160#issuecomment-1010314146

Having all these messages to written to `stderr` isn't ideal for library users.

tbzatek:
Well, the messages currently logged on `stderr` are more customized than the `nvme_errno_to_string()` translation but I find the error codes useful enough at the current state.

Perhaps we should set `nvme_log_level` by default to `-1` and let the clients (e.g. `nvme-cli`) set their own console logging level. Not every (consumer) library has the luxury of providing initialisation functions that can set up things in the nested (`libnvme`) libraries.

hreinecke:
Please check PR #202. I've added a 'nvme_root_t' argument to nvme_msg, which now holds the filedescriptor where the messages should be sent to.
Applications can then set the filedescriptor to whatever they want.
Would that work for you?

tasleson:
> Would that work for you?

Yes, this works thanks

igaw:
PR #202 has been merged. Closing this issue.

PR #189, opened 2022-01-20
POC: Add logging function callback

tasleson:
I spent some time today exploring if have a call back function which was embedded in the `nvme_root_t` to prevent global data would be possible.  This is rough, and incomplete, and not even run once for testing.

I created a macro `nvme_msg_n` which takes an additional argument which is the `nvme_root_t` pointer.  In some cases its use is pretty clean:

```c
nvme_msg_n(r, LOG_ERR, "Failed to write to %s, %s\n", ...
```

In other cases, kind of ugly and I'm not even sure if all the embedded pointers are correctly setup for the context.

```c
nvme_msg_n(c->s->h->r, LOG_ERR, "failed to resolve host %s info\n", c->traddr);
```

In some other cases we have nothing existing to get a handle to the `nvme_root_t`, so I added an argument to the function.  In other cases I just removed the logging and returned an error code.

At the moment I'm not sure how to make this cleaner.  Thoughts, suggestions?

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,7 +1,5 @@
 LIBNVME_1_0 {
 	global:
-		__nvme_get_log_page;
-		__nvme_msg;
 		nvme_admin_passthru64;
 		nvme_admin_passthru;
 		nvme_attach_ns;
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -282,7 +282,7 @@ static int inet6_pton(const char *src, uint16_t port,
 
 	char  *tmp = strdup(src);
 	if (!tmp)
-		nvme_msg(LOG_ERR, "cannot copy: %s\n", src);
+		return -ENOMEM;		// Just logging here looks like a bug
 
 	const char *scope = NULL;
 	char *p = strchr(tmp, SCOPE_DELIMITER);
@@ -297,8 +297,6 @@ static int inet6_pton(const char *src, uint16_t port,
 	if (IN6_IS_ADDR_LINKLOCAL(&addr6->sin6_addr) && scope) {
 		addr6->sin6_scope_id = if_nametoindex(scope);
 		if (addr6->sin6_scope_id == 0) {
-			nvme_msg(LOG_ERR,
-				 "can't find iface index for: %s (%m)\n", scope);
 			goto free_tmp;
 		}
 	}
@@ -330,7 +328,6 @@ static int inet_pton_with_scope(int af, const char *src, const char * trsvcid,
 		unsigned long long tmp = strtoull(trsvcid, NULL, 0);
 		port = (uint16_t)tmp;
 		if (tmp != port) {
-			nvme_msg(LOG_ERR, "trsvcid out of range: %s\n", trsvcid);
 			return -ERANGE;
 		}
 	} else {
@@ -350,7 +347,7 @@ static int inet_pton_with_scope(int af, const char *src, const char * trsvcid,
 			ret = inet6_pton(src, port, addr);
 		break;
 	default:
-		nvme_msg(LOG_ERR, "unexpected address family %d\n", af);
+		ret = -EINVAL;
 	}
 
 	return ret;
@@ -378,7 +375,7 @@ static int hostname2traddr(nvme_ctrl_t c)
 
 	ret = getaddrinfo(c->traddr, NULL, &hints, &host_info);
 	if (ret) {
-		nvme_msg(LOG_ERR, "failed to resolve host %s info\n", c->traddr);
+		nvme_msg_n(c->s->h->r, LOG_ERR, "failed to resolve host %s info\n", c->traddr);
 		return -ENVME_CONNECT_RESOLVE;
 	}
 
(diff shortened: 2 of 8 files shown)


tasleson:
Could resolve: https://github.com/linux-nvme/libnvme/issues/187

src/nvme/fabrics.c
```diff
@@ -282,7 +282,7 @@ static int inet6_pton(const char *src, uint16_t port,
 
 	char  *tmp = strdup(src);
 	if (!tmp)
-		nvme_msg(LOG_ERR, "cannot copy: %s\n", src);
+		return -ENOMEM;		// Just logging here looks like a bug
```
igaw, line 285: Yes, this should just return the error code.

src/nvme/fabrics.c
```diff
@@ -350,7 +347,7 @@ static int inet_pton_with_scope(int af, const char *src, const char * trsvcid,
 			ret = inet6_pton(src, port, addr);
 		break;
 	default:
-		nvme_msg(LOG_ERR, "unexpected address family %d\n", af);
+		ret = -EINVAL;
```
igaw, line 350: I don't mind these error handling cleanups in inet*(). But it would be better to go into a separate patch.

src/nvme/fabrics.c
```diff
@@ -571,13 +568,13 @@ int nvmf_add_ctrl_opts(nvme_ctrl_t c, struct nvme_fabrics_config *cfg)
...
 
-	ret = __nvmf_add_ctrl(argstr);
+	ret = __nvmf_add_ctrl(c, argstr);
 	free(argstr);
 	if (ret < 0) {
 		errno = -ret;
 		ret = -1;
 	} else {
-		nvme_msg(LOG_INFO, "nvme%d: ctrl connected\n", ret);
+		nvme_msg_n(c->s->h->r, LOG_INFO, "nvme%d: ctrl connected\n", ret);
```
igaw, line 577: Indeed, looks a bit ugly. Though I don't have a good idea how to do it differently.

src/libnvme.map
```diff
@@ -1,7 +1,5 @@
 LIBNVME_1_0 {
 	global:
-		__nvme_get_log_page;
```
igaw, line 3: You can't remove this one. This is a public function (even documented) see https://github.com/linux-nvme/libnvme/blob/5aea0216cfb2410f9cdb7e2e25d04fbb4022a317/src/nvme/linux.h#L83

Bad name. We should rename it. Though I don't have a better name yet.

src/libnvme.map
```diff
@@ -1,7 +1,5 @@
...
-		__nvme_msg;
```
igaw, line 4: Can't remove this either at this point. The nvme-cli uses the nvme_msg() in fabrics.c. I think we should remove the use of nvme_msg in fabrics.c. nvme-cli should use it's own error logging macro.

igaw (changes requested):

tbzatek:
Interesting concept, however to be really useful I would need a context passed to `log_fn` - a custom pointer (`user_data`) set along with `log_fn`, passed back within the callback. From a library user perspective getting no context means I would still need to use TLS for temporary storage and in the case of `nvme_root_t` I would need an internal translation table. Setting logging function per-`nvme_root_t` could only help to differentiate between logging codepaths, but connection to my internal object would be still missing.

tasleson:
@tbzatek 
> Interesting concept, however to be really useful I would need a context passed to `log_fn` - a custom pointer (`user_data`) set along with `log_fn`, passed back within the callback. From a library user perspective getting no context means I would still need to use TLS for temporary storage and in the case of `nvme_root_t` I would need an internal translation table. Setting logging function per-`nvme_root_t` could only help to differentiate between logging codepaths, but connection to my internal object would be still missing.

Your original comment on this https://github.com/linux-nvme/libnvme/pull/160#issuecomment-1010314146 didn't have user supplied data.  We could add a user supplied data pointer when registering the callback and store it in the `nvme_root_t` structure.  I'm assuming each thread of execution is retrieving their own `nvme_root_t` correct?  Otherwise the library caller could embed the callback in a structure and leverage things like `container_of` approach to associate the callback with some other data for correlation.

tbzatek:
Yes, my original proposal was about general logging function with no context as everything was on a global level. But now since you would be operating on a local `nvme_root_t` instance, supplying an additional user data pointer when registering a particular logging callback shouldn't create much more mess and would greatly improve usability.

>  I'm assuming each thread of execution is retrieving their own nvme_root_t correct?

Not necessarily but I understood it should be a good practice. However heavy async app may create `nvme_root_t` somewhere and push an actual operation over it to a worker thread pool elsewhere. I expect the whole codepath should operate on a private `nvme_root_t` instance though.
--> And this should be documented somewhere...

tasleson:
> Not necessarily but I understood it should be a good practice. However heavy async app may create `nvme_root_t` somewhere and push an actual operation over it to a worker thread pool elsewhere. I expect the whole codepath should operate on a private `nvme_root_t` instance though.

I believe we are in agreement that a single `root_nvme_t` shouldn't be used concurrently from two different threads.  Obtaining a handle in one thread, passing it to another for processing would adhere to this.

If/when I update this PR I'll include a user supplied data pointer.  I can see how that would be beneficial in a multi-threaded app.

src/libnvme.map
igaw: I removed the nvme_msg() calls in nvme-cli. So we can remove the __nvme_msg from the export map.

hreinecke:
Can you check PR #202 ? That does most what you want; if needed we could even add a logging function callback to nvme_root_t, but I guess with being able to redirect the message logging to a different filedescriptor we should be mostly there.

igaw:
Let's close this PR as we have some good progress in #202.

PR #190, opened 2022-01-21
Display failed namespaces

hreinecke:
When a namespace is reconnecting or otherwise failed we cannot do I/O, and so the namespace initialization will fail. This patch adds a fallback to read (some) information from sysfs such that we still can display the namespace eg in 'nvme list'.

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -70,9 +70,14 @@ int nvme_ns_rescan(int fd)
 
 int nvme_get_nsid(int fd, __u32 *nsid)
 {
+	int __nsid;
+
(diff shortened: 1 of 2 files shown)


igaw:
Still this:

```
nvme-subsys16 - NQN=nqn.2014-08.org.nvmexpress:NVMf:uuid:de63429f-50a4-4e03-ade6-0be27b75be77
\
 +- (null) tcp traddr=10.161.8.24,trsvcid=4421 (null)
 +- nvme16 tcp traddr=10.161.8.24,trsvcid=4420 connecting
```

igaw:
This got superseded by #194

PR #192, opened 2022-01-21
ioctl: Initialize all members in nvme_get_log_args

igaw:
Commit 781b9105cc43 ("nvme: Add Media Unit Status log page(LID:
0x10)") introduces a new member in nvme_get_log_args. Older compilers
are not happy when we don't initialize all members in the inline
functions.

In file included from ../subprojects/libnvme/src/libnvme.h:19:0,
                 from ../subprojects/libnvme/test/cpp.cc:10:
../subprojects/libnvme/src/nvme/ioctl.h: In function ‘int nvme_get_log_cmd_effects(int, nvme_csi, nvme_cmd_effects_log*)’:
../subprojects/libnvme/src/nvme/ioctl.h:1328:2: sorry, unimplemented: non-trivial designated initializers not supported
  };


igaw:
FWIW, updated nvme-cli depenency to libnvme accordingly.

PR #193, opened 2022-01-21
ioctl: fixup nvme_get_log_args initialization

hreinecke:
Some (older) gcc compiler require the order of initialisation for
structure to match the order of declaration; not doing so results in
a build failure 'Sorry, non-trivial designated initializers are not supported.'


hreinecke:
Ah. Already fixed.

igaw:
Yes, run into the same problem :)

Issue #148, opened 2021-12-16
Define MASK, SHIFT to use NVME_GET


sc108-lee:
I would like to do some modify based on usage of NVME_GET
I Just wrote only for nvme_id_nsfeat currently, Example below!
Any opinion?

From 519337c64316d8805b3d988f1b475591347f22f2 Mon Sep 17 00:00:00 2001
From: Steven Seungcheol Lee <sc108.lee@samsung.com>
Date: Thu, 16 Dec 2021 17:17:24 +0900
Subject: [PATCH] types: Define enum to use NVME_GET define

Signed-off-by: Steven Seungcheol Lee <sc108.lee@samsung.com>
---
 doc/libnvme.rst               | 12 ++++++------
 doc/man/enum nvme_id_nsfeat.2 | 22 +++++++++++-----------
 src/nvme/types.h              | 25 +++++++++++++++----------
 3 files changed, 32 insertions(+), 27 deletions(-)

diff --git a/src/nvme/types.h b/src/nvme/types.h
index a9cf415..8a08770 100644
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -1668,32 +1668,37 @@ struct nvme_id_ns {
 

 enum nvme_id_nsfeat {
-	NVME_NS_FEAT_THIN		= 1 << 0,
-	NVME_NS_FEAT_NATOMIC		= 1 << 1,
-	NVME_NS_FEAT_DULBE		= 1 << 2,
-	NVME_NS_FEAT_ID_REUSE		= 1 << 3,
-	NVME_NS_FEAT_IO_OPT		= 1 << 4,
+	NVME_NS_FEAT_THINP_SHIFT	= 0,
+	NVME_NS_FEAT_THINP_MASK		= 0x1,
+	NVME_NS_FEAT_NSABP_SHIFT	= 1,
+	NVME_NS_FEAT_NSABP_MASK		= 0x1,
+	NVME_NS_FEAT_DAE_SHIFT		= 2,
+	NVME_NS_FEAT_DAE_MASK		= 0x1,
+	NVME_NS_FEAT_UIDREUSE_SHIFT	= 3,
+	NVME_NS_FEAT_UIDREUSE_MASK	= 0x1,
+	NVME_NS_FEAT_OPTERF_SHIFT	= 4,
+	NVME_NS_FEAT_OPTERF_MASK	= 0x1,
 };
 --

igaw:
[...]

PR #197, opened 2022-01-24
[WIP] types: Introduce getter macros

igaw:
I started to work on #148. This will take a while to go through all enums add those missing MASK/SHIFT enums.


igaw:
FWIW, @hreinecke this is going to be a big effort... I think I work first on the errno return code thing.

igaw:
After a discussion with Hannes, we decided to postpone this plan. The effort to fixup all definition is a considerable and there is a good chance to introduce a lot of regressions. This is something for v2.

PR #198, opened 2022-01-25
TP8010 support

martin-belanger:
TP8010 defines the Central Discovery Controller (CDC). To tell a CDC apart from a DDC (Direct Discovery Controller), the Identify response now carries a Discovery Controller Type (dctype) attribute.

TP8010 also defines a new Discovery Information Management (DIM) PDU which is used to perform explicit registration with CDCs and DDCs. This new PDU carries extra information such as the Entity Name (ENAME), the Entity Version (EVER), the Symbolic Name, etc.

This adds support for TP8010 as follows:
- Add the dctype to struct nvme_id_ctrl
- Add the "register" argument - This is used to tell the kernel that the user wants to perform explicit registration using the DIM PDU. Without this, the driver will operate as it did before.
- Add the "hostsymname" argument - This is used to specify an optional Symbolic Name. The Symbolic name is configured with the file `"/etc/nvme/hostsymname"`.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -37,12 +37,15 @@
 #include "log.h"
 #include "private.h"
 
+#define NVMF_HOSTSYMNAME_SIZE	256
+
 #define NVMF_HOSTID_SIZE	37
 #define UUID_SIZE		37  /* 1b4e28ba-2fa1-11d2-883f-0016d3cca427 + \0 */
 
 const char *nvmf_dev = "/dev/nvme-fabrics";
 const char *nvmf_hostnqn_file = "/etc/nvme/hostnqn";
 const char *nvmf_hostid_file = "/etc/nvme/hostid";
+const char *nvmf_hostsymname_file = "/etc/nvme/hostsymname";
 
 const char *arg_str(const char * const *strings,
 		size_t array_size, size_t idx)
@@ -418,7 +421,7 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 {
 	struct nvme_fabrics_config *cfg = nvme_ctrl_get_config(c);
 	const char *transport = nvme_ctrl_get_transport(c);
-	const char *hostnqn, *hostid, *hostkey, *ctrlkey;
+	const char *hostnqn, *hostid, *hostkey, *ctrlkey, *hostsymname;
 	bool discover = false, discovery_nqn = false;
 
 	if (!transport) {
@@ -447,6 +450,7 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 	hostnqn = nvme_host_get_hostnqn(h);
 	hostid = nvme_host_get_hostid(h);
 	hostkey = nvme_host_get_dhchap_key(h);
+	hostsymname = nvme_host_get_symname(h);
 	ctrlkey = nvme_ctrl_get_dhchap_key(c);
 	if (add_argument(argstr, "transport", transport) ||
 	    add_argument(argstr, "traddr",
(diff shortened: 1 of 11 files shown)


src/nvme/fabrics.c
```diff
@@ -459,8 +463,11 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
 			 nvme_ctrl_get_trsvcid(c)) ||
 	    (hostnqn && add_argument(argstr, "hostnqn", hostnqn)) ||
 	    (hostid && add_argument(argstr, "hostid", hostid)) ||
+	    (hostsymname && add_argument(argstr, "hostsymname", hostsymname)) ||
```
hreinecke, line 466: Can you please split this off into several parts?
I would like to see:
a) a patch for adding the 'hostsymname' option
b) a patch for adding explicit registration support (ie command description and tooling) to libnvme/nvme-cli such that users can send / test the commands from userspace
c) a patch for adding the 'register' option to nvme-cli, once the discussion on linux-nvme has been resolved.

src/nvme/fabrics.c
```diff
@@ -992,3 +999,8 @@ char *nvmf_hostid_from_file()
 {
 	return nvmf_read_file(nvmf_hostid_file, NVMF_HOSTID_SIZE);
 }
+
+char *nvmf_hostsymname_from_file()
+{
+	return nvmf_read_file(nvmf_hostsymname_file, NVMF_HOSTSYMNAME_SIZE);
```
hreinecke, line 1005: Not sure if I want to have yet anouther file in /etc/nvme; shouldn't we rather use the json configuration here?

src/nvme/fabrics.h
```diff
@@ -211,6 +211,16 @@ char *nvmf_hostnqn_from_file();
...
 
+/**
+ * nvmf_hostsymname_from_file() - Read the host symbolic name from the
+ * config default location in /etc/nvme.
+ *
+ * Return: The symbolic name or NULL if no symbolic name has been
+ * configured. The caller is responsible fro freeing the memory allocated
+ * by this function (if any).
+ */
+char * nvmf_hostsymname_from_file();
```
hreinecke, line 222: Some comment as above.

src/nvme/private.h
```diff
@@ -115,6 +116,7 @@ struct nvme_host {
 	char *hostnqn;
 	char *hostid;
 	char *dhchap_key;
+	char *symname;
```
hreinecke, line 119: Please name is 'hostsymname' to be consistent with prior usage.

src/nvme/tree.c
```diff
@@ -62,6 +62,9 @@ nvme_host_t nvme_default_host(nvme_root_t r)
 	hostid = nvmf_hostid_from_file();
 
 	h = nvme_lookup_host(r, hostnqn, hostid);
+
+	nvme_host_set_symname(h, nvmf_hostsymname_from_file(), true);
```
hreinecke, line 66: See above; we should rely on the json configuration here.

src/nvme/tree.c
```diff
@@ -376,6 +394,7 @@ static void __nvme_free_host(struct nvme_host *h)
 		free(h->hostid);
 	if (h->dhchap_key)
 		free(h->dhchap_key);
+	nvme_host_set_symname(h, NULL, true);
```
hreinecke, line 397: Just do a 'free' here; no need for an additional function.

src/nvme/types.h
```diff
@@ -5316,6 +5334,7 @@ enum nvme_admin_opcode {
 	nvme_admin_security_recv	= 0x82,
 	nvme_admin_sanitize_nvm		= 0x84,
 	nvme_admin_get_lba_status	= 0x86,
+	nvme_admin_discovery_info_mgmt	= 0x21,
```
hreinecke, line 5337: As mentioned above, please add the definitions for the command structure, too.

hreinecke (changes requested):

src/nvme/fabrics.c
igaw: Good point. JSON tends to be way simpler to parse.

martin-belanger: We're already asking people ton configure the hostid and hostnqn as files under /etc/nvme. So for consistency it makes sense to have the symbolic name added in the same manner. If you prefer that it be done through the json parser I have no objection.

martin-belanger:
Considering the reviews I've receiving on the kernel, I will need to make quite a bit of changes. I'm closing this pull request and will submit a new one later this week.

PR #199, opened 2022-01-27
nvme: Add Supported Capacity Configuration List log page(LID: 0x11)

Arunpandian15:

### DIFF
--- a/src/nvme/ioctl.h
+++ b/src/nvme/ioctl.h
@@ -1798,6 +1798,38 @@ static inline int nvme_get_log_media_unit_stat(int fd, __u16 domid,
 }
 
 /**
+ * nvme_get_log_support_cap_config_list() -
+ * @fd:		File descriptor of nvme device
+ * @domid:	Domain Identifier selection, if supported
+ *
+ * Return: The nvme command status if a response was received (see
+ * &enum nvme_status_field) or -1 with errno set otherwise
+ */
+static inline int nvme_get_log_support_cap_config_list(int fd, __u16 domid,
+			struct nvme_supported_cap_config_list_log *cap)
+{
+	struct nvme_get_log_args args = {
+		.lpo = 0,
+		.result = NULL,
+		.log = cap,
+		.args_size = sizeof(args),
+		.fd = fd,
+		.timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
+		.lid = NVME_LOG_LID_SUPPORTED_CAP_CONFIG_LIST,
+		.len = sizeof(*cap),
+		.nsid = NVME_NSID_NONE,
+		.csi = NVME_CSI_NVM,
+		.lsi = NVME_LOG_LSI_NONE,
+		.domid = domid,
+		.lsp = NVME_LOG_LSP_NONE,
+		.uuidx = NVME_UUID_NONE,
+		.rae = false,
+		.ot = false,
+	};
+	return nvme_get_log(&args);
+}
+
+/**
  * nvme_get_log_reservation() -
  * @fd:		File descriptor of nvme device
  * @rae:	Retain asynchronous events
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3544,6 +3544,101 @@ struct nvme_media_unit_stat_log {
 };
 
 /**
+ * struct nvme_media_unit_config_desc -
+ * @muid: Media Unit Identifier
+ * @mudl: Media Unit Descriptor Length
+ *
+ * Media Unit Configuration Descriptor
+ * Structure Definitions
+ */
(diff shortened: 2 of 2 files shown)


src/nvme/ioctl.h
```diff
@@ -1789,6 +1789,39 @@ static inline int nvme_get_log_media_unit_stat(int fd, __u16 domid,
...
+		.lsp = NVME_LOG_LSP_NONE,
+		.domid = domid,
+		.uuidx = NVME_UUID_NONE,
+		.rae = false,
+		.ot = false,
+	};
+	return nvme_get_log(&args);
+}
+
+
```
igaw, line 1824: Please avoid this double new line. I see there are already a few of those in this section. Please send a separate cleanup patch.

src/nvme/types.h
```diff
@@ -3319,6 +3319,92 @@ struct nvme_media_unit_stat_log {
...
+struct nvme_end_grp_config_desc {
+	__le16	endgid;
+	__le16	cap_adj_factor;
+	__u8	rsvd4[12];
+	__u8	tegcap[16];
+	__u8	segcap[16];
+	__u8	end_est[16];
+	__u8	rsvd64[16];
+	__le16	egsets;
+	__le16	nvmsetid[256];
```
igaw, line 3373: Is this member field fixed size to 256 bytes? The NVME spec I am looking at says it's variable size 
```
(EGSETS*2)+81 :
(EGSETS*2)+80
```
but maybe I am looking at an outdated spec (Revision 2.0a July 23rd, 2021)

src/nvme/types.h
```diff
@@ -3319,6 +3319,92 @@ struct nvme_media_unit_stat_log {
...
+/**
+ * struct nvme_cap_config_desc -
+ * @cap_config_id:	Capacity Configuration Identifier
+ * @domainid:		Domain Identifier
+ * @egcn:		Number Endurance Group Configuration
+ * 			Descriptors
+ * 
+ * Capacity Configuration structure definitions
+ */
+struct nvme_capacity_config_desc {
```
igaw, line 3387: This is even missing in the spec I am looking at. Looks like I am reading the wrong spec. Can you point me to right one?

src/nvme/ioctl.h
Arunpandian15: @igaw 
There are many places in this section double new line is being used, taken that as reference. Will change it to single new line and give you separate patch

src/nvme/types.h
Arunpandian15: You could find the Capacity Configuration Descriptor in Figure 251(Page-235) in NVMe 2.0 spec

src/nvme/types.h
Arunpandian15: @igaw 
It's variable size as you mentioned. Maximum it can hold 65535. Will split the structure and incorporate variable size. Thanks!

src/nvme/ioctl.h
igaw: I know the file is a bit inconsistent. So the best thing would do have just one separate cleanup patch.

src/nvme/types.h
igaw: Alright, I was already wondering if I am looking at the wrong spec :)

src/nvme/types.h
igaw: Ah yes. I got confused by Figure 253 (Channel Configuration Descriptor) which missing from this patch. Could you add it to make this section complete?

Arunpandian15: Channel configuration descriptor is already part of this patch.
struct nvme_channel_config_desc {
	__le16	chanid;
	__le16	chmus;
	struct nvme_media_unit_config_desc mu_config_desc[];
};
Could you please check

igaw: Right you are! Sorry for the noise.

igaw (changes requested):

src/nvme/ioctl.h
```diff
@@ -1789,6 +1789,39 @@ static inline int nvme_get_log_media_unit_stat(int fd, __u16 domid,
...
+		.log = cap,
+		.len = sizeof(*cap),
+		.nsid = NVME_NSID_NONE,
+		.csi = NVME_CSI_NVM,
+		.lsi = NVME_LOG_LSI_NONE,
+		.lsp = NVME_LOG_LSP_NONE,
+		.domid = domid,
+		.uuidx = NVME_UUID_NONE,
+		.rae = false,
+		.ot = false,
```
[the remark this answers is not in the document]
Arunpandian15: Updated, Thanks!

igaw:
Fixed a few whitespace damages.

PR #200, opened 2022-01-28
meson: libnvme build as static lib

sc108-lee:
libnvme usually wrapped with commit hash (eg nvme-cli)
since they might updated at the same time when new spec comes in
it could be override via meson .build --default-library=shared

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,6 +15,7 @@ project(
         'warning_level=1',
         'buildtype=release',
         'prefix=/usr',
+        'default_library=static',
     ]
 )
 


igaw:
We should keep this to the default shared library. This library is should be considered as standalone. There is not just nvme-cli, e.g nvme-stas. I expect more users of the library. 

What is the problem you want to solve here? When building nvme-cli meson is taking care which libnvme version you build against and as you pointed out you can just say 'meson .build --default-library=static' in nvme-cli.

hreinecke:
Traditionally, the build process would build _both_, ie shared _and_ static library.
That comes in handy when you are developing things, as then you don't have to worry LD_LIBRARY_PATH etc when testing the application.
So there is a value to have a static library; however, documentation would serve as well.

igaw:
If you build with meson's wrap thing, it takes care with rpath. You really don't have to change anything here for getting this working. And you can build both library types by --default-library=both.

hreinecke:
And indeed, that's what I did; updated 'README.md' to document how to build a static library.

sc108-lee:
I was trying to build this as static to not build release of libnvme whenever main project released.
But After you guys talked about, the package manager should be known about it.
so leave it as you guys said

PR #204, opened 2022-01-31
ioctl: Fix struct nvme_get_log_args memberm alignment

igaw:
Commit 781b9105cc43 ("nvme: Add Media Unit Status log page(LID:
0x10)") placed the domid member between two u8 members. The access
domid member would not be 16 bit aligned which is caused problems on
certain architectures. Let's fix the aligemnt by moving domid before
lsp. This give us following layout

```
struct nvme_get_log_args {
        int                        args_size;            /*     0     4 */
        int                        fd;                   /*     4     4 */
        __u32 *                    result;               /*     8     8 */
        enum nvme_cmd_get_log_lid  lid;                  /*    20     4 */
        __u64                      lpo;                  /*    24     8 */
        void *                     log;                  /*    32     8 */
        __u32                      len;                  /*    40     4 */
        __u32                      nsid;                 /*    44     4 */
        enum nvme_csi              csi;                  /*    48     4 */
        __u16                      lsi;                  /*    52     2 */
        __u16                      domid;                /*    54     2 */
        __u8                       lsp;                  /*    56     1 */
        __u8                       uuidx;                /*    57     1 */
        _Bool                      rae;                  /*    58     1 */
        _Bool                      ot;                   /*    59     1 */

        /* size: 64, cachelines: 1, members: 16 */
        /* padding: 4 */
} __attribute__((__aligned__(8)));
```


igaw:
BTW, I am pondering on if we should move all proper u64 up, followed by the pointer types, u32, u16, u8 members.

igaw:
It turns out my commit message is not completely true: https://wiki.debian.org/ArchitectureSpecificsMemo#Alignment
That means except for SPARC there is no problem to expected from a misalignment in packed structs. Don't know how relevant this is though. Is SPARC still alive?

hreinecke:
Yeah, I had been wondering about this, too. We should be fixing it up, SPARC or not SPARC :-)

igaw:
Okay let me spin a new fixup series for this.

Issue #208, opened 2022-02-01
nvme_read_config should return error code


igaw:
Parsing a JSON file can fail but the caller has no clue when this happens. Thus this function should return an error code

hreinecke:
PR merged.

PR #211, opened 2022-02-01
Return json parsing errors

hreinecke:
Parsing the JSON configuration file might fail, so we should be signalling that to the caller.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -145,6 +145,12 @@ int nvme_read_config(nvme_root_t r, const char *config_file)
 		err = json_read_config(r, config_file);
 		if (!err)
 			r->config_file = strdup(config_file);
+		/*
+		 * The json configuration file is optional,
+		 * so ignore errors when opening the file.
+		 */
+		if (err < 0 && errno != EPROTO)
+			err = 0;
 	}
 #else
 	errno = ENOTSUP;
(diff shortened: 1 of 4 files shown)


hreinecke:
Resolves issue #208.

src/nvme/tree.c
```diff
@@ -145,6 +145,12 @@ int nvme_read_config(nvme_root_t r, const char *config_file)
 		err = json_read_config(r, config_file);
 		if (!err)
 			r->config_file = strdup(config_file);
+		/*
+		 * The json configuration file is optional,
+		 * so ignore errors when opening the file.
```
igaw, line 150: I am slightly confused by this comment. The only error json_read_config() we will propegate here is when we can't open the file. When the parser fails it will we always be EPROTO. Shouldn't this be the other way round?

igaw (commented):
Looks good just my question if the logic is inverted with the error filtering

src/nvme/tree.c
hreinecke: But it is?
~~~
+               if (err < 0 && errno != EPROTO)
+                       err = 0;
~~~
IE if 'err' == -1 an error occurred, _and_ if the returned errno is not EPROTO we reset 'err' to 0, ie we're blanking out any errors when opening the file.
As the file is optional and opening errors are okay.

igaw: You are right. Got confused again :)

PR #212, opened 2022-02-01
[libnvme] Add support for retrieving Telemetry Log Page data area 4.  

jeffreyalien:
To add support for telemetry log page data area 4, the size ptr and data area parms where added to functions:   nvme_get_telemetry_log, nvme_get_ctrl_telemetry, nvme_get_host_telemetry, and nvme_get_new_host_telemetry.  

There is a corresponding change to the nvme cli code that is dependent on this pull request.  

Signed-off-by: Jeff Lien <jeff.lien@wdc.com>


igaw:
Is this PR replaced by #213?

PR #214, opened 2022-02-02
Sanitize kernel-doc comments

hreinecke:
Run kernel-doc over the source files and update comments to keep sphinx happy.


igaw:
I think we could add a script for regenerating the doc as we have in nvme-cli. And add a check list what to do before a release. I am going to forget it for sure.

PR #215, opened 2022-02-02
Scripts for regenerating documentation

hreinecke:
Import the 'kernel-doc' script from the linux kernel sources and add scripts to regenerate libnvme.rst and the manual pages.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -10,6 +10,8 @@ Welcome to libnvme's documentation!
    :maxdepth: 2
    :caption: Contents:
 
+   index
+   libnvme
 
 
 Indices and tables
--- a/doc/kernel-doc
+++ b/doc/kernel-doc
@@ -1,0 +1,2523 @@
+#!/usr/bin/env perl
+# SPDX-License-Identifier: GPL-2.0
+
+use warnings;
+use strict;
+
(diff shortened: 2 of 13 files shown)


igaw:
Any particular reason why the scripts are in internal? We have a `make.bat` in the doc dir (do we need it?). Also I was thinking on hooking it up as new conditional target to meson (in case this is supported. need to study the docs)

hreinecke:
No particular reason. Just wanting to share what I've been able to find out; getting there was painful enough, and others shouldn't be made to suffer the same way.
Of course having it as a meson target is preferred. And while we're at it, the current 'meson -Dman=true' thingie doesn't actually generate the man pages; it just pushes everything from libnvme.rst into libnvme.2. And even that is in the wrong location.
But that's a different story :-)

igaw:
When the user doesn't tell meson to build the documentation via -Dman=true it installs the pre compiled ones? Would this work? Or do we an option to install only the documentation on request?

hreinecke:
Yes, that was the plan. With normal build the pre-compiled manpages get installed, and with 'meson -Dman=true' the manpages get regenerated.

hreinecke:
Only that 'meson -Dman=true' doesn't do that, cf issue #218.

igaw:
I've just moved the scripts to the doc dir.

Issue #154, opened 2021-12-21
nvme_ctrl_alloc() can't handle duplicate connections


tbzatek:
Okay, I admit my use case is a little weird and might not reflect real world scenario: loop transport, two (duplicate) connections, resulting in a single subsystem and two controllers in sysfs. The libnvme scanning logic would however only find one controller.

https://github.com/linux-nvme/libnvme/blob/ddd725e780625d25dd199dc367dda5503e13c685/src/nvme/tree.c#L1270-L1271

Looking at how `nvme_lookup_ctrl()` is called and how are the keys matched inside I see no way to distinguish seemingly duplicate controller entries, just having different name and sysfs path.

This is a broader issue, posting now in light of ongoing tree scanning logic rewrite in #108 to get some consideration.

----

```
# old/nvme list -v
NVM Express Subsystems

Subsystem        Subsystem-NQN                                                                                    Controllers
---------------- ------------------------------------------------------------------------------------------------ ----------------
nvme-subsys0     test_subnqn                                                                                      nvme0, nvme1

NVM Express Controllers

Device   SN                   MN                                       FR       TxPort Address        Subsystem    Namespaces      
-------- -------------------- ---------------------------------------- -------- ------ -------------- ------------ ----------------
nvme0    cce1b445357e708d9d62 Linux                                    5.14.17- loop                  nvme-subsys0 nvme0n1, nvme0n2, nvme0n3
nvme1    cce1b445357e708d9d62 Linux                                    5.14.17- loop                  nvme-subsys0 nvme0n1, nvme0n2, nvme0n3

NVM Express Namespaces

Device       NSID     Usage                      Format           Controllers     
------------ -------- -------------------------- ---------------- ----------------
nvme0n1      1          1.07  GB /   1.07  GB    512   B +  0 B   nvme0, nvme1
nvme0n2      2          1.07  GB /   1.07  GB    512   B +  0 B   nvme0, nvme1
nvme0n3      3          1.07  GB /   1.07  GB    512   B +  0 B   nvme0, nvme1
```

```
# git-master/nvme list -v
Subsystem        Subsystem-NQN                                                                                    Controllers
---------------- ------------------------------------------------------------------------------------------------ ----------------
nvme-subsys0     test_subnqn                                                                                      nvme1

Device   SN                   MN                                       FR       TxPort Address        Subsystem    Namespaces      
-------- -------------------- ---------------------------------------- -------- ------ -------------- ------------ ----------------
[...]

PR #219, opened 2022-02-03
Restart controller lookup

hreinecke:
Under some circumstances nvme_lookup_ctrl() might not have enough information to uniquely identify a controller. In these cases nvme_lookup_ctrl() will only ever return the first element, and any other matching element will be ignored.
This patchset adds a new parameter 'p' to nvme_lookup_ctrl() to allow the lookup to be restarted after a certain point.


igaw:
FWIW, I did a quick test on our lab machine. There is no regression as far I can tell.

hreinecke:
And seems to resolve issue #154 ; successfully tested here locally. _And_ doesn't even impact nvme-cli :-)

PR #221, opened 2022-02-03
tree: add nvme_dump_tree()

hreinecke:
Dump the internal tree in JSON format to stdout for debugging.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -189,6 +189,16 @@ int nvme_dump_config(nvme_root_t r)
 #endif
 }
 
+int nvme_dump_tree(nvme_root_t r)
+{
+#ifdef CONFIG_JSONC
+	return json_dump_tree(r);
+#else
+	errno = ENOTSUP;
+	return -1;
+#endif
+}
+
 nvme_host_t nvme_first_host(nvme_root_t r)
 {
 	return list_top(&r->hosts, struct nvme_host, entry);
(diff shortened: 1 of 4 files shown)


igaw:
Sure, but do we really need our internal json.c code anymore? We require json-c do be available nowadays.

hreinecke:
Actually not. Guess I'll need to clean it up.

igaw:
I wanted to remove those CONFIG_JSON and CONFIG_LIBUUID section before the release anyway :)

hreinecke:
Actually, the internal json implementation is long gone anyway. So the remaining question is whether to make json-c mandatory during build; but seeing that we're already having it as a subproject we might as well.

igaw:
Yes, we have the hard dependency on json-c and libuuid since we dropped the Makefile. There was some fallouts around this move but it seems things have settled down now. I don't see a reason to keep it in.

src/nvme/tree.c
```diff
@@ -189,6 +189,16 @@ int nvme_dump_config(nvme_root_t r)
 #endif
 }
 
+int nvme_dump_tree(nvme_root_t r)
+{
+#ifdef CONFIG_JSONC
```
igaw, line 194: So this one could be removed now. #223 is in the tree now.

igaw:
Dropped the usage of the json wrappers and rebased it to the current head.

Issue #218, opened 2022-02-02
meson -Dman=true doesn't rebuild the man pages


hreinecke:
Running 'meson -Dman=true' will not regenerated the man pages; primarily because it's running off a separate subdirectory, and any man pages will be stored there (and not in the master directory 'man/doc'). But more importantly, it will just create single 'libnvme.2' file based on the contents of 'doc/libnvme.rst', and not the individual man pages we have now.

And furthermore, the 'libnvme.rst' file is never regenerated based on the contents of the actual source files.

igaw:
In nvme-cli we will only install the documentation when -Ddocs=all, -Ddocs=man or -Ddocs=html is used. First it will try to generated the docs but if the needed tooling is not available it falls back to use the pre compiled ones. I think this scheme makes also sense to implement here. Thoughts?

Issue #184, opened 2022-01-18
pre compiled man pages are not installed


igaw:
The current configuration will only install man pages when they are build from scratch. The pre compiled man pages in the source tree are not considered when installing.

igaw:
Good question. I don't know.

Sure we have: https://github.com/linux-nvme/libnvme/tree/master/doc/man

hreinecke:
Primary reason is that compiling man-pages requires additional tooling, so building the package would be getting ever more complex.
I would vote for regenerating manpage on every release.

igaw:
I agree. In this case we should install them too.

hreinecke:
Please check PR #215 . That should have all what's required here.

PR #222, opened 2022-02-03
doc: Mesonize documentation generation

igaw:
Introduce a new config flag -Ddocs-build=true which controls if the
build system should generate documentation or not. If disabled but
-Ddocs=man is used then the pre compiled documentation will be
installed.

Furthermore, we are hooking up the sphinx tooling so that we generate
the HTML documentation ouf of the rst files.

Fixes: #218
Fixes: #184


hreinecke (approved):
Exactly what I had in mind.

igaw:
There are some fallouts. Like the documents have sport 'Kernel API doc'. Also I wonder if it wouldn't be simpler to move the complete doc generation into a script and just have meson provide the API files with the proper path to the script. It's a bit aweful what I came up with.

igaw:
I decided to keep the meson build steps as they were. I don't think we can make it way simpler without loosing the nice tracking feature. 

- Fixed the kernel module name.
- sphinx warning fixed

PR #223, opened 2022-02-04
Make JSON-C mandatory

hreinecke:
We already have a json-c subproject, so we should make json-c mandatory during build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -40,7 +40,10 @@ libuuid_dep = dependency('uuid', required: true)
 conf.set('CONFIG_LIBUUID', libuuid_dep.found(), description: 'Is libuuid required?')
 
 # Check for json-c availability
-json_c_dep = dependency('json-c', version: '>=0.13', fallback : ['json-c', 'json_c_dep'])
+json_c_dep = dependency('json-c',
+			version: '>=0.13',
+			required: true,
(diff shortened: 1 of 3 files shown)


igaw:
I don't think we need the last patch. When the fallback feature is used, it implies `required == true`.

```
Run-time dependency json-c found: NO (tried pkgconfig and cmake)
Not looking for a fallback subproject for the dependency json-c because:
Use of fallback dependencies is disabled.

meson.build:43:0: ERROR: Dependency 'json-c' is required but not found.
```

eli-schwartz:
> When the fallback feature is used, it implies `required == true`.

Actually, as per the docs at https://mesonbuild.com/Reference-manual_functions.html#dependency a dependency defaults to `required: true` unless you explicitly specify `false`.

For advocates of "explicit is always better than implicit", it is possible to verbosely specify the default values to any meson function. It is effectively a no-op. There may also be cases where the boolean `true` is dynamically assigned to a variable based on conditionals, and then that variable (expanding to `true`) gets passed to the `required` kwarg, so it is generally useful to accept the default values anyway.

igaw:
@eli-schwartz Thanks for the input. Okay, let's be explicit in this case!

PR #229, opened 2022-02-10
tree: Remove CONFIG_JSONC ifdefs

igaw:
We have an hard dependency on json-c, so we can rely it's always
available. Remove old ifdefs for json-c.


hreinecke:
Yep, works for me.

PR #230, opened 2022-02-11
ioctl: nvme_io_args.dspec(Directive Specific) value is 16 bits

sc108-lee:
Command Dword 13 [31:16]

### DIFF
--- a/src/nvme/ioctl.h
+++ b/src/nvme/ioctl.h
@@ -3827,8 +3827,8 @@ static inline int nvme_flush(int fd, __u32 nsid) {
  * @reftag:	This field specifies the Initial Logical Block Reference Tag
  *		expected value. Used only if the namespace is formatted to use
  *		end-to-end protection information.
- * @dsm:	Data set management attributes, see &enum nvme_io_dsm_flags
  * @dspec:	Directive specific value
+ * @dsm:	Data set management attributes, see &enum nvme_io_dsm_flags
  */
 struct nvme_io_args {
 	__u64 slba;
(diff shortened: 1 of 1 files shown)


src/nvme/ioctl.h
```diff
@@ -3848,7 +3848,7 @@ struct nvme_io_args {
 	__u16 apptag;
 	__u16 appmask;
 	__u8 dsm;
-	__u8 dspec;
+	__u16 dspec;
```
igaw, line 3851: Can you move the dspec before dsm so that we have all __u16 grouped together. This avoids holes in the data structures.

igaw (changes requested):

src/nvme/ioctl.h
sc108-lee: addressed :)

igaw: You need also to update all inline function in the header which use struct nvme_io_args. Older compilers are complaining if the order of the struct initialization are not strictly following the order of the deceleration.

doc/html/_sources/libnvme.rst.txt
```diff
@@ -5245,8 +5245,8 @@ The nvme command status if a response was received (see
     __u16 control;
     __u16 apptag;
     __u16 appmask;
+    __u16 dspec;
     __u8 dsm;
-    __u8 dspec;
```
igaw, line 5249: BTW, could you drop the generated documentation update in this patch? This makes backporting unnecessary complex. And the plan is to update the generated documentation right before each release.

igaw (changes requested):

doc/html/_sources/libnvme.rst.txt
sc108-lee: Thanks for letting me know, I will remove doc updates :)

sc108-lee:
@igaw there is no direct use of this struct in inline function

igaw:
Right you are! :)

PR #232, opened 2022-02-11
nvme: Updated latest offset value for Interrupt Mask Clear register

Arunpandian15:

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -165,7 +165,7 @@ enum nvme_register_offsets {
 	NVME_REG_CAP			= 0x0000,
 	NVME_REG_VS			= 0x0008,
 	NVME_REG_INTMS			= 0x000c,
-	NVME_REG_INTMC			= 0x0010,
+	NVME_REG_INTMC			= 0x000f,
 	NVME_REG_CC			= 0x0014,
 	NVME_REG_CSTS			= 0x001c,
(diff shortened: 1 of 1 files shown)


Arunpandian15:
You can check this on NVMe Base spec 2.0, section 3.1.3

igaw:
Spec seems to be inconsistent or I don't get it.

In Figure 35, the offset for INTMC is 0xf but in chapter 3.1.3.4 the offset is listed as 0x10.

@hreinecke, @keithbusch any ideas what's going on here?

Arunpandian15:
Hi @igaw 
I think 0x10 is the correct value
INTMS offset is 0xC and size is 4 then INTMC offset will come as 0x10
Please confirm then we can close this PR.
Thanks!

igaw:
With the register size of 4 and INTMS is at offset 0xc, INTMC can't start at 0xf. I agree with you on that. This looks like a typo in the spec. The Linux kernel also defines at 0x10

```c
enum {
        NVME_REG_CAP    = 0x0000,       /* Controller Capabilities */
        NVME_REG_VS     = 0x0008,       /* Version */
        NVME_REG_INTMS  = 0x000c,       /* Interrupt Mask Set */
        NVME_REG_INTMC  = 0x0010,       /* Interrupt Mask Clear */
```

PR #234, opened 2022-02-11
[libnvme] Add uuidx parm to get log page API

jeffreyalien:
Add uuidx parm to nvme_get_log_page and nvme_get_log_page_padded functions

Signed-off-by: Jeff Lien <jeff.lien@wdc.com>

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -123,7 +123,6 @@ LIBNVME_1_0 {
 		nvme_get_log_fw_slot;
 		nvme_get_log_lba_status;
 		nvme_get_log_page;
-		nvme_get_log_page_padded;
 		nvme_get_log_persistent_event;
 		nvme_get_log_predictable_lat_event;
 		nvme_get_log_predictable_lat_nvmset;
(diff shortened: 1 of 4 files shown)


igaw:
Looks okay, but I am wondering if it wouldn't be better to turn the argument list into an `struct args`. `nvme_get_log_pages()` and `nvme_get_log_pages_padded()` are basically a convenience wrapper for `nvme_get_log()`. Since we have the `struct args` for `nvme_get_log()` we might just want to do the same for `nvme_get_log_pages()`.

If we do this I would also merge `nvme_get_log_pages()` and `nvme_get_log_pages_padded()` into one function and make the padded feature enabled by a member in the struct args.

hreinecke:
The whole point in having a 'struct args' is that it's easy to create wrappers and that you do _not_ have to modify existing wrappers all the time.
Hence please add a new wrapper ('nvme_get_log_pages_uuidx()' ?) instead of updating the existing one.

jeffreyalien:
Yea, I was thinking the same thing as I made the change.  I should have gone with that thought.  I'll make that change along with combining those 2 get log functions.

jeffreyalien:
@igaw @hreinecke  Please review updated pull request.  Thanks.

hreinecke (approved):
Much better.

igaw:
Removed nvme_get_log_page_padded() from libnvme.map.

PR #235, opened 2022-02-13
meson: small fixes

birkelund:


birkelund:
Breaks nvme-cli, due to internal/config.h

PR #236, opened 2022-02-13
meson: small fixes

birkelund:


igaw:
Can you please explain what you try to fix. I am worried about the current_build_dir() returning an absolute path. I understood that adding absolute paths is not recommended in meson (maybe I got this part wrong). So please help me to understand what is broken and why this fix is correct.

birkelund:
The `declare_dependency` should not include `ccan/` and `.` as include directories - since these are internal and may clash with the upper project - only the `src/` dir should be added to the list of searchable paths (`-I`).

You are right about the paths for the `-include`. Since gcc's `-include` also looks in the search path, I shouldn't remove `.` in patch 2. Instead, drop that patch and just make sure that we do not leak the ccan or project includedirs.

igaw:
Sorry! I didn't realize you updated the PR. Looks good!

birkelund:
> Sorry! I didn't realize you updated the PR. Looks good!

No problem, thanks for merging :)

PR #238, opened 2022-02-14
Corrected FW log page used  in nvme_get_log_fw_slot

samiWaheed:
Looks like a copy and paste error.  Corrected log page in  nvme_get_log_fw_slot to  NVME_LOG_LID_FW_SLOT instead of  NVME_LOG_LID_SMART


igaw:
Yep, good catch!

PR #241, opened 2022-02-15
Add support for TP8010 explicit registration

martin-belanger:
This adds support for TP8010. More precisely, it adds the ability to send a new PDU, the Discovery Information Management (DIM) PDU, which is used to perform an explicit registration with Direct Discovery Controllers (DDC) and Central Discovery Controllers (CDC).

The following new APIs are added:
```
nvme_host_set_hostsymname()
nvme_host_get_hostsymname()
nvmf_registration_ctl()
nvmf_is_registration_supported()
```
Note: The Host Symbolic Name is an optional parameter needed for by DIM PDU. When defined, it will be included in the DIM, otherwise it is omitted.

The kernel exposes a new attribute, the Discovery Controller Type (dctype) through the sysfs (kernel patch pending merge). This is used to determine whether explicit registration using the DIM PDU is supported. When the dctype (a previously reserved field of the Identify command response) is set to 1 (DDC) or 2 (CDC), we know that the remote Discovery Controller supports TP8010 and explicit registration. For pre-TP8010 DCs or any other type of controllers (admin, I/O), the dctype is set to 0 indicating that the dctype is not reported, in which case explicit registration will not be attempted.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -10,6 +10,7 @@
 #ifndef _LIBNVME_TYPES_H
 #define _LIBNVME_TYPES_H
 
+#include <stddef.h>
 #include <stdbool.h>
 #include <stdint.h>
 
@@ -909,7 +910,10 @@ struct nvme_id_psd {
  * 	       that a host is allowed to place in a capsule. A value of 0h
  * 	       indicates no limit.
  * @ofcs:      Optional Fabric Commands Support, see &enum nvme_id_ctrl_ofcs.
- * @rsvd1806:  Reserved
+ * @dctype:    Discovery Controller Type (DCTYPE). This field indicates what
+ *             type of Discovery controller the controller is (see enum
+ *             nvme_id_ctrl_dctype)
+ * @rsvd1807:  Reserved
  * @psd:       Power State Descriptors, see &struct nvme_id_psd.
  * @vs:	       Vendor Specific
  */
@@ -1007,7 +1011,8 @@ struct nvme_id_ctrl {
 	__u8			fcatt;
 	__u8			msdbd;
 	__le16			ofcs;
-	__u8			rsvd1806[242];
+	__u8			dctype;
+	__u8			rsvd1807[241];
 
 	struct nvme_id_psd	psd[32];
 	__u8			vs[1024];
@@ -1117,6 +1122,18 @@ enum nvme_id_ctrl_cntrltype {
 	NVME_CTRL_CNTRLTYPE_IO			= 1,
 	NVME_CTRL_CNTRLTYPE_DISCOVERY		= 2,
 	NVME_CTRL_CNTRLTYPE_ADMIN		= 3,
+};
+
+/**
+ * enum nvme_id_ctrl_cntrltype - Discovery Controller types
+ * @NVME_CTRL_DCTYPE_NOT_REPORTED: Not reported (I/O, Admin, and pre-TP8010)
+ * @NVME_CTRL_DCTYPE_DDC: Direct Discovery controller
+ * @NVME_CTRL_DCTYPE_CDC: Central Discovery controller
+ */
+enum nvme_id_ctrl_dctype {
+	NVME_CTRL_DCTYPE_NOT_REPORTED	= 0,
+	NVME_CTRL_DCTYPE_DDC		= 1,
+	NVME_CTRL_DCTYPE_CDC		= 2,
 };
 
 /**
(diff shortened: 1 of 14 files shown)


igaw:
Please split this up into smaller patches.

src/nvme/types.h
```diff
@@ -4696,6 +4718,162 @@ struct nvmf_discovery_log {
...
+ *
+ *              Bytes           Description
+ *   @type      01:00           Extended Attribute Type (EXATTYPE) - see enum nvmf_exattype
+ *   @len       03:02           Extended Attribute Length (EXATLEN)
+ *   @val       (EXATLEN-1)     Extended Attribute Value (EXATVAL) - size allocated for array must be a multiple of 4 bytes
+ *              +4:04
+ */
+struct nvmf_ext_attr
+{                                                           /* Bytes */
+	__le16			type;                       /* 01:00 */
```
igaw, line 4764: and given those things are already documented above do we really need those infos again?

src/nvme/types.h
```diff
@@ -4696,6 +4718,162 @@ struct nvmf_discovery_log {
...
+	__le16			len;                        /* 03:02 */
+	__u8			val[];                      /* (EXATLEN-1)+4:04 */
+};
+#define SIZEOF_EXT_EXAT_WO_VAL offsetof(struct nvmf_ext_attr, val)
```
igaw, line 4768: What's wrong with just using sizeof(struct nvmf_ext_attr) when needed?

src/nvme/util.c
```diff
@@ -550,3 +559,154 @@ char *hostname2traddr(struct nvme_root *r, const char *traddr)
...
+}
+
+static size_t copy_value(char * buf, size_t buflen, const char * value)
+{
+	size_t val_len;
+	memset(buf, 0, buflen);
+	if (value[0] == '"') value++;   /* Remove leading " */
+	val_len = strcspn(value, "\""); /* Remove trailing " */
+	memcpy(buf, value, MIN(val_len, buflen-1));
+	return val_len;
```
igaw, line 633: This function could use some new lines

src/nvme/util.c
```diff
@@ -550,3 +559,154 @@ char *hostname2traddr(struct nvme_root *r, const char *traddr)
...
+	}
+
+	return NULL;
+}
+
+size_t read_file(const char * fname, char *buffer, size_t *bufsz)
+{
+	FILE *file = fopen(fname, "re");
+	if (file) {
+		char *p = fgets(buffer, *bufsz, file);
```
igaw, line 614: Use early exit pattern and why is it no error when the file can't be opened?

src/nvme/util.c
```diff
@@ -550,3 +559,154 @@ char *hostname2traddr(struct nvme_root *r, const char *traddr)
 	freeaddrinfo(host_info);
 	return ret_traddr;
 }
+
+char* startswith(const char *s, const char *prefix)
+{
+	size_t l = strlen(prefix);
+	if (strncmp(s, prefix, l) == 0)
```
igaw, line 566: new line between deceleration and code.

btw, you are not even using the helper (which I don't disagree strongly to introduce it)

src/nvme/util.c
```diff
@@ -550,3 +559,154 @@ char *hostname2traddr(struct nvme_root *r, const char *traddr)
...
+		*s-- = '\0';
+	}
+
+	return kv;
+}
+
+char* kv_keymatch(const char *kv, const char *key)
+{
+	char *value = startswith(kv, key);
+	if (NULL != value) {
```
igaw, line 596: if (value) ?

src/nvme/util.c
```diff
@@ -550,3 +559,154 @@ char *hostname2traddr(struct nvme_root *r, const char *traddr)
...
+{
+	size_t val_len;
+	memset(buf, 0, buflen);
+	if (value[0] == '"') value++;   /* Remove leading " */
+	val_len = strcspn(value, "\""); /* Remove trailing " */
+	memcpy(buf, value, MIN(val_len, buflen-1));
+	return val_len;
+}
+
+size_t getEntityName(char *buffer, size_t bufsz)
```
igaw, line 636: No camel casing

src/nvme/util.h
```diff
@@ -424,4 +424,149 @@ static inline void nvme_id_ns_flbas_to_lbaf_inuse(__u8 flbas, __u8 *lbaf_inuse)
...
+ * is more readable than "!strcmp(a, b)".
+ *
+ * @a: first string
+ * @b: other string
+ *
+ * Example:
+ *	if (streq(string, ""))
+ *		printf("String is empty!\n");
+ */
+#define streq(a,b) (strcmp((a),(b)) == 0)
```
igaw, line 570: First of all this define already exists in ccan, second don't use it. There is no point in papering over the fact C sucks when it comes to to strings. So everyone should know what '!strcmp()' means. streq is something I have to lookup to figure out what it does.

martin-belanger:
Thanks for the review @igaw . I will address all of those and resubmit the pull request. I also just found that there is a conflict with nvme-cli. I defined a method, read_file(), that also exists in nvme-cli and their definitions clash when building nvme-cli. So I need to fix that as well. I will also break this pull request into 3 commits: One for adding the Symbolic Name, one for adding the cntrltype and dctype, and one for the adding the DIM PDU. That should make reviewing a bit easier.

PR #243, opened 2022-02-16
Add support for TP8010

martin-belanger:
This adds support for TP8010. More precisely, it adds the ability to send a new PDU, the Discovery Information Management (DIM) PDU, which is used to perform an explicit registration with Direct Discovery Controllers (DDC) and Central Discovery Controllers (CDC).

This pull request is composed of 3 commits:

1. Add the symbolic name
2. Add the cntrltype and dctype to the controller structure. 
3. Add support for the DIM command.

The following new APIs are added:
```
nvme_host_set_hostsymname()
nvme_host_get_hostsymname()
nvmf_registration_ctl()
nvmf_is_registration_supported()
```

Note: The Host Symbolic Name is an optional parameter needed for by DIM PDU. When defined, it will be included in the DIM, otherwise it is omitted.

The kernel exposes a new attribute, the Discovery Controller Type (dctype) through the sysfs (kernel patch was applied to Linux-nvme repo earlier today). This is used to determine whether explicit registration using the DIM PDU is supported. When the dctype (a previously reserved field of the Identify command response) is set to 1 (DDC) or 2 (CDC), we know that the remote Discovery Controller supports TP8010 and explicit registration. For pre-TP8010 DCs or any other type of controllers (admin, I/O), the dctype is set to 0 indicating that the dctype is not reported, in which case explicit registration will not be attempted.

Signed-off-by: Martin Belanger [martin.belanger@dell.com](mailto:martin.belanger@dell.com)

### DIFF
--- a/doc/config-schema.json
+++ b/doc/config-schema.json
@@ -25,6 +25,10 @@
 		    "description": "NVMe host ID",
 		    "type": "string"
 		},
+		"hostsymname": {
+		    "description": "NVMe host symbolic name",
+		    "type": "string"
+		},
 		"required": [ "hostnqn" ],
 		"subsystems": {
 		    "description": "Array of NVMe subsystem properties",
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -6,7 +6,8 @@
  * Authors: Hannes Reinecke <hare@suse.de>
  */
 
-%module nvme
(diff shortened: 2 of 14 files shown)


igaw:
#248 is merged so let's close this one.

Thanks!

PR #245, opened 2022-02-17
Cleanup patch for linux.h, log.h, tree.h

Arunpandian15:


igaw:
I've splitted it a bit more out and updated small format issues

Issue #244, opened 2022-02-16
__nvme_msg() is not checking for NULL before dereferencing pointer 


martin-belanger:
Function [__nvme_msg()](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/log.c#L37) could be called with a NULL `r` parameter. In fact, we do test for NULL on the very first line of this function. However, we do not check for NULL in other places where we're dereferencing this pointer (see line 58, 67, 72, etc...). 

I've seen it happen, which causes the application to coredump.

PR #246, opened 2022-02-17
log: Allow nvme_msg usage with invalid root pointer

igaw:
24ac082f481d ("Add 'nvme_root_t' argument to nvme_msg()") and
19b12831a9ec ("Move global logging variables into nvme_root_t") moved
the logging controller knobs into the root object. In case the pointer
is invalid we should not try to use it. Instead just bailing out just
log the line.

Fixes: #244

### DIFF
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -53,27 +53,32 @@ __nvme_msg(nvme_root_t r, int lvl,
 	};
 	char *header __cleanup__(cleanup_charp) = NULL;
 	char *message __cleanup__(cleanup_charp) = NULL;
-	int idx;
+	int idx = 0;
 
-	if (r->log_timestamp) {
+	if (r && lvl > r->log_level)
+		return;
+
+	if (r && r->log_timestamp) {
 		struct timespec now;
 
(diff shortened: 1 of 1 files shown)


martin-belanger (commented):
`r` also needs to be checked at lines 72-73

igaw:
This time even tested. Should work :)

src/nvme/log.c
```diff
@@ -81,7 +83,7 @@ __nvme_msg(nvme_root_t r, int lvl,
 		message = NULL;
 	va_end(ap);
 
-	if (lvl <= r->log_level)
+	if (!r || lvl <= r->log_level)
```
martin-belanger, line 86: Shouldn't we test that the message is at the right level at the beginning of the function and return right away if we're not going to print it? This would avoid performing a whole lot of operations for nothing.

In other words:
```
/* Check if message is at the right level. If not, get out immediately */
if (r && lvl > r->log_level)  
    return;
```

igaw: Good point. I'll update it accordingly :)

igaw:
I've also reformatted to code slightly, so it's easier to read (at least in my opinion :))

PR #248, opened 2022-02-18
TP8010 support refactored

igaw:
This is based on #243 with a lot of small nitpicking cleanups and refactoring things (see also my initial review in #241).

Note: I also renamed a lot of the types to match with what is written in the spec.

@martin-belanger I haven't tested this yet. I am trying to get a test setup running. In case you have time to give this a go I would really appreciate it.

### DIFF
--- a/doc/config-schema.json
+++ b/doc/config-schema.json
@@ -25,6 +25,10 @@
 		    "description": "NVMe host ID",
 		    "type": "string"
 		},
+		"hostsymname": {
+		    "description": "NVMe host symbolic name",
+		    "type": "string"
+		},
 		"required": [ "hostnqn" ],
 		"subsystems": {
 		    "description": "Array of NVMe subsystem properties",
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -6,7 +6,8 @@
  * Authors: Hannes Reinecke <hare@suse.de>
  */
 
-%module nvme
+%module(docstring="Python bindings for libnvme") nvme
+%feature("autodoc", "1");
 
 %include "exception.i"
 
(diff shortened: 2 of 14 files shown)


igaw:
Export map updated (s/nvmf_registration_ctl/nvmf_registration_ctlr/)

martin-belanger:
> This is based on #243 with a lot of small nitpicking cleanups and refactoring things (see also my initial review in #241).
> 
> Note: I also renamed a lot of the types to match with what is written in the spec.
> 
> @martin-belanger I haven't tested this yet. I am trying to get a test setup running. In case you have time to give this a go I would really appreciate it.

I will give it a try right now.

martin-belanger:
Hi @igaw - Thanks for your review and refactoring. It must have taken you some time to do all this and I truly appreciate it.

So, I tested everything. I connected to a DDC and a CDC and performed explicit registration using your code and everything works as intended.

There are just couple of minor things (now it's my turn to nitpick... :wink:).
1. The abbreviation `ctl` was meant as `control`. For example, `registration_ctl` meant `registration_control` and `nvmf_registration_ctl` meant `nvmf_registration_control`. You renamed `ctl` to `ctlr`, which, I presume, is the abbreviation for `controller`. It would be better to use `ctrl` instead of `ctlr` as the abbreviation for `control` (or not abbreviate at all and just use `control`). In any case, this is very minor and it doesn't really matter that much.
2. The function `nvme_fetch_cntrltype_dctype_from_id()` was meant to be internal to `libnvme`. You added it to `libnvme.map`, which makes it public. Again, it this is not a big deal and can stay as it is. 

Regards,
Martin

igaw:
1. oh, that was a typo, I wanted to use it consistent. So it should be ctrl. And after looking at the existing function names I think we should use a verb instead of a noun: `nvmf_register_ctrl`.
2. nvme_fetch_cntrltype_dctype_from_id if this is an internal function it should not be in this header file. Let me check were we could put it.

igaw:
I"ve moved nvme_fetch_cntrltype_dctype_from_id to the last patch and marked is static. Looks way better :)

martin-belanger (approved):
Looks good

igaw:
Just one thing I am still pondering on is the name of `nvme_send_dim_command`. It doesn't really fit into the theme of the naming pattern.  Maybe `nvme_dim_send`?

igaw:
Renamed `nvme_send_dim_command` to  `nvme_dim_send` and also added it to the export map

PR #249, opened 2022-02-18
Fix to update size for telemetry log page when telemetry controller initiated data is unavailable.

AkhileshRN:
…nitiated data is unavailable.


jeffreyalien (commented):
This looks good to me.

PR #251, opened 2022-02-21
SWIG: remove explicit invocation

glimchb:
python distutils handles swig already, so no need to
explicitely define it in meson build


glimchb:
bug? closing for a sec

PR #253, opened 2022-02-22
fabrics: add default port number for NVMe/TCP I/O controllers

igaw:
As per section 7.4.9.3 "Transport Service Identifier" of the NVMe over
Fabrics 1.1 specification, the default IANA port number for a NVMe/TCP
discovery controller is 8009. But at the same time, it also clearly
states that NVMe/TCP I/O controllers should not use TCP port number
8009, but may instead use 4420 as the default here.

So make sure to fill these values appropriately, and pass it down.

[dwagner: backport from monolithic branch]

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -922,8 +922,13 @@ void nvme_free_ctrl(nvme_ctrl_t c)
 static void discovery_trsvcid(nvme_ctrl_t c)
 {
 	if (!strcmp(c->transport, "tcp")) {
-		/* Default port for NVMe/TCP discovery controllers */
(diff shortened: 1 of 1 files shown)


igaw:
@martin-gpy

martin-gpy:
Daniel,

We may need to backport this patch as well - https://github.com/linux-nvme/nvme-cli/commit/f74ac1b841b0aae73969debaed0444f0ecb03dba

igaw:
Please do not use this PR for a discussion. Open a new issue.

igaw:
And you don't have to do it. This patch is not necessary. The variable is always initialized (see nvme_create_ctrl)

PR #257, opened 2022-02-27
 pyproject.toml instead of setup.py for new packaging

glimchb:
This allows to build with meson and package for Pypi

Build like this:
```
    python -m build
```


glimchb:
see https://thiblahute.gitlab.io/mesonpep517/pyproject.html

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,52 @@
...
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+  push:
+    branches: [ main ]
+    # Publish semver tags as releases.
+    tags: [ 'v*.*.*' ]
```
igaw, line 17: Would it possible to have also a trigger for the release candidate tags, but then it would be pushed to a pypi devel package? Would this be useful?

glimchb: @igaw sure, it is controlled by 
```
      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
below. once merged and working, we can adjust as we want, no problem

igaw: Ah I see, so we can do string ops on 'github.ref' to figure out what type of tag it is.

glimchb: exactly

igaw:
Looks pretty sane to me. I think it would worth also to add some info to the README. At least the python3-build dependency should be mentioned for 'python3 -m build .'

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,52 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        sudo apt-get install libjson-c-dev meson
+        python -m pip install --upgrade pip
+        pip install build --sdist
```
glimchb, line 38: will add `--wheel` in a separate PR, cause of the complications with `manylinux`

glimchb:
tested here https://github.com/glimchb/libnvme/runs/5360519421?check_suite_focus=true

igaw:
I've played a bit around with the Test PyPI instance and it looks good. The only thing I am worried about is that it's impossible to fix any mistake. If the release uploaded that's it. We can't overwrite/delete or anything later. This is a one shot operation. Hence having the triggers enabled right now without the release candidate filtering sounds like a bad idea. If I merge this and I do another release candidate the '1.0' release is done for PyPI. Or do I miss something here?

glimchb:
@igaw 
- gh-action-pypi-publish will override released packages with the same name as default behavior. Once we like what is happening we can turn off the default override option.
- as you own pypi account you can manually remove any individual release that you want
- Yep, let's create first release and then update the tokens to be per-project

igaw:
FTR, I was getting these warning.

```
HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
This filename has already been used, use a different version. See https://pypi.org/help/#file-name-reuse for more information.
```

Obviously, this happens when trying to reupload `libnvme-1.0.tar.gz`.

So you are saying it is possible to overwrite although the file-name-reuse warning?

glimchb:
yes, you can
see https://github.com/pypa/gh-action-pypi-publish#tolerating-release-package-file-duplicates
```
   with:
     skip_existing: true
```

glimchb:
@igaw if you feel more comfortable, we can do test pypi only for now and leave the production pypi for a separate PR

igaw:
I am not talking about the workflow. I am talking about the PyPI repository. It clearly states once a file is uploaded with a given name, there is nothing which could changed afterwards. All release candidates have the same version number 1.0. So the first upload to PyPI would be the 1.0 release even if it's a release candidate. This is what I am concerned about.

glimchb:
@igaw want me to add commit to this PR for remove the production publish for time being ? no prob

igaw:
Thanks. Let's see how things work out from here :)

glimchb:
@igaw looks like workflow is not triggered at all https://github.com/linux-nvme/libnvme/actions/workflows/python-publish.yml

igaw:
Isn't this to be expected, there is no tag?

glimchb:
@igaw `master` vs `main` branch names in workflow...

glimchb:
@igaw fixed here https://github.com/linux-nvme/libnvme/pull/266
workflow runs and fails on missing token, did you add one for `test` pypi ?

igaw:
I don't have enough permissions to set the token. This is something @keithbusch needs to do or he boost my rights :)

igaw:
FYI, I've manually disabled the workflow for the time being because it  marks all builds as failed.

glimchb:
Sure @igaw can you please email or ping @keithbusch to add us new tokens or promote you to admin ?

igaw:
pypi tokens added as secret and workflow reanabled.

igaw:
As predicted, it will not work as if we upload the same file name again:

`This filename has already been used, use a different version. See https://test.pypi.org/help/#file-name-reuse for more information.`

https://github.com/linux-nvme/libnvme/runs/5423013685?check_suite_focus=true

So the trigger needs to be changed so that only uploads for releases/tags are done.

glimchb:
Shoot, I was under impression that at least for test.pypi.org they allow overrides... 
I will fix it on Sunday

igaw:
No hurry, take your time. It's not that we can't upload any artifacts later.

glimchb:
thanks @igaw

PR #259, opened 2022-02-28
types: status code type value update as per NVMe specification

Arunpandian15:
updated the status code type value as per the NVMe specification 2.0b


igaw:
I suppose this makes sense to match with the spec. The old values were already shifted to the right position for matching with the completion queue entry. As nvme-cli or libnvme is using it anyway there is no regression to be expected. Let's merge it.

PR #265, opened 2022-03-01
CI: use cibuildwheel

glimchb:
### === This change is not ready yet ===

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -35,9 +35,13 @@ jobs:
       run: |
         sudo apt-get install libjson-c-dev meson
         python -m pip install --upgrade pip
-        pip install build
+        pip install build cibuildwheel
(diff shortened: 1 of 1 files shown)


igaw:
Can you rebase this on the master. Patch two is already merged. And please make two commits for the  CI changes and meson.build change with an explanation in the commit message.

glimchb:
not sure about `link_with`... figuring this out still...

igaw:
FWIW, I am not really convinced that we are building the Python extension correctly. I noticed that we have in source tree artifices after the build which is a clear indication something is wrong. I haven't found time to really try to figure out what's going on.

glimchb:
@igaw I agree, something is not 100% there, just learning bit by bit how it's done properly

igaw:
Could you add some explanation to the commit message why this is needed?

PR #274, opened 2022-03-04
types: fix status code type bug (wrong masking)

sc108-lee:
the error comes from fc274c57c6b3063d35dd76e36b777adc4da29e0d
using NVME_GET for masked value #148


sc108-lee:
@Arunpandian15 you might want to see this

Bug:
NVMe status: unrecognized(0x13)
Meant to be:
NVMe status: PRP Offset Invalid: The Offset field for a PRP entry is invalid(0x13)

Arunpandian15:
@sc108-lee Thanks for catching up. understood the change point.

PR #276, opened 2022-03-06
CI: Publish only released package to TestPyPI

glimchb:

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -39,6 +39,10 @@ jobs:
     - name: Build package
       run: python -m build --sdist
     - name: Publish package to TestPyPI
+      # upload to PyPI on every tag starting with 'v'
+      # if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
+      # alternatively, to publish when a GitHub Release is created, use the following rule:
+      if: github.event_name == 'release' && github.event.action == 'published'
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         user: __token__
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -8,7 +8,7 @@
 project(
     'libnvme', ['c', 'cpp'],
     meson_version: '>= 0.47.0',
-    version: '1.0',
+    version: '1.0rc5',
```
igaw, line 11: I was following what systemd is doing here and there were fallouts with a postfix by some tooling (sorry can't remember but I think we have some discussion on this in the 'issues'.

.github/workflows/python-publish.yml
```diff
@@ -39,13 +39,18 @@ jobs:
     - name: Build package
       run: python -m build --sdist
     - name: Publish package to TestPyPI
+      # upload to PyPI on every tag starting with 'v'
+      # if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
+      # alternatively, to publish when a GitHub Release is created, use the following rule:
+      if: github.event_name == 'release' && github.event.action == 'published'
```
igaw, line 45: BTW,  the trigger  for tags is not firing for 'v1.0' because it expects 'v1.0.x' as far I can tell. And there is the question if we shouldn't just trigger for the publish event.

meson.build
```diff
@@ -18,7 +18,7 @@ project(
     ]
 )
 
-library_version = meson.project_version() + '.0'
+library_version = meson.project_version().split('rc').get(0) + '.0'
```
igaw, line 21: This is not working for soname. library_version needs to be semver formatted.

igaw (changes requested):

PR #282, opened 2022-03-08
CI: add python wheel build in addition to source package

glimchb:
When using source distribution,
python package will be compiled during pip install.

Wheels make the end-to-end installation of Python packages faster:
- wheels are typically smaller in size than source distributions
- pip fetches a prebuilt wheel and avoids the intermediate step of building

So why cibuildwheel and manylinux?

See https://github.com/pypa/manylinux and https://github.com/pypa/cibuildwheel

The goal of the manylinux project is to provide a convenient way to distribute binary Python extensions as wheels on Linux

See https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml

Skipping musllinux and auditwheel repair.
Skipping MacOS and Windows for now.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -6,47 +6,62 @@
 # separate terms of service, privacy policy, and support
 # documentation.
 
-name: Upload Python Package
+name: Build and upload to PyPI
 
 on:
-  release:
-    types: [published]
   push:
     branches: [ master ]
-    # Publish semver tags as releases.
-    tags: [ 'v*.*.*' ]
   pull_request:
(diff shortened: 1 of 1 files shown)


glimchb:
tested here https://github.com/glimchb/libnvme/actions/runs/1952382632
@igaw the PUSH to PYPI will not happen now but build will happen on every PR - that's what we want now
I added detailed description to the commit message about what it is and why we need it

igaw:
Thanks for figuring this out. Looks like a big pain to provide the binding for all platforms.

glimchb:
Indeed :-(

Issue #95, opened 2021-11-08
Add versioning to exported symbols


igaw:
Add versioning to all exported symbols,  see also discussion in #51

igaw:
I've been looking at https://github.com/GNOME/glib is doing the symbol versioning. 

1) There is no the linker map file.  Instead every symbol which needs to be exported is annotated directly in the source code
```
GLIB_AVAILABLE_IN_ALL
GArray* g_array_new               (gboolean          zero_terminated,
				   gboolean          clear_,
				   guint             element_size);
```
That means the symbols are either 'hidden', 'internal' or 'default'. The versioning is not done by moving the symbols into a different linker section. It's probably safe to assume doing anything else then 'hidden', 'internal' or 'default' is asking for troubles.

2) There are macro definitions for every release which allow to attach versions symbols or if it's getting deprecated. But this is only done on source code level. This means this kind of versioning is useful during compile time. IIRC, the library issues depreciated warnings. See https://github.com/GNOME/glib/blob/main/glib/gversionmacros.h for more details.

3) The library versioning is done via
```
glib_version = meson.project_version()
glib_api_version = '2.0'
version_arr = glib_version.split('.')
major_version = version_arr[0].to_int()
minor_version = version_arr[1].to_int()
micro_version = version_arr[2].to_int()

interface_age = minor_version.is_odd() ? 0 : micro_version
binary_age = 100 * minor_version + micro_version

soversion = 0
# Maintain compatibility with previous libtool versioning
# current = minor * 100 + micro
current = binary_age - interface_age
library_version = '@0@.@1@.@2@'.format(soversion, current, interface_age)
darwin_versions = [current + 1, '@0@.@1@'.format(current + 1, interface_age)]
```
which is almost what we currently have. The soversion part is currently hard coded to '1'. This should be made a bit more flexible.

Given this, I think we should think add something like the version macros as glib has and replace the libnvme.map with it. The rest looks good IMO.

igaw:
I did some more research, as glib seems not the best example. 

Anyway, @hreinecke's proposal seems to be used also by other projects, e.g. dpdk or systemd,
```
LIBSYSTEMD_248 {
global:
        sd_bus_open_user_machine;
        sd_bus_message_send;

        sd_event_source_set_ratelimit;
        sd_event_source_get_ratelimit;
        sd_event_source_is_ratelimited;

        sd_device_get_action;
        sd_device_get_seqnum;
        sd_device_new_from_stat_rdev;
[...]

PR #286, opened 2022-03-13
Version: add RC (release candidate) to the meson version

glimchb:
This is to help uploading python package to pip

see discussion in https://github.com/linux-nvme/libnvme/issues/95 and https://github.com/linux-nvme/libnvme/issues/287

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -52,7 +52,7 @@ jobs:
   upload_pypi:
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' && github.event.action == 'published'
+    #if: github.event_name == 'release' && github.event.action == 'published'
     steps:
       - uses: actions/download-artifact@v2
         with:
@@ -62,6 +62,7 @@ jobs:
       - name: Publish package to TestPyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
+          verbose: true
           user: __token__
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
           repository_url: https://test.pypi.org/legacy/
--- a/meson.build
+++ b/meson.build
@@ -8,7 +8,7 @@
 project(
     'libnvme', ['c', 'cpp'],
     meson_version: '>= 0.47.0',
-    version: '1.0',
+    version: '1.0rc7',
     license: 'LGPLv2+',
     default_options: [
         'c_std=gnu99',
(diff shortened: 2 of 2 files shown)


glimchb:
@igaw Looks like `secrets.TEST_PYPI_API_TOKEN` is empty 

https://github.com/linux-nvme/libnvme/runs/5674662663?check_suite_focus=true

igaw:
Strange, I am pretty sure I added the token. Anyway, created a new one and added it to github.

glimchb:
> Strange, I am pretty sure I added the token. Anyway, created a new one and added it to github.

@igaw, sorry, not working still...

https://github.com/linux-nvme/libnvme/runs/5708205580?check_suite_focus=true
```
Warning:  It looks like you are trying to use an API token to authenticate in the package index and your token value does not start with "pypi-" as it typically should. This may cause an authentication error. Please verify that you have copied your token properly if such an error occurs.
```

igaw:
I think, you don't have the permission to upload. Which make sense that only maintainers are allowed to to trigger the upload.

glimchb:
@igaw I don't think it's about permissions, the message is clear that the secret is not well-defined... 
I can't see the secrets here, but they need to start with  "pypi-" 
we should have 2 secrets `secrets.TEST_PYPI_API_TOKEN` and `secrets.PYPI_API_TOKEN`

igaw:
The token is correctly prefixed. I made sure of that.

glimchb:
let me try again to trigger the upload with verbose mode now

glimchb:
@igaw sorry to persist on this... I see `Environment secrets` vs `Repository secrets`... the secrets are under repository, right ?

glimchb:
thanks @igaw !

glimchb:
- `pyproject.toml` supports dynamic version, so we don't have to change meson version, so this is good.

- but according to https://thiblahute.gitlab.io/mesonpep517/pyproject.html 
```
NOTE: The project version and name are extracted from the meson.build [project()](http://mesonbuild.com/Reference-manual.html#project) section.
```
This is why I'm considering to switch to 
https://meson-python.readthedocs.io/en/stable/usage/start.html

glimchb:
@igaw I just was hoping it would be good if we could actually test if we have the pypi stuff correctly setup and it works before we switch to https://pypi.org/project/meson-python/

glimchb:
closing this once, since @igaw is working on versioning already

igaw:
After ponder on this for while I think we just follow the Linux and model here. We keep the version string until the first rc or proper release. This seems to be an accepted way to handle it.

PR #289, opened 2022-03-13
dist: add missing BuildRequires to spec file

### LINKED ISSUE #288: RPM: spec file is missing `BuildRequires:meson`
see https://docs.fedoraproject.org/en-US/packaging-guidelines/Meson/

You **MUST** add following BuildRequires:
```
BuildRequires: meson
```

glimchb:
According to https://docs.fedoraproject.org/en-US/packaging-guidelines/Meson/

Fixes #288


glimchb:
@igaw how do you usually build RPM with meson ? I do it probably the wrong way like this:
```
meson dist -C .build --formats gztar
cp .build/meson-dist/libnvme-1.0.tar.gz ~/rpmbuild/SOURCES/
rpmbuild -ba .build/libnvme.spec
```
any better way ?

igaw:
As far I can tell, most distro do no use upstream spec files, instead they have their own which usually embedded into a build system. The spec file provided here is thought as quick way for a developer to build a rpm to distribute to the testers.

So I can't really say if you are doing wrong as I am using the build service from openSUSE. But looking at your snipped, I think this could be added to the Makefile as special target if meson doesn't really provide anything in this regard.

glimchb:
added here https://github.com/linux-nvme/libnvme/pull/290 @igaw

PR #290, opened 2022-03-14
rpm: add makefile rule for rpms

glimchb:
see https://linux.die.net/man/8/rpmbuild
```
-ba
Build binary and source packages (after doing the %prep, %build, and %install stages).
```


glimchb:
thanks @igaw

PR #292, opened 2022-03-14
tree: nvme_configure_ctrl(): Set errno to zero on success

tbzatek:
The `nvme_get_ctrl_attr()` call might return `NULL` with `errno` set accordingly. That's fine in many cases, such as in `nvme_configure_ctrl()`, returning 0 afterwards. However the call stack up to `nvme-cli` through `nvmf_add_ctrl()` keeps the `errno` unchanged while reporting success and `nvmf_connect()` in `nvme-cli` surprisingly looks primarily at `errno` (patch will follow).

Anyway, for `libnvme` I though that explicitly setting `errno` to zero at the end of `nvme_configure_ctrl()` would not hurt. There might be a lof of other cases all over the code that might uncover hidden issues in `nvme-cli`.

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -433,21 +433,22 @@ static char *__nvme_get_attr(const char *path)
 {
 	char value[4096] = { 0 };
 	int ret, fd;
+	int saved_errno;
 
 	fd = open(path, O_RDONLY);
-	if (fd < 0) {
-#if 0
-		nvme_msg(LOG_DEBUG, "Failed to open %s: %s\n", path,
-			 strerror(errno));
-#endif
+	if (fd < 0)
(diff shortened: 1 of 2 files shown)


igaw:
My only question is: Is NULL a valid return value from `nvme_get_ctrl_attr()`? If so we should set the errno there to 0, no?

tbzatek:
> My only question is: Is NULL a valid return value from `nvme_get_ctrl_attr()`? If so we should set the errno there to 0, no?

Good question. Currently there's no way to distinguish between an empty string and an error case. That's not really a problem as most of the callers can cope with NULL just fine and don't care about the reason. It's a public API though.

Pushed a sketch that could clarify that.

Issue #295, opened 2022-03-18
No trsvcid specified for 'rdma'


igaw:
```sh
# nvme connect -t rdma -a 192.1.11.151 -n nqn.XXX -Q 1024 -l 3600
No trsvcid specified for 'rdma'
# nvme connect -t rdma -trsvcid=4420 -a 192.1.11.151 -n nqn.XXX -Q 1024 -l 3600
```
@martin-gpy suggested:
```patch
@@ -1014,16 +1014,8 @@ struct nvme_ctrl *nvme_create_ctrl(nvme_
 		c->cfg.host_iface = strdup(host_iface);
 	if (trsvcid)
 		c->trsvcid = strdup(trsvcid);
-	else if (discovery)
+	else 
 		discovery_trsvcid(c);
-	else if (!strncmp(transport, "rdma", 4) ||
-		 !strncmp(transport, "tcp", 3)) {
-		nvme_msg(r, LOG_ERR, "No trsvcid specified for '%s'\n",
-			 transport);
-		errno = EINVAL;
-		__nvme_free_ctrl(c);
-		c = NULL;
-	}
```

PR #296, opened 2022-03-18
tree: Add support for default trsvcid for all controllers

igaw:
When the user doesn't provide trsvcid always use the default values
for tcp/rdma transport.

Fixes: #295

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -933,7 +933,7 @@ void nvme_free_ctrl(nvme_ctrl_t c)
 #define ____stringify(x...) #x
 #define __stringify(x...) ____stringify(x)
 
-static void discovery_trsvcid(nvme_ctrl_t c)
+static void set_default_trsvcid(nvme_ctrl_t c)
 {
 	if (!strcmp(c->transport, "tcp")) {
 		if (c->discovery_ctrl) {
@@ -970,7 +970,6 @@ struct nvme_ctrl *nvme_create_ctrl(nvme_root_t r,
 				   const char *host_iface, const char *trsvcid)
 {
 	struct nvme_ctrl *c;
-	bool discovery = false;
 
 	if (!transport) {
 		nvme_msg(r, LOG_ERR, "No transport specified\n");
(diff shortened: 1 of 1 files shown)


src/nvme/tree.c
```diff
@@ -1014,16 +1012,8 @@ struct nvme_ctrl *nvme_create_ctrl(nvme_root_t r,
...
-	else if (!strncmp(transport, "rdma", 4) ||
-		 !strncmp(transport, "tcp", 3)) {
-		nvme_msg(r, LOG_ERR, "No trsvcid specified for '%s'\n",
-			 transport);
-		errno = EINVAL;
-		__nvme_free_ctrl(c);
-		c = NULL;
-	}
+	else
+		set_default_trsvcid(c);
```
hreinecke, line 1016: Sure this is correct?
Spec states the '4420' is the default port for RDMA, and 8009 is the default _discovery_ port for TCP.
So we cannot eliminate the check for discovery for tcp.

hreinecke (changes requested):

src/nvme/tree.c
igaw: What about the check in set_default_trsvcid() itself. Isn't this taking care of the discovery port for TCP being 8009 and IO ports on 4420?

martin-gpy: Isn't that already handled in set_default_trsvcid()? 
For NVMe/TCP - discovery controllers would use port 8009, whereas I/O controllers would use 4420
For NVMe/RDMA - both discovery controllers & I/O controllers would use 4420

martin-gpy: > Sure this is correct? Spec states the '4420' is the default port for RDMA, and 8009 is the default _discovery_ port for TCP. So we cannot eliminate the check for discovery for tcp.

[](https://github.com/martin-gpy)Isn't that already handled in set_default_trsvcid()?
For NVMe/TCP - discovery controllers would use port 8009, whereas I/O controllers would use 4420
For NVMe/RDMA - both discovery controllers & I/O controllers would use 4420
@martin-gpy

hreinecke: Ah, right, you are correct.

PR #298, opened 2022-03-18
linux: fixup log page offset in nvme_get_log_page()

hreinecke:
When resubmitting the log page with a different length in
nvme_get_log_page() we need to store the original length
of the log page to avoid the length being overwritten.


hreinecke:
Fixes https://github.com/linux-nvme/nvme-cli/issues/1447.

Issue #291, opened 2022-03-14
[loop] Invalid subsystem records after scanning


tbzatek:
This is somewhat similar to #154 except this fails with just a single connection. We've been using the loop transport for simple setups in our CI and found this via segfaulting nvme-cli:

Steps to reproduce:
1. Setup a simple target using the loop transport: [target.json](https://github.com/linux-nvme/libnvme/files/8246505/target.txt)
2. `# nvme connect --transport=loop --hostnqn=test_hostnqn --nqn=test_subnqn`
3. `# nvme list --verbose`
```
Subsystem        Subsystem-NQN                                                                                    Controllers
---------------- ------------------------------------------------------------------------------------------------ ----------------
nvme-subsys0     test_subnqn                                                                               
(null)           test_subnqn                                                                                      nvme0

Device   SN                   MN                                       FR       TxPort Address        Subsystem    Namespaces      
-------- -------------------- ---------------------------------------- -------- ------ -------------- ------------ ----------------
nvme0    d55ce7cdd937df5c7dea Linux                                    5.17.0-0 loop                  (null)       

Device       Generic      NSID     Usage                      Format           Controllers     
------------ ------------ -------- -------------------------- ---------------- ----------------
nvme0n2      ng0n2        2          1.07  GB /   1.07  GB    512   B +  0 B   
nvme0n1      ng0n1        1          1.07  GB /   1.07  GB    512   B +  0 B   
```
4. `# nvme list --output-format=json --verbose`
```
Segmentation fault (core dumped)
```
(this fails on `nvme_subsystem_get_name()` returning `NULL`)
```
#0  0x00007ffff7ed7dbd in __strlen_avx2 () from /lib64/libc.so.6
#1  0x00007ffff7f78e2d in json_object_new_string (s=0x0) at /usr/src/debug/json-c-0.15-3.fc36.x86_64/json_object.c:1290
#2  0x000000000043e8e5 in json_detail_list (r=0x4d4be0) at ../nvme-print.c:7102
```

I haven't looked at where this actually comes from, reporting early.

tbzatek:
Alright, managed to find several issues, my unique hostnqn being part of the story, will post details tomorrow.

tbzatek:
[...]

PR #304, opened 2022-03-22
tree: Fix subsystem initialization in nvme_scan_ctrl()

tbzatek:
The subsystem instance would normally get properly initialized
in later stages of topology scanning, unless there's a host
mismatch and the instance would stand alone.

Related to #291


hreinecke (approved):
Yeah, seems that we' re missing the call to nvme_init_subsystem().

PR #307, opened 2022-03-22
Add lbstm support for create-ns

bjpaupor:
Signed-off-by: Brandon Paupore <brandon.paupore@wdc.com>

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -1700,6 +1700,7 @@ enum nvme_lbaf_rp {
  *	      remains fixed throughout the life of the namespace and is
  *	      preserved across namespace and controller operations
  * @lbaf:     LBA Format, see &struct nvme_lbaf.
+ * @lbstm:    Logical Block Storage Tag Mask for end-to-end protection
  * @vs:	      Vendor Specific
  */
 struct nvme_id_ns {
@@ -1743,7 +1744,8 @@ struct nvme_id_ns {
 	__u8			nguid[16];
 	__u8			eui64[8];
 	struct nvme_lbaf	lbaf[64];
-	__u8			vs[3712];
+	__le64			lbstm;
+	__u8			vs[3704];
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/nvme/types.h
```diff
@@ -1743,7 +1743,8 @@ struct nvme_id_ns {
 	__u8			nguid[16];
 	__u8			eui64[8];
 	struct nvme_lbaf	lbaf[64];
-	__u8			vs[3712];
+	__le64			lbstm;
```
igaw, line 1746: Please also update the documentation accordingly.

src/nvme/util.c
```diff
@@ -402,7 +402,7 @@ void nvme_init_dsm_range(struct nvme_dsm_range *dsm, __u32 *ctx_attrs,
 }
 
 void nvme_init_id_ns(struct nvme_id_ns *ns, __u64 nsze, __u64 ncap, __u8 flbas,
-		__u8 dps, __u8 nmic, __u32 anagrpid, __u16 nvmsetid)
+		__u8 dps, __u8 nmic, __u32 anagrpid, __u16 nvmsetid, __u64 lbstm)
```
igaw, line 405: As this is breaking the API, do we really need to add lbstm?  Actually I wonder if we shouldn't just remove the
function as nvme_id_ns has many members and this function is likely always missing out for someone. The only user for this function I found is in create_ns() in nvme-cli. We could easily ditch this function IMO.

igaw (changes requested):

src/nvme/util.c
bjpaupor: Taking it out and handling this in nvme-cli makes sense to me, removed the function.

src/nvme/types.h
bjpaupor: Thanks for the catch, updated.

src/nvme/util.c
igaw: This should be a new commit. It has nothing to do with introducing the lbstm into struct nvme_id_ns. 
Also can nvme_init_id_ns also remove from the src/libnvme.map export file? Obviously nvme-cli needs a fix as well.

Thanks!

igaw:
Looks good. Let's get the nvme-cli change aligned and then merge both changes at the same time. Thanks!

PR #308, opened 2022-03-23
test: print test logs for pass case

Arunpandian15:
test log print for passed cases and aligning
log output.


Arunpandian15:
Included PASSED as prefix for pass cases. Thanks @igaw

PR #309, opened 2022-03-23
tree: Handle NULL subsysname in nvme_scan_ctrl()

tbzatek:
As noted couple of lines earlier, the subsysname string might
be NULL. In that case, defer initialization of nvme_subsystem_t
in hope that it would get initialized while scanning subsystems.


tbzatek:
This is a followup to #304 - Github was somewhat broken yesterday, giving me Internal server error on any `git push` and eventually succeeded... only to make my changes ignored in the webui and as checked after the merge, also in the merged code. 

Resubmitting, just a tiny fix.

igaw:
Thanks!

Issue #316, opened 2022-03-28
Should kato be set by default when setting persistent?


martin-belanger:
The function `nvme_ctrl_set_persistent()` sets the boolean `c->persistent` as follows:
```
void nvme_ctrl_set_persistent(nvme_ctrl_t c, bool persistent)
{
	c->persistent = persistent;
}
```
From what I can tell, the boolean `c->persistent` is not really used for anything (at least from the perspective of configuring the kernel). Should we use `c->persistent` to set a default KATO value for persistent connections to discovery controllers? 

For example:
```
void nvme_ctrl_set_persistent(nvme_ctrl_t c, bool persistent)
{
	c->persistent = persistent;
	if (c->persistent && c->discovery_ctrl && c->cfg.keep_alive_tmo == 0) {
		c->cfg.keep_alive_tmo = DEFAULT_KATO; /* Default could be 120 sec (as in base specs) or any value we agree on */
	}
}
```

igaw:
You mean we should move `set_discovery_kato` into libnvme?

martin-belanger:
> You mean we should move `set_discovery_kato` into libnvme?

Not exactly. I guess I don't see what `nvme_ctrl_set_persistent()` is used for. All this does is set a boolean, but this boolean is not used to change the way we configure the kernel. It's just an internal boolean that one can set or read. Maybe the `persistent` boolean should not even exist if it's not going to be used for anything useful.

igaw:
Yes, the whole decision if a controller is removed is completely in nvme-cli. 

The only place I found some code around c->persistent is in json_update_attributes()
```
		if (!strcmp("persistent", key_str) &&
		    !nvme_ctrl_is_persistent(c))
			nvme_ctrl_set_persistent(c, true);
		if (!strcmp("discovery", key_str) &&
		    !nvme_ctrl_is_discovery_ctrl(c))
			nvme_ctrl_set_discovery_ctrl(c, true);
```
But that's all. So not really used. It looks like we could drop this bits from the library.

martin-belanger:
I will submit a patch that removes it then. Thanks for the feedback.

hreinecke:
Ehh ... the only thing the 'persistent' flag does is _not_ removing the discovery controller:
~~~
                               if (e->subtype != NVME_NQN_NVME &&
                                    !persistent) {
                                        nvme_disconnect_ctrl(child);
                                        nvme_free_ctrl(child);
                                }
~~~
[...]

PR #317, opened 2022-03-28
Remove unused persistent flag.

martin-belanger:
The persistent flag had no effect on the configuration of controllers. Removing it to eliminate confusion.

Related to issue https://github.com/linux-nvme/libnvme/issues/316


martin-belanger:
nvme-cli needs the persistent flag and thus this cannot be removed.

PR #324, opened 2022-03-30
Rework nvme_scan_subsystem()

hreinecke:
When scanning the subsystem in sysfs we cannot infer the corresponding host that easily, so rework the logic to not allocate the subsystem with the wrong host.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -40,6 +40,7 @@ static void __nvme_free_host(nvme_host_t h);
 static void __nvme_free_ctrl(nvme_ctrl_t c);
 static int nvme_subsystem_scan_namespace(nvme_root_t r,
 					 struct nvme_subsystem *s, char *name);
+static int nvme_init_subsystem(nvme_subsystem_t s, const char *name);
 static int nvme_scan_subsystem(nvme_root_t r, const char *name,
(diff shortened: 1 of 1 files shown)


src/nvme/tree.c
```diff
@@ -1374,9 +1374,11 @@ nvme_ctrl_t nvme_scan_ctrl(nvme_root_t r, const char *name)
 	free(subsysnqn);
 
 	ret = 0;
-	if (s && !s->name && subsysname)
-		ret = nvme_init_subsystem(s, subsysname);
-	free(subsysname);
+	if (subsysname) {
+		if (s && !s->name)
+			ret = nvme_init_subsystem(s, subsysname);
+		free(subsysname);
```
tbzatek: Oh, I lived in assumption that `free()` called on NULL pointer does nothing. At least the manpage says so but I was always wondering if all the libcs out there follow this semantic.
(FWIW, it's me to blame for this `subsysname` free here)

tbzatek (approved):

src/nvme/tree.c
igaw: IIRC free(NULL) is a no op for C99

hreinecke: Ok, ok. I've dropped that patch now.

PR #325, opened 2022-03-30
fabrics: Invoke nvmf_dim() with provided tas argument.

martin-belanger:
Looks like a copy-paste issue. One of the arguments to
nvmf_register_ctrl() is "tas" (i.e. the DIM Task). This argument
should be passed to the nvmf_dim() API, but instead a fixed value
of NVMF_DIM_TAS_REGISTER was passed.

The "tas" field specifies whether to perform a "registration",
"deregistration", or a "registration update".

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1306,6 +1306,5 @@ int nvmf_register_ctrl(nvme_ctrl_t c, enum nvmf_dim_tas tas, __u32 *result)
 	 * to retrieve the source address from the socket and use that
 	 * as the registration address.
 	 */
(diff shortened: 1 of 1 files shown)


igaw:
Patch looks good. Could change the subject title, e.g. `'s/Bug fix/fabrics:/'`. Thanks.

martin-belanger:
@igaw - I fixed the subject title

igaw:
Thanks! Sorry for being a bit annoying on these subjects but I think we should try to aim to get to higher contribution quality. And one of those is proper commit messages. Feel free to poke me, when I fail to follow my own rules :)

PR #326, opened 2022-03-31
Namespace path scanning rework

hreinecke:
The namespace paths are added as a bit of an afterthought, and are just cobbled on the namespaces.
This patch reworks the path logic to scan /sys/block, lookup matching path in the tree, and then adding it to the namespace.
With this we have a cross-check if controllers are scanned correctly, and we are insulated against any changes in the sysfs layout.


src/nvme/filters.c
```diff
@@ -116,3 +116,9 @@ int nvme_scan_ctrl_namespaces(nvme_ctrl_t c, struct dirent ***ns)
 	return scandir(nvme_ctrl_get_sysfs_dir(c), ns,
 		       nvme_namespace_filter, alphasort);
 }
+
+int nvme_scan_namespace_paths(struct dirent ***paths)
```
igaw, line 120: libnvme.map export is missing.

hreinecke: True.

hreinecke: Fixed with the updated patch.

hreinecke:
Turned out to be more complex as I thought, and also had looking me at the way we're handling paths currently. So better get the release out of the door, and have it merged post-release.

igaw:
Closing it as no one is working on it.

PR #327, opened 2022-03-31
Update function prototype documentation

hreinecke:
CI has been complaining, so fix things up.


igaw:
Thanks! This was my plan :) Although the CI will only complain for 10 warnings max. This is the current limit by github enforced.

PR #329, opened 2022-04-01
tree: always allocate config file in nvme_read_config()

hreinecke:
Even if there was an error when reading the config file we still
should store the config file name, as it might used later on to
update/write the config file.
So parsing errors don't really matter, and we'll get notified for
I/O errors on writing anyway.


hreinecke:
Fixes issue https://github.com/linux-nvme/nvme-cli/issues/1469

PR #330, opened 2022-04-01
Update nvme_scan_filter_t

hreinecke:
Update nvme_scan_filter_t to accept 4 arguments (subsystem, controller, namespace, f_args) and apply the filter for subsystem namespaces, too. This allows for a more targeted filter implementations and to remove the hack to display the ANA state.


igaw:
Yeah, this is way more generic and allows the user of the API to hold context without heaving to fallback using global variables.

PR #332, opened 2022-04-04
Remove default discovery port bits

igaw:
Trying to solve the problem reported in https://github.com/linux-nvme/nvme-cli/issues/1475


igaw:
This looks a bit dodgy. I think the best way would be to add a 'bool discovery_ctrl' to `nvme_create_ctrl`

igaw:
Or we move the whole default port settings into the frontend...

igaw:
With https://github.com/linux-nvme/nvme-cli/pull/1480 merged we don't do the default port setting in the library anymore.

PR #335, opened 2022-04-06
tree: add sysfs getters and setters

igaw:
Some code changes from working on  https://github.com/linux-nvme/nvme-cli/issues/1476 which might make
sense to add to the library.


igaw:
There is no hurry for this PR, as we don't change the existing API. We can still add it later.

igaw:
Bit rotting, no point in keeping it open

Issue #271, opened 2022-03-02
self: compare between `mesonpy` and `mesonpep517` python build backends


glimchb:
learn more and compare:
- https://pypi.org/project/meson-python/
- https://pypi.org/project/mesonpep517/

igaw:
We can't upload the binaries with mesonpep517, because we don't do the repair step. This leads to the situation that PyPi says `linux_x86_64` is not supported. The repair step would fix this up to `manylinux_`.

The repair step doesn't work because mesonpep517 is not packaging up libnvme.so only the Python c binding library is in the dist package. There is also no way to workaround in tricking meson to install libnvme.so into the python side-config directory because config_settings is not supported in v0.2. And there was no release for over one year. I guess we can say this project is not ready for this taks

mesonpy is not working either, at least not out of the box and I am running out of steam. I think we can't really upload precompiled binaries at this point. Uploading just the source code to PyPi is currently the only working solution I suppose.

(I've spend way too much time today to figure this out.)

glimchb:
thanks @igaw  indeed a lot o work you put into this and I put originally about the same amount of time
I don't object to your decision to remove PyPi integration attempt for now
We can re-introduce it once we figure out how to do it properly

igaw:
So the only thing I am still pondering around, if uploading just the sdist would actually make sense nonetheless. I am sure we are not the only one in this situation. Are there any projects which just upload the Python binding?

glimchb:
yes, I know many project just upload sdist without `wheels`
This means it will be built during `pip install`
`wheels` is just an optimization

igaw:
FWIW, I've uploaded the binding to PyPi manually (yeah, I know, sneaky thing...) but I wanted to get the project name claimed on PyPi before someone else is getting it. So the next release should work through the normal CI means.

glimchb:
just FYI,
I was struggling last time with wheels as well, that's why I left `CIBW_REPAIR_WHEEL_COMMAND_LINUX` explicitly empty
I was hoping that using `LD_LIBRARY_PATH` to pint to `libnvme.so` location in `CIBW_REPAIR_WHEEL_COMMAND_LINUX` we can make wheels to work

just documenting here for later... once we come back to this problem...

glimchb:
> FWIW, I've uploaded the binding to PyPi manually (yeah, I know, sneaky thing...) but I wanted to get the project name claimed on PyPi before someone else is getting it. So the next release should work through the normal CI means.

[...]

PR #339, opened 2022-04-12
ci: publish only sdist on PyPi

igaw:
Pre building the binaries with wheels is difficult, as we would also
need to ship the libnvme.so with the Python binding.

This is a well known limitation of this kind of setup (binding to a
shared library). Most project fallback to just ship the sdist.

Though there is a drawback as the C library is missing and the user
has to provide the library himself, with all problems which come along
with setup. But it seems common practice with other Python bindings,
so we don't want to be hostile to the ones which know what they are
doing.

Normal users should just use the distribution packages anyway.

Fixes: #271

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,67 +1,58 @@
-# This workflow will upload a Python Package using Twine when a release is created
-# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
-
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-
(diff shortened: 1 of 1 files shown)


igaw:
FWIW, when we figured out how we can get the versioning a bit more useful (looking into this as next) we can try to get TestPyPi hook more useful

glimchb (approved):
lgtm

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["mesonpep517", "wheel", "meson==0.61.2", "ninja"] # PEP 508 specifications.
```
glimchb, line 2: it's ok to leave wheel as dependency here, it will be used during `pip install` on the machine itself regardless of pypi

igaw: okay, I added it back.

PR #340, opened 2022-04-12
build: Add git ref to the binary

igaw:
In order to be able to figure out which binary is in use (for example
in debugging situation) it's really helpful to have the 'git describe'
ref added to the binary.

The simplest way to expose the it via good old Source Code Control
System string id. No need to come up with something complicated else
as there is no agreement on how to do this. So let's add this simple
magic string to library.

You can either query with 'what' or just

  $strings .build/src/libnvme.so | grep '@(#)'
  @(#)libnvme 1.0-1-g5ff5d22+

The idea how to extract the version string is shamelessly copied from
the systemd project.

### DIFF
--- a/meson-vcs-tag.sh
+++ b/meson-vcs-tag.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+set -eu
+set -o pipefail
+
+dir="${1:?}"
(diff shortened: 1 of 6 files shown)


igaw:
I was wondering if we should add something like a function like `const char *nvme_get_version(enum nvme_version_type type)` which could be used to query the version. I would pass in a type argument so that it would be possible to ask for different types of version e.g. project version and git version.

PR #341, opened 2022-04-12
ci: create first dependabot.yml

glimchb:


igaw:
Let's see how this works out.

igaw:
yep, we should start doing this also for nvme-cli.

glimchb:
will post now

PR #343, opened 2022-04-12
doc: Build ReST documantion as standalone target

igaw:
In order to feed readthedocs with documecation we have to build the
ReST documentation and store it the source tree.

To avoid cluttering the root dir of the documantion directory move the
generation of the *.rst docs into the rst subdir. Unfortunatly, meson
doesn't have an option to generate the newly files into subdir. So
move the rst build instruction into rst/meson.build.

While at it, split the whole documentation into sections.

We want to replace any VERSION string in the documantion. Instead just
addressing conf.py we process all files in one go for consistency reasons.


igaw:
This is to enable readthedocs to build/host the documetation. We need static rst files for this.

glimchb:
this is great!

PR #344, opened 2022-04-13
doc: add quickstart and installation

glimchb:


igaw:
Nice. Let's see what happens when I merge it. In theory the webhook to readthedocs should update the documentation.

glimchb:
I didn't know how to test this locally @igaw ...

igaw:
I was able to run it locally via `sphinx-build -a doc /tmp/mydocs -v ` after installing the toolchain

glimchb:
cool, it works! thanks for the tip

igaw:
BTW, the meson.build files and the update-docs.sh script are a bit hackish, it works but there is a lot room for improvements. Just wanted to get things going...

glimchb:
yep, I noticed ;-)

igaw:
Another thing we might want to do is to rename the 'rst' dir into something like 'api'...

igaw:
The build got triggered...

glimchb:
looks good https://libnvme.readthedocs.io/en/latest/index.html

PR #353, opened 2022-04-14
fabrics: Introduce connection connect error mapping

igaw:
The kernel returns status information via error codes. Currently we
map all error codes to ENVME_CONNECT_WRITE if the operation
fails. Some of those error codes should be treated differently, such
as EALREADY which means the connection attempt failed because the
connection is already established.

Many of the error codes are overloaded so we have to guess if they are
actually from the nvme subsystem and not from the write
command. Though in practice it's almost certainty the nvme
subsystem. Still it's a guessing game due the very limited API
between userland and kernel.

See: #https://github.com/linux-nvme/nvme-cli/issues/1505


tbzatek (approved):
This looks useful, thanks!

PR #354, opened 2022-04-15
fix broken includes for the uuid.h header

eli-schwartz:
Per `man 3 uuid`, the correct include is `<uuid.h>`, and correspondingly the util-linux build system installs the uuid.h file to
`$(includedir)/uuid/` and uuid.pc adds `-I${includedir}/uuid` to the dependency Cflags.

Including it as `<uuid/uuid.h>` simply fails altogether. The exception is if, by coincidence, the libuuid includedir happens to be the same as one of the compiler's own internal include directories... which is pretty common if that is `/usr/include`. In this case, `<uuid/uuid.h>` fails to be found in the uuid dependency's export directories, but the compiler picks up a copy in its builtin search path.

In other cases, such as when util-linux is compiled into a custom prefix and added to the PKG_CONFIG_PATH, it either:
- fails with missing header errors, if libuuid-dev is not installed
- seems to succeed, but compiles against a different version of the headers than the library that is linked to, resulting in potentially various problems from bizarre linker errors, to subtly broken incompatible symbols that crash, or corrupt the results

The problem is even worse if you try to use util-linux as a subproject dependency fallback. In this case, `<uuid.h>` is correctly exported, but there is no disk location for `<uuid/uuid.h>` and absolutely no way to generate such a location in the builddir. This completely prevents the use of subproject fallbacks.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -8,27 +8,27 @@
 executable(
     'telemetry-listen',
     ['telemetry-listen.c'],
-    link_with: libnvme,
+    dependencies: libnvme_dep,
     include_directories: [incdir, internal_incdir]
 )
 
 executable(
     'display-columnar',
     ['display-columnar.c'],
-    link_with: libnvme,
+    dependencies: libnvme_dep,
     include_directories: [incdir, internal_incdir]
 )
 
 executable(
     'display-tree',
     ['display-tree.c'],
-    link_with: libnvme,
+    dependencies: libnvme_dep,
     include_directories: [incdir, internal_incdir]
 )
 
 executable(
     'discover-loop',
     ['discover-loop.c'],
-    link_with: libnvme,
+    dependencies: libnvme_dep,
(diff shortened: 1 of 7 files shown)


eli-schwartz:
CI shows that I forgot to adapt the python extension.

eli-schwartz:
Let's try that again...

igaw:
I just have a browser access right now, so only github web button pushes for me :)

igaw:
I wonder if we are really depend on the ccan/list include in nmve.i. I'll add this to my todo list to check.

igaw:
Thanks a lot!

eli-schwartz:
At the moment I reverted back to not narrowing the list of include directories there, because I don't have swig locally to test that. Extra includes don't hurt all that much -- they don't break correct code and they aren't exposed as a public declare_dependency or pkg-config dependency. :D

PR #357, opened 2022-04-29
util: Add feature length for host memory buffer

t-8ch:
Without this the command "nvme get-feature /dev/nvme0 -H -f 0x0d"
segfaults because it retrieves no data but tries to use it in
nvme_show_host_mem_buffer()


igaw:
Thanks!

PR #358, opened 2022-04-29
Changes to support TP 4068 and TP 4084 

jeffreyalien:
TP 4068 - Protection Information Enchancement
TP 4084 - Time-to-Ready Enhancements

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,6 +1,7 @@
 LIBNVME_1_1 {
 	global:
 		nvme_get_version;
+		nvme_init_copy_range_f1;
 };
 
 LIBNVME_1_0 {
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -267,11 +267,13 @@ enum nvme_cmd_dword_fields {
 	NVME_FORMAT_CDW10_PI_SHIFT				= 5,
 	NVME_FORMAT_CDW10_PIL_SHIFT				= 8,
 	NVME_FORMAT_CDW10_SES_SHIFT				= 9,
+	NVME_FORMAT_CDW10_LBAFU_SHIFT				= 12,
 	NVME_FORMAT_CDW10_LBAF_MASK				= 0xf,
 	NVME_FORMAT_CDW10_MSET_MASK				= 0x1,
 	NVME_FORMAT_CDW10_PI_MASK				= 0x7,
 	NVME_FORMAT_CDW10_PIL_MASK				= 0x1,
 	NVME_FORMAT_CDW10_SES_MASK				= 0x7,
+	NVME_FORMAT_CDW10_LBAFU_MASK				= 0x3,
 	NVME_SANITIZE_CDW10_SANACT_SHIFT			= 0,
 	NVME_SANITIZE_CDW10_AUSE_SHIFT				= 3,
 	NVME_SANITIZE_CDW10_OWPASS_SHIFT			= 4,
@@ -1159,11 +1161,25 @@ int nvme_get_features_iocs_profile(int fd, enum nvme_get_features_sel sel,
 
 int nvme_format_nvm(struct nvme_format_nvm_args *args)
 {
-	__u32 cdw10 = NVME_SET(args->lbaf, FORMAT_CDW10_LBAF) |
-			NVME_SET(args->mset, FORMAT_CDW10_MSET) |
-			NVME_SET(args->pi, FORMAT_CDW10_PI) |
-			NVME_SET(args->pil, FORMAT_CDW10_PIL) |
-			NVME_SET(args->ses, FORMAT_CDW10_SES);
+	const size_t size_v1 = sizeof_args(struct nvme_format_nvm_args, lbaf, __u64);
+	const size_t size_v2 = sizeof_args(struct nvme_format_nvm_args, lbafu, __u64);
+	__u32 cdw10;
+
+	if (args->args_size < size_v1 || args->args_size > size_v2) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	cdw10 = NVME_SET(args->lbaf, FORMAT_CDW10_LBAF) |
+		NVME_SET(args->mset, FORMAT_CDW10_MSET) |
+		NVME_SET(args->pi, FORMAT_CDW10_PI) |
+		NVME_SET(args->pil, FORMAT_CDW10_PIL) |
+		NVME_SET(args->ses, FORMAT_CDW10_SES);
+
+	if (args->args_size == size_v2) {
+		/* set lbafu extension */
+		cdw10 |= NVME_SET(args->lbafu, FORMAT_CDW10_LBAFU);
+	}
 
 	struct nvme_passthru_cmd cmd = {
 		.opcode		= nvme_admin_format_nvm,
@@ -1172,10 +1188,6 @@ int nvme_format_nvm(struct nvme_format_nvm_args *args)
 		.timeout_ms	= args->timeout,
 	};
 
-	if (args->args_size < sizeof(*args)) {
-		errno = EINVAL;
-		return -1;
-	}
 	return nvme_submit_admin_passthru(args->fd, &cmd, args->result);
 }
 
@@ -1588,16 +1600,81 @@ int nvme_io_passthru(int fd, __u8 opcode, __u8 flags, __u16 rsvd,
 			     timeout_ms, result);
 }
 
+static int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw14,
+		__u8 pif, __u8 sts, __u64 reftag, __u64 storage_tag)
+{
+	__u32 cdw2 = 0, cdw3 = 0, cdw14;
+
+	switch (pif) {
+	/* 16b Protection Information */
+	case 0:
+		cdw14 = reftag & 0xffffffff;
+		cdw14 |= ((storage_tag << (32 - sts)) & 0xffffffff);
+		break;
+	/* 32b Protection Information */
+	case 1:
+		cdw14 = reftag & 0xffffffff;
+		cdw3 = reftag >> 32;
+		cdw14 |= ((storage_tag << (80 - sts)) & 0xffff0000);
+		if (sts >= 48)
+			cdw3 |= ((storage_tag >> (sts - 48)) & 0xffffffff);
+		else
+			cdw3 |= ((storage_tag << (48 - sts)) & 0xffffffff);
+		cdw2 = (storage_tag >> (sts - 16)) & 0xffff;
+		break;
+	/* 64b Protection Information */
+	case 2:
+		cdw14 = reftag & 0xffffffff;
+		cdw3 = (reftag >> 32) & 0xffff;
+		cdw14 |= ((storage_tag << (48 - sts)) & 0xffffffff);
+		if (sts >= 16)
+			cdw3 |= ((storage_tag >> (sts - 16)) & 0xffff);
+		else
+			cdw3 |= ((storage_tag << (16 - sts)) & 0xffff);
+		break;
+	default:
+		perror("Unsupported Protection Information Format");
+		errno = EINVAL;
+		return -1;
+	}
+
+	*cmd_dw2 = cdw2;
+	*cmd_dw3 = cdw3;
+	*cmd_dw14 = cdw14;
+	return 0;
+}
+
 int nvme_io(struct nvme_io_args *args, __u8 opcode)
 {
-	__u32 cdw2  = args->storage_tag & 0xffffffff;
-	__u32 cdw3  = (args->storage_tag >> 32) & 0xffff;
-	__u32 cdw10 = args->slba & 0xffffffff;
-	__u32 cdw11 = args->slba >> 32;
-	__u32 cdw12 = args->nlb | (args->control << 16);
-	__u32 cdw13 = args->dsm | (args->dspec << 16);
-	__u32 cdw14 = args->reftag;
-	__u32 cdw15 = args->apptag | (args->appmask << 16);
+	const size_t size_v1 = sizeof_args(struct nvme_io_args, dsm, __u64);
+	const size_t size_v2 = sizeof_args(struct nvme_io_args, pif, __u64);
+	__u32 cdw2, cdw3, cdw10, cdw11, cdw12, cdw13, cdw14, cdw15;
+
+	if (args->args_size < size_v1 || args->args_size > size_v2) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	cdw10 = args->slba & 0xffffffff;
+	cdw11 = args->slba >> 32;
+	cdw12 = args->nlb | (args->control << 16);
+	cdw13 = args->dsm | (args->dspec << 16);
+	cdw15 = args->apptag | (args->appmask << 16);
+
+	if (args->args_size == size_v1) {
+		cdw2 = (args->storage_tag >> 32) & 0xffff;
+		cdw3 = args->storage_tag & 0xffffffff;
+		cdw14 = args->reftag;
+	} else {
+		if (nvme_set_var_size_tags(&cdw2, &cdw3, &cdw14,
+				args->pif,
+				args->sts,
+				args->reftag_u64,
+				args->storage_tag)) {
+			errno = EINVAL;
+			return -1;
+		}
+	}
 
 	struct nvme_passthru_cmd cmd = {
 		.opcode		= opcode,
@@ -1617,10 +1694,6 @@ int nvme_io(struct nvme_io_args *args, __u8 opcode)
 		.timeout_ms	= args->timeout,
 	};
 
-	if (args->args_size < sizeof(*args)) {
-		errno = EINVAL;
-		return -1;
-	}
 	return nvme_submit_io_passthru(args->fd, &cmd, args->result);
 }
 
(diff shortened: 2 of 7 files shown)


bjpaupor (approved):

src/nvme/types.h
```diff
@@ -247,11 +250,14 @@ enum nvme_cap {
 	NVME_CAP_MPSMAX_MASK		= 0xf,
 	NVME_CAP_PMRS_MASK		= 0x1,
 	NVME_CAP_CMBS_MASK		= 0x1,
+	NVME_CAP_CRMS_MASK		= 0x1,
```
igaw, line 253: I think the mask should be 0x3 as `CRMS` is two bits widths.

src/nvme/types.h
```diff
@@ -308,6 +317,8 @@ enum nvme_cc {
 	NVME_CC_SHN_NONE	= 0,
 	NVME_CC_SHN_NORMAL	= 1,
 	NVME_CC_SHN_ABRUPT	= 2,
+	NVME_CC_CRWME       = 0,
+	NVME_CC_CRIME       = 1,
```
igaw, line 321: Please align the = with the other ones right above.

src/nvme/ioctl.c
```diff
@@ -1597,12 +1597,12 @@ int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw14,
 	__u32 cdw2 = 0, cdw3 = 0, cdw14;
 
 	switch (pif) {
-	// 16b Protection Information
+	/* 16b Protection Information */
```
igaw, line 1600: Please merge these code fixups into previous patch.

src/libnvme.map
```diff
@@ -175,6 +175,7 @@ LIBNVME_1_0 {
 		nvme_identify_secondary_ctrl_list;
 		nvme_identify_uuid;
 		nvme_init_copy_range;
+		nvme_init_copy_range_f1;
```
igaw, line 178: This has to go a new section `LIBNVME_1_1`. I think we should also add this to the function documentation which which version a new function was introduced.

src/nvme/ioctl.c
```diff
@@ -262,16 +262,18 @@ enum nvme_cmd_dword_fields {
 	NVME_VIRT_MGMT_CDW10_RT_MASK				= 0x7,
 	NVME_VIRT_MGMT_CDW10_CNTLID_MASK			= 0xffff,
 	NVME_VIRT_MGMT_CDW11_NR_MASK				= 0xffff,
-	NVME_FORMAT_CDW10_LBAF_SHIFT				= 0,
+	NVME_FORMAT_CDW10_LBAFL_SHIFT				= 0,
```
igaw, line 265: I fear we can't just rename definitions anymore. This breaks backwards compatibility.

src/nvme/ioctl.c
```diff
@@ -262,16 +262,18 @@ enum nvme_cmd_dword_fields {
...
 	NVME_FORMAT_CDW10_MSET_SHIFT				= 4,
 	NVME_FORMAT_CDW10_PI_SHIFT				= 5,
 	NVME_FORMAT_CDW10_PIL_SHIFT				= 8,
 	NVME_FORMAT_CDW10_SES_SHIFT				= 9,
-	NVME_FORMAT_CDW10_LBAF_MASK				= 0xf,
```
igaw, line 270: Same here, we need to leave the NVME_FORMAT_CDW10_LBAF_MASK in place. But please update the documentation that it is deprecated.

src/nvme/ioctl.h
```diff
@@ -2917,8 +2918,9 @@ struct nvme_format_nvm_args {
 	enum nvme_cmd_format_pi pi;
 	enum nvme_cmd_format_pil pil;
 	enum nvme_cmd_format_ses ses;
-	__u8 lbaf;
-};
+	__u8 lbafu;
+	__u8 lbafl;
+} __attribute__((packed, aligned(__alignof__(__u32*))));
```
igaw, line 2923: No attributes are needed. We sorted the members according their natural size so we don't have any holes. But given
the argument structure size changes (our versioning scheme) we need also backwards compatibility, which seems to miss from this patch.

src/nvme/ioctl.h
```diff
@@ -3871,7 +3877,7 @@ struct nvme_io_args {
 	int fd;
 	__u32 timeout;
 	__u32 nsid;
-	__u32 reftag;
+	__u64 reftag;
```
igaw, line 3880: This type of change is not possible with the 1.0 release. This breaks our stable APi guarantee.

src/nvme/ioctl.h
```diff
@@ -3880,7 +3886,9 @@ struct nvme_io_args {
 	__u16 appmask;
 	__u16 dspec;
 	__u8 dsm;
-};
+	__u8 sts;
+	__u8 pif;
+} __attribute__((__packed__, aligned(__alignof__(__u64))));
```
igaw, line 3891: Again no need for these attributes.

src/nvme/ioctl.h
```diff
@@ -4044,7 +4052,7 @@ struct nvme_copy_args {
 	int fd;
 	__u32 timeout;
 	__u32 nsid;
-	__u32 ilbrt;
+	__u64 ilbrt;
```
igaw, line 4055: This breaks the stable API.

src/nvme/ioctl.h
```diff
@@ -4357,7 +4365,7 @@ struct nvme_zns_append_args {
...
```
igaw, line 4368: Here too.

igaw (changes requested):
Unfortunately, this PR breaks the stable API guarantee we gave with the initial 1.0 release. You need to add these changes in a way that it is backwards compatible. E.g. when a struct *_args is extended we have to support the old version as the new one. So there needs some sort of compatibility code introduced:
```

if (args->size == v1_size) {
  /* access members of v1 */
} else if (args->size = v2_size) {
  /* access members up to v2 */
}
```

src/nvme/ioctl.h
jeffreyalien: In order to check the structure size (to determine version 1 or 2), I needed to keep the packed attribute.  Without it, the compiler would pad it out to the next 4 byte boundary.

src/nvme/types.h
jeffreyalien: Fixed this

src/nvme/ioctl.c
jeffreyalien: Fixed this in original commit.

src/libnvme.map
jeffreyalien: Moved nvme_init_copy_range_f1 into the LIBNVME_1_1 section.  I will update the function documentation in a separate commit.

src/nvme/ioctl.c
jeffreyalien: Fixed this - decided to use the old name for both cases to keep it backward compatible.

src/nvme/ioctl.h
jeffreyalien: Fixed this but had to keep the packed attribute so the exact structure size could be used to differentiate between the version in use.

src/nvme/ioctl.h
jeffreyalien: Made changes to set the reftag field accordingly based on the version of the structure.

src/nvme/ioctl.h
jeffreyalien: Made changes to set the ilbrt field accordingly based on the version of the structure.

src/nvme/ioctl.h
jeffreyalien: Made changes to set the ilbrt field accordingly based on the version of the structure.

src/nvme/ioctl.h
jeffreyalien: I had to keep the packed attribute to get the exact structure size.  Without it the structure is padded out to a 4 byte boundary and the size is not different between the old and new versions of the struct.

jeffreyalien:
I see the packed attribute is causing CI jobs to fail because of warning messages.  I'll have to figure out a different fix to determine the structure size of for each version.

igaw:
> I see the packed attribute is causing CI jobs to fail because of warning messages. I'll have to figure out a different fix to determine the structure size of for each version.

Maybe this could work:

```
if  (args->size == offsetof(last_member_of_struct_v1))
  // version 1
else if (args->size == offsetof(last_member_of_struct_v2))
  // version 2
```

src/nvme/ioctl.h
igaw: Oh, you are right. I hoped we don't need those attributes...

src/nvme/ioctl.h
igaw: Okay, I see the problem now.

jeffreyalien:
@igaw Looks like I can fix the warning messages in the CI process by removing the packed attribute from the nvme_io_args structure.  The nvme_copy_args structure also does not need the packed attribute.  The nvme_format_nvm_args and nvme_zns_append_args structures still need it though.  This should resolve the failures from the CI jobs - they were all from references to pointers to specific fields in the nvme_io_args structure.  Does that sound acceptable to you?

igaw:
I think you need to add explicit padding for the new member you want to add if it doesn't align naturally. Let's say you have this here

v1:
```c
struct foo_args {
  u64 bar;
  u32 baz;
};
```

and now you want to add another u64 member.

v2:
```c
struct foo_args {
  u64 bar;
  u32 baz;
  u8 rsvd6[4];
  u64 bax;
};
```

The compiler naturally adds padding to the v1 version of the struct, hence we need to add this to our calculation as well. Something like does the trick:

```c
#include <stddef.h>                                                                                                                                             
#include <stdio.h>                                                                                                                                              
#include <stdint.h>                                                                                                                                             
                                                                                                                                                                
typedef uint64_t u64;                                                                                                                                           
typedef uint32_t u32;                                                                                                                                           
typedef uint8_t u8;                                                                                                                                             
                                                                                                                                                                
struct foo_args_v1 {                                                                                                                                            
  u64 bar;                                                                                                                                                      
  u32 baz;                                                                                                                                                      
};                                                                                                                                                              
                                                                                                                                                                
struct foo_args_v2 {                                                                                                                                            
  u64 bar;                                                                                                                                                      
  u32 baz;                                                                                                                                                      
  u8  rsvd6[4];                                                                                                                                                 
  u64 box;                                                                                                                                                      
};                                                                                                                                                              
                                                                                                                                                                
int main(int argc, char *argv[])                                                                                                                                
{                                                                                                                                                               
        int pad;                                                                                                                                                
        int baz;                                                                                                                                                
                                                                                                                                                                
        baz = offsetof(struct foo_args_v1, baz) + sizeof(u32);                                                                                                  
        pad = (sizeof(u64) - (baz%sizeof(u64))) % sizeof(u64);                                                                                                  
        printf("v1: baz %d pad %d\n", baz, pad);                                                                                                                
                                                                                                                                                                
        printf("v1: sizeof(struct foo_args) = %ld\n", sizeof(struct foo_args_v1));                                                                              
        printf("v1: offsetof(struct foo_args, baz) + sizeof(baz) + pad = %ld\n", offsetof(struct foo_args_v1, baz) + sizeof(u32) + pad);                        
                                                                                                                                                                
                                                                                                                                                                
        baz = offsetof(struct foo_args_v2, baz) + sizeof(u32);                                                                                                  
        pad = (sizeof(u64) - (baz%sizeof(u64))) % sizeof(u64);                                                                                                  
        printf("v2: baz %d pad %d\n", baz, pad);                                                                                                                
        printf("v2: sizeof(struct foo_args) = %ld\n", sizeof(struct foo_args_v2));                                                                              
        printf("v2: offsetof(struct foo_args, baz) + sizeof(baz) + pad  = %ld\n", offsetof(struct foo_args_v2, baz) + sizeof(u32) + pad);                       
        printf("v2: offsetof(struct foo_args, box) + sizeof(box) = %ld\n", offsetof(struct foo_args_v2, box) + sizeof(u64));                                    
}                                                                                                                                                               
```

```
v1: baz 12 pad 4
v1: sizeof(struct foo_args) = 16
v1: offsetof(struct foo_args, baz) + sizeof(baz) + pad = 16
v2: baz 12 pad 4
v2: sizeof(struct foo_args) = 24
v2: offsetof(struct foo_args, baz) + sizeof(baz) + pad  = 16
v2: offsetof(struct foo_args, box) + sizeof(box) = 24
```

We want probably to introduce a helper function for this kind of calculation

```c
#define sizeof_args(type, member, align)					\
  ({										\
	  type s;								\
	  size_t t = offsetof(type, member) + sizeof(s.member);			\
	  size_t p = (sizeof(align) - (t % sizeof(align))) % sizeof(align);	\
	  t + p;								\
  })

 printf("v1: sizeof_args(struct foo_args, baz, u64) = %ld\n", sizeof_args(struct foo_args_v1, baz, u64));
 printf("v2: sizeof_args(struct foo_args, baz, u64) = %ld\n", sizeof_args(struct foo_args_v2, baz, u64));                                                            
 printf("v2: sizeof_args(struct foo_args, box, u64) = %ld\n", sizeof_args(struct foo_args_v2, box, u64)); 
```

```
v1: baz 12 pad 4
v1: sizeof(struct foo_args) = 16
v1: offsetof(struct foo_args, baz) + sizeof(baz) + pad = 16
v1: sizeof_args(struct foo_args, baz, u64) = 16
v2: baz 12 pad 4
v2: sizeof(struct foo_args) = 24
v2: offsetof(struct foo_args, baz) + sizeof(baz) + pad  = 16
v2: offsetof(struct foo_args, box) + sizeof(box) = 24
v2: sizeof_args(struct foo_args, baz, u64) = 16
v2: sizeof_args(struct foo_args, box, u64) = 24
```

jeffreyalien:
Thanks for the feedback @igaw.  I was hoping we could make a simpler fix but don't see how.  I'll make changes based on this design and update the PR.

jeffreyalien:
@igaw I think I have all the issues resolved now and have verified nvme cli with both libnvme 1.0 and 1.1 versions.  Let me know if you have any comments or find any issues I missed.  Thanks

igaw (changes requested):
Almost there! Sorry for being pedantic but I want to get the versioning handling into good shape so that we have a blueprint for future changes.

src/nvme/ioctl.c
```diff
@@ -1617,10 +1669,26 @@ int nvme_io(struct nvme_io_args *args, __u8 opcode)
...
+			errno = EINVAL;
+			return -1;
+		}
+	} else {
+		cmd.cdw2 = args->storage_tag & 0xffffffff;
+		cmd.cdw3 = (args->storage_tag >> 32) & 0xffff;
+		cmd.cdw14 = args->reftag;
+	}
+
 	return nvme_submit_io_passthru(args->fd, &cmd, args->result);
```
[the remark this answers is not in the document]
bjpaupor: Also the original changes were addressing a bug in the current handling of storage_tag, in that the spec defines both variable-sized tags as able to take the back 16 bits of cdw2, the full cdw3, and full cdw14. 

Calling it a "storage tag" feels a bit wrong when it could be entirely part of a larger reference tag, but in any case it shouldn't be setting the full bits of cdw2 and should be setting cdw3.

src/nvme/ioctl.c
```diff
@@ -1172,7 +1184,9 @@ int nvme_format_nvm(struct nvme_format_nvm_args *args)
 		.timeout_ms	= args->timeout,
 	};
 
-	if (args->args_size < sizeof(*args)) {
+	if (args->args_size < sizeof_args(struct nvme_format_nvm_args, lbaf, __u64)) {
+		printf("%s: io_args < sizeof_args), %d < %ld\n",
```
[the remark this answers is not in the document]
jeffreyalien: Changed the code to the suggested way above.

src/nvme/ioctl.c
```diff
@@ -1588,27 +1602,65 @@ int nvme_io_passthru(int fd, __u8 opcode, __u8 flags, __u16 rsvd,
 			     timeout_ms, result);
 }
 
+int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw14,
```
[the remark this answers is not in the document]
jeffreyalien: Done

src/nvme/ioctl.c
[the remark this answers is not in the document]
jeffreyalien: Changed the code to the suggested way above.  Also fixed the original leg to be correct based on @bjpaupor  comment.

src/nvme/ioctl.c
```diff
@@ -1654,20 +1722,33 @@ int nvme_copy(struct nvme_copy_args *args)
...
 		return -1;
 	}
+
+	if (args->args_size == sizeof_args(struct nvme_copy_args, ilbrt_u64, __u64)) {
+		cmd.cdw3 = (args->ilbrt_u64 >> 32) & 0xffffffff;
+		cmd.cdw14 = args->ilbrt_u64 & 0xffffffff;
+	} else {
+		cmd.cdw14 = args->ilbrt;
+	}
+
```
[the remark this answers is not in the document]
jeffreyalien: Changed the code to the suggested way above.

src/nvme/types.h
```diff
@@ -4300,7 +4313,28 @@ struct nvme_copy_range {
 	__le64			slba;
 	__le16			nlb;
 	__u8			rsvd18[6];
-	__le32			eilbrt;
```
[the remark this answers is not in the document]
jeffreyalien: Added eilbrt back into the structure.

src/nvme/ioctl.h
```diff
@@ -4044,7 +4073,7 @@ struct nvme_copy_args {
 	int fd;
 	__u32 timeout;
 	__u32 nsid;
-	__u32 ilbrt;
+ 	__u32 ilbrt;
```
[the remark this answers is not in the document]
igaw: fixed this myself

igaw:
I've fixed a few whitespace and docs issues. No code changes.

igaw (approved):

PR #359, opened 2022-05-02
ci: add static library compilation

glimchb:
Now possible since we can use uuid for fallback

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -48,6 +48,26 @@ jobs:
           name: Linux_Meson_log
           path: build/meson-logs/meson-log.txt
 
+  build-static:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


igaw:
I was wondering if we should build libnvme also as static library in this build job.

glimchb:
we are, aren't we?
After the build is done:
```
 $ ar -t .build/src/libnvme.a
nvme_cleanup.c.o
nvme_fabrics.c.o
nvme_filters.c.o
nvme_ioctl.c.o
nvme_linux.c.o
nvme_log.c.o
nvme_tree.c.o
nvme_util.c.o
nvme_json.c.o
ccan_list_list.c.o
ccan_str_debug.c.o
ccan_str_str.c.o
```

glimchb:
Will open PR now

PR #360, opened 2022-05-13
tree: avoid segfault if subsysnqn is not available

martin-gpy:
A nvme config command specifying the -M & -J options would end
up in a segfault if subsysnqn is NULL, as shown below:

nvme config -M -J /etc/nvme/config.json

Avoid this segfault by proceeding with the subsysnqn strcmp in
nvme_lookup_subsystem() only if subsysnqn is available.


igaw:
Thanks!

PR #367, opened 2022-06-03
libnvme: export nvme_free_dirents

sc108-lee:
it should be called after called nvme_scan_* that use dirents

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -4,6 +4,7 @@ LIBNVME_1_1 {
 	global:
 		nvme_get_version;
 		nvme_init_copy_range_f1;
+		nvme_free_dirents;
 };
 
 LIBNVME_1_0 {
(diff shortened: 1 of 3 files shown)


src/libnvme.map
```diff
@@ -71,6 +71,7 @@ LIBNVME_1_0 {
 		nvme_flush;
 		nvme_format_nvm;
 		nvme_free_ctrl;
+		nvme_free_dirents;
```
igaw, line 74: This should go into the LIBNVME_1_1 section

igaw (changes requested):

src/libnvme.map
sc108-lee: nice catch! updated

sc108-lee:
Close this since its included in #368

PR #368, opened 2022-06-03
Get registers

sc108-lee:


src/nvme/linux.c
```diff
@@ -494,6 +496,139 @@ char *nvme_get_path_attr(nvme_path_t p, const char *attr)
...
+
+	for (i = 0; i < num_ctrls; i++) {
+		if (strncmp(dev, ctrls[i]->d_name, strlen(ctrls[i]->d_name)) == 0) {
+			base = ctrls[i]->d_name;
+			break;
+		}
+	}
+
+	if (!base) {
+		fprintf(stderr, "Can't find %s in scaned ctrls\n", dev);
```
martin-belanger, line 588: minor typo: `scaned` -> `scanned`

src/nvme/linux.c
```diff
@@ -494,6 +496,139 @@ char *nvme_get_path_attr(nvme_path_t p, const char *attr)
...
+		goto free_ctrls;
+	}
+
+	sprintf(path, "/sys/class/nvme/%s/transport", base);
```
martin-belanger, line 592: In `libnvme`, temporary strings are usually dynamically allocated with `asprintf()`. Obviously, this also requires a `free()` before exiting.

src/nvme/linux.h
```diff
@@ -176,6 +176,25 @@ int nvme_namespace_detach_ctrls(int fd, __u32 nsid, __u16 num_ctrls, __u16 *ctrl
  */
 int nvme_open(const char *name);
 
+/**
+ * nvme_get_registers() - Read registers mmap or with fabrics command
+ * @fd:			Opended target device file descriptor
```
martin-belanger, line 181: typo: `Opended` -> `Opened`

src/nvme/linux.c
sc108-lee: thanks for checking fixed :)

src/nvme/linux.c
sc108-lee: thanks for checking fixed :)

src/nvme/linux.h
sc108-lee: thanks for checking fixed :)

igaw:
This PR is part of https://github.com/linux-nvme/nvme-cli/pull/1541

The current head doesn't issue any ioctl anymore when scanning the topology. Thus we don't need add this here anymore.

PR #370, opened 2022-06-05
types.h: add missing fields for identify controller DS

ankit-sam:
Various fields and their description are missing from identify controller data structure. This patch includes all those.


src/nvme/types.h
```diff
@@ -1253,6 +1259,7 @@ enum nvme_id_ctrl_oacs {
 	NVME_CTRL_OACS_VIRT_MGMT		= 1 << 7,
 	NVME_CTRL_OACS_DBBUF_CFG		= 1 << 8,
 	NVME_CTRL_OACS_LBA_STATUS		= 1 << 9,
+	NVME_CTRL_OACS_CMD_FEAT_LD		= 1 << 10,
```
igaw, line 1262: If I am not completely mistaken, this has nothing to do with DS, right? Could you split these changes into smaller commits which describe each changes.

src/nvme/types.h
```diff
@@ -1263,28 +1270,53 @@ enum nvme_id_ctrl_oacs {
...
+ * 					area 4 for telemetry host-initiated and
+ * 					telemetry.
  */
 enum nvme_id_ctrl_lpa {
 	NVME_CTRL_LPA_SMART_PER_NS		= 1 << 0,
 	NVME_CTRL_LPA_CMD_EFFECTS		= 1 << 1,
 	NVME_CTRL_LPA_EXTENDED			= 1 << 2,
 	NVME_CTRL_LPA_TELEMETRY			= 1 << 3,
 	NVME_CTRL_LPA_PERSETENT_EVENT		= 1 << 4,
+	NVME_CTRL_LPA_LI0_LI5_LI12_LI13		= 1 << 5,
```
igaw, line 1318: same here, the change is okay but it doesn't match the commit message.

igaw (changes requested):
Please split the changes and have a matching commit message. Thanks

ankit-sam:
> Please split the changes and have a matching commit message. Thanks

Hi Daniel,

Thanks for the review comments.
I have split the patch into 6 small patches, please have a look.

igaw:
Excellent! Thanks for the nice series. This is way simpler to review it too!

igaw (approved):

PR #371, opened 2022-06-08
meson: only build test/cpp if a c++ compiler is available

jk-ozlabs:
Currently, the meson project configuration specifies both 'c' and 'ccp'
languages. We use the latter to build a single test binary.

This change makes the cpp language optional, so we can compile the rest
of libnvme without a C++ environment.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,7 +6,7 @@
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
 project(
-    'libnvme', ['c', 'cpp'],
+    'libnvme', ['c'],
     meson_version: '>= 0.47.0',
     version: '1.0',
     license: 'LGPL-2.1-or-later',
@@ -22,6 +22,7 @@ library_version = meson.project_version() + '.0'
 
 ################################################################################
 cc = meson.get_compiler('c')
+cxx_available = add_languages('cpp', required: false)
 
 prefixdir  = get_option('prefix')
 libdir     = join_paths(prefixdir, get_option('libdir'))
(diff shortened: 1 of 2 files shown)


jk-ozlabs:
An alternative (with the same outcome, of being able to build without a C++ env) would be to remove the `test/cpp` binary. Is this still useful?

martin-belanger:
> An alternative (with the same outcome, of being able to build without a C++ env) would be to remove the `test/cpp` binary. Is this still useful?

The tests are still useful. The reason they have been turned off (for now) is because GitHub Actions run on a version of Ubuntu that doesn't have the nvme-fabrics driver (i.e. the interface `/dev/nvme-fabrics` is missing). However, I've been experimenting with `pyfakefs` with my project ([nvme-stas](https://github.com/linux-nvme/nvme-stas)) to "fake" a `/dev/nvme-fabrics`. It's working pretty well, and we may be able to use it for `libnvme` as well. I will try it when I get a chance, and if it works I'll submit a patch.

jk-ozlabs:
> The tests are still useful.

Oh, I'm not proposing we remove the tests - just whether this single C++ binary was particularly useful as a test.

We don't generate any C++-specific bindings, or have any particular C++ support in the core library. This is mostly just a re-implementation of a small part of what's already in `test/test.c`.

I guess it does bring a bit of value in that the library headers are buildable as a C++ interface though; in which case we may not want to delete it.

igaw:
Depending on the C++ compiler just for the test is a bit silly. The CI build environment comes with a C++ compiler so there is no change except we don't ask for the compiler.

PR #374, opened 2022-06-10
fabrics: do not retrieve discovery log page twice

hreinecke:
When retrieving the discovery log page we cannot re-issue the
'get log page' command with just the header, as that would overwrite
the number of records in the prior attempt, causing the log
page evaluation to fail with strange error messages like
'can only fetch 8 of 1 records'.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/1555

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -827,18 +827,12 @@ int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
 				 name, retries, max_retries, errno);
 			goto out_free_log;
 		}
-
+		if (genctr == le64_to_cpu(log->genctr))
+			break;
 		genctr = le64_to_cpu(log->genctr);
 		nvme_msg(r, LOG_DEBUG,
 			 "%s: discover genctr %" PRIu64 ", retry\n",
 			 name, genctr);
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -827,18 +827,12 @@ int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
...
 			goto out_free_log;
 		}
-
+		if (genctr == le64_to_cpu(log->genctr))
+			break;
 		genctr = le64_to_cpu(log->genctr);
 		nvme_msg(r, LOG_DEBUG,
 			 "%s: discover genctr %" PRIu64 ", retry\n",
 			 name, genctr);
-		ret = nvme_discovery_log(nvme_ctrl_get_fd(c), hdr, log, true);
```
igaw, line 835: But what is updating the `genctr` if has changed?

igaw:
My attempt to address this: #378

src/nvme/fabrics.c
hreinecke: Hmm. Yes, slightly more massaging is required here.

igaw:
Applied https://github.com/linux-nvme/libnvme/pull/378 which replaces this PR.

PR #378, opened 2022-06-13
fabrics: restructrure nvmf_get_discovery_log

igaw:
Move the initial header read into the do while loop. This simplifies
the loop flow into first read header if there are records fetch them
too.

With this we keep the genctr and numrec in sync for both get log
invocation.

Fixes: #https://github.com/linux-nvme/nvme-cli/issues/1555

```bash
localhost:~/nvme-cli # .build/nvme discover -t tcp -a 10.162.198.42 -vvvv
Error opening /etc/nvme/config.json, No such file or directory
connect ctrl, 'nqn=nqn.2014-08.org.nvmexpress.discovery,transport=tcp,traddr=10.162.198.42,trsvcid=8009,hostnqn=nqn.2014-08.org.nvmexpress:uuid:2cd2c43b-a90a-45c1-a8cd-86b33ab273b5,hostid=2cd2c43b-a90a-45c1-a8cd-86b33ab273b5,ctrl_loss_tmo=600'
connect ctrl, response 'instance=0,cntlid=1'
nvme0: ctrl connected
lookup subsystem /sys/class/nvme-subsystem/nvme-subsys0/nvme0
nvme0: get header (try 0/10)
nvme0: get header and 10 records (length 11264 genctr 2)
```

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -773,54 +773,27 @@ int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
 			   int max_retries)
 {
 	nvme_root_t r = c->s && c->s->h ? c->s->h->r : NULL;
-	struct nvmf_discovery_log *log;
-	int hdr, ret, retries = 0;
+	struct nvmf_discovery_log *log = NULL;
+	int ret, retries = 0;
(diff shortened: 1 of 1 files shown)


igaw:
@hreinecke any objection on this version? I'd like to get the bug report fixed. I have tested it and it should make things worse.

PR #382, opened 2022-06-13
build: add 'uninstall' to Makefile

martin-belanger:
Changes in this patch:

- This patch adds an `uninstall` target to the Makefile, which only had an `install` target but no `uninstall`. Note that for `uninstall` to work, an `install` must have previously been performed. meson/ninja keep installation data in the build directory that is required in order to perform an uninstall.
- Fixed dependency issues for `install`, `test`, and `dist`. These targets depend on `libnvme` being built. The previous dependency just required the configuration (meson) as the dependency and not actually building `libnvme`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,13 +30,21 @@ ifneq ("$(wildcard ${BUILD-DIR})","")
 	rm -rf ${BUILD-DIR}
 endif
 
-.PHONY: install dist
-install dist: ${BUILD-DIR}
-	cd ${BUILD-DIR} && meson $@
+.PHONY: install
+install: ${NAME}
+	cd ${BUILD-DIR} && sudo meson $@
+
+.PHONY: uninstall
+uninstall: ${BUILD-DIR}
+	sudo ninja -C ${BUILD-DIR} $@
 
 .PHONY: test
-test: ${BUILD-DIR}
+test: ${NAME}
 	ninja -C ${BUILD-DIR} $@
+
+.PHONY: dist
+dist: ${NAME}
+	cd ${BUILD-DIR} && meson $@
 
 .PHONY: rpm
 rpm: dist


eli-schwartz:
> Fixed dependency issues for `install`, `test`, and `dist`. These targets depend on `libnvme` being built. The previous dependency just required the configuration (meson) as the dependency and not actually building `libnvme`

For clarity, this is absolutely wrong, though not always for the same reason.
- `dist` does not depend on anything being built
- `install` and `dist` do depend on the project being built, but both will also internally launch `ninja` to rebuild, whether you need it or not, unless also invoked with `--no-rebuild`
  - `install` depends on the `all` default target being built in order to install those same files
  - `test` additionally depends on the tests being built. As an internal implementation detail of meson... hmm, let's call it a bug, actually, since lots of people complain about it, and it's one I intend to fix in the future, the tests are registered to build by default, even when explicitly marked as `build_by_default: false`. Once this bug is fixed, building the project's default target (the one needed for installing) will *not* be sufficient to make the tests work.

martin-belanger:
@eli-schwartz - I agree with you. meson will indeed invoke "build", if needed, while doing an "install" operation. However, that's not necessarily a good thing. One must be `root` in order to `install`. However, one does not necessarily want to `build` as `root`. That's why we need to separate `install` from `build` and only apply `sudo` to the `install` operation and not the `build`.

That's what the following rule does. Note how `sudo` is only invoked for `meson install` and not the `build`.
```
install: ${NAME}
	cd ${BUILD-DIR} && sudo meson $@
```

igaw:
The `Makefile` only exists for those who have muscle memory during development. That is `make`, `make clean` ,  `make test` and maybe `make rpm`. For anything else use meson directly. 

Having a `sudo` in the `Makefile` is completely broken. So I am think we should just remove these targets.

martin-belanger:
> Having a `sudo` in the `Makefile` is completely broken. So I am think we should just remove these targets.

Duly noted. I was just trying to make it easier... of well.

igaw:
The thing I fear is that we end up trying to emulate all the autohell targets which will never work. The distro have everything in place for handling meson projects, so they wont use the Makefile anyway.

And for normal users, I rather point to meson's excellent documentation than spending time trying to solve autohell inducted problems ;)

eli-schwartz:
Indeed, sudo breaks things if you e.g. try to do `DESTDIR=$PWD/pkg make install`

Meson internally detects if you don't have permission to write to the installation files, and will reinvoke itself using polkit (not sudo, which I find sad), so it can be argued it's actually better to never run meson install as root!

(Of course this also has its own weird edge case. DESTDIR is mandatory in distro packaging recipes, except unusual cases like nixos. But if you forget it, you should get obvious permission denied errors. Running pkexec would simply hang forever, or else crash if it can't be run in a console-less container.)

...

Sudo also breaks things in any case that you don't pass --no-rebuild to `meson install` because ninja writes a logfile as root. So this PR actually makes things worse by running one instance of ninja as root.

...

tl;dr wrappers on top of wrappers on top of wrappers are complicated and have lots of moving parts. It can be very difficult for *anyone* other than the developers of those moving parts to have *any clue* what's going on. Fortunately, I'm a developer of the meson component of these moving parts. :)

PR #383, opened 2022-06-13
Fix/Add Bit Definitions NVMe Commands

jeffreyalien:
2 Fixes were made with this Pull Request

- Added Storage Tag Check IO Control Flag bit definition
- Fixed the LSP and LSI mask values used in the git log command.


igaw:
Thanks for the followup!

PR #386, opened 2022-06-17
MI: Update MI Admin API to suit core

jk-ozlabs:
This changeset unifies the first set of MI Admin commands to use the `_args`-struct-style API, and updates the example code in `mi-mctp` to suit, and clarify the usage of the `identify_partial` function.

This is dependent on https://github.com/linux-nvme/libnvme/pull/384 to expose those `_args` types.

As always, let me know if you have any queries/comments!


igaw:
Nice cleanup. The `_args`  structs usage here is a good improvement.

Issue #210, opened 2022-02-01
build warnings for __u64 in printf format string on PPC64


igaw:
There are few build warnings for test/register.c  and examples/discover-loop.c

https://build.opensuse.org/build/Base:System/openSUSE_Factory/ppc64/libnvme/_log

```
[   32s] ../test/register.c: In function ‘nvme_print_registers’:
[   32s] ../test/register.c:71:28: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘__u64’ {aka ‘long unsigned int’} [-Wformat=]
[   32s]    71 |         printf("%-10s : %llx\n", "CAP", cap);
[   32s]       |                         ~~~^            ~~~
[   32s]       |                            |            |
[   32s]       |                            |            __u64 {aka long unsigned int}
[   32s]       |                            long long unsigned int
[   32s]       |                         %lx
```

hreinecke:
I've cleaned up the function documentation with PR #214 . Maybe it's worthwhile including that into the package to reduce the number of errors/warning generating during package build.

igaw:
project has moved on build.opensuse.org: https://build.opensuse.org/build/Base:System/openSUSE_Factory/ppc64/libnvme/_log

igaw:
I thought this is easy:

```patch
-       printf("%-10s : %llx\n", "ASQ", asq);
-       printf("%-10s : %llx\n", "ACQ", acq);
+       printf("%-10s : %" PRIx64 "\n", "ASQ", asq);
+       printf("%-10s : %" PRIx64 "\n", "ACQ", acq);
```

though now I get:
```

In file included from ../test/register.c:15:
/usr/include/inttypes.h:121:41: note: format string is defined here
  121 | # define PRIx64         __PRI64_PREFIX "x"
../test/register.c:116:16: warning: format ‘%lx’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘__u64’ {aka ‘long long unsigned int’} [-Wformat=]
  116 |         printf("%-10s : %" PRIx64 "\n", "ASQ", asq);
      |                ^~~~~~~~~~~                     ~~~
      |                                                |
      |                                                __u64 {aka long long unsigned int}
In file included from ../test/register.c:15:
/usr/include/inttypes.h:121:41: note: format string is defined here
  121 | # define PRIx64         __PRI64_PREFIX "x"
[...]

PR #389, opened 2022-06-20
Fix %llx/%lx build warnings on powerpc

jk-ozlabs:
powerpc64 uses the `int-l64` type definitions, so we get an unsigned long
for our `__u64` type. This causes several build warnings when we're
printing with `%llx` in tests/register.c and example/discover-loop.c, as
it's an unsigned long, not an unsigned long long.

One way to fix this would be to use the PRIx64 definitions from
inttypes.h - however, those are keyed off the `__WORDSIZE` definition,
essentially:

    # if __WORDSIZE == 64
    #  define PRIx64	"lx"
    # else
    #  define PRIx64	"llx"
    # endif

\- and that breaks on x86_64, where our `__u64` is an unsigned long long,
and would need `%llx`.

The powerpc types header does give us an option to use the int-l64
definitions though:

    /*
     * This is here because we used to use l64 for 64bit powerpc
     * and we don't want to impact user mode with our change to ll64
     * in the kernel.
     *
     * However, some user programs are fine with this.  They can
     * flag __SANE_USERSPACE_TYPES__ to get int-ll64.h here.
     */
    #if !defined(__SANE_USERSPACE_TYPES__) && defined(__powerpc64__) && !defined(__KERNEL__)
    # include <asm-generic/int-l64.h>
    #else
    # include <asm-generic/int-ll64.h>
    #endif

... and in our case we are in fact fine with this. So, #define
`__SANE_USERSPACE_TYPES__` for the two example programs.

We refrain from doing this to the library headers though, as we have no
idea what libnvme users will need for `__u64`.

### DIFF
--- a/examples/discover-loop.c
+++ b/examples/discover-loop.c
@@ -12,6 +12,8 @@
  * system (no existing connection required). The output will look more
  * interesting with more targets.
  */
+#define __SANE_USERSPACE_TYPES__
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/test/register.c
+++ b/test/register.c
@@ -11,6 +11,8 @@
  * for your pci device found in /sys/class/nvme/nvmeX/device/resource0
  */
 
+#define __SANE_USERSPACE_TYPES__
+
 #include <fcntl.h>
 #include <inttypes.h>
 #include <libnvme.h>


jk-ozlabs:
CC @mpe for actual powerpc type sanity checking

igaw:
Thanks for digging into this. This should fix #210

mpe:
Yep looks good to me.

`__SANE_USERSPACE_TYPES__` began on powerpc but is now also defined on alpha and mips, two other very important architectures :)

It's also fairly widely used so it's not like you'd be the first to do this.

PR #390, opened 2022-06-20
doc: Add kernel-doc to header files

igaw:
Add missing documentation bits to all header files except types.h.
This a very rudimentary and surely needs a to be refined later. But
let's first try to get kernel-doc to report 0 warning/errors so we can
actually detect offenders added by new code.


jk-ozlabs:
Let me know if you need any work done on the MI header definitions.

igaw:
Thanks. I'll try to get it done today. Slowly, I get the feeling the types.h file should split and reordered. It's pretty wild...

PR #392, opened 2022-06-21
MI: add further admin commands

jk-ozlabs:
This series adds a few new admin commands to the set supported by MI:

 - Get Log Page
 - Security Send
 - Security Receive

As usual, these share the `_args` structs used for the core libnvme interface.

We also add a "raw" interface, as a basic interface for performing arbitrary commands.

For each of the added commands, extend the `mi-mctp` example application to show usage. In the case of the security commands, we use Security Protocol 0 to query the other protocols supported by the controller.

### DIFF
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -323,11 +323,104 @@ int do_get_log_page(nvme_mi_ep_t ep, int argc, char **argv)
 	return 0;
 }
 
+int do_admin_raw(nvme_mi_ep_t ep, int argc, char **argv)
+{
+	struct nvme_mi_admin_req_hdr req;
+	struct nvme_mi_admin_resp_hdr *resp;
+	struct nvme_mi_ctrl *ctrl;
+	size_t resp_data_len;
+	unsigned long tmp;
+	uint8_t buf[512];
+	uint16_t ctrl_id;
+	uint8_t opcode;
+	__le32 *cdw;
+	int i, rc;
+
+	if (argc < 2) {
+		fprintf(stderr, "no controller ID specified\n");
+		return -1;
+	}
+
+	if (argc < 3) {
+		fprintf(stderr, "no opcode specified\n");
+		return -1;
+	}
+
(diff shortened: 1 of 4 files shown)


igaw:
Looks good.

It's very cool that you add the tests along side. Now we just need to get it hooked up into Codecov Report :)

igaw:
Got it wrong, just 'examples'... Just in case you could transform these into tests, I would really appreciate it. We need to improve on the testing side.

jk-ozlabs:
We do have a couple of tests for the mi code, but it looks like they're not showing up in the coverage report. I assume that's because they're pulling the sources in directly:

```meson
mi_sources = [
    '../src/nvme/mi.c',
    '../src/nvme/log.c',
    '../src/nvme/cleanup.c',
]

mi = executable(
    'test-mi',
    ['mi.c'] + mi_sources,
    # ...
)
```

rather than linking against the shared library. Would this cause the coverage for mi.c to be missing?

We do use a couple of internal definitions in the tests, so it's not just a matter of dropping those `mi_sources`, unfortunately. Is this something we can fix in the coverage-collection process?

Once that's sorted, I can definitely expand on the mi-layer tests.

I also have a few tests queued for the MCTP layer too, will send a PR for that after some cleanup.

jk-ozlabs:
Hm, nope, something else is going on.

I see we have about 40% coverage in `src/nvme/mi.c`:

```shell
$ meson setup obj -Db_coverage=true --werror && ninja -C obj test coverage
[...]
$ grep nvme/mi.c obj/meson-logs/coverage.txt | cut -c -65
src/nvme/mi.c                                264     107    40%  
```

but the codecov reports are showing 0% for `mi.c`:

https://app.codecov.io/gh/linux-nvme/libnvme/compare/392/tree/src/nvme

Any ideas?

igaw:
Not yet. I haven't really looked into the Codecov setup. Maybe @glimchb might have an idea.

glimchb:
will take a look

PR #395, opened 2022-06-21
Use ubuntu-22.04 for newer gcov

MoshiBin:


glimchb:
@igaw please approve for CI

.github/workflows/meson.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           path: build/meson-logs/meson-log.txt
 
   code-coverage:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: install libraries
         run: sudo apt-get install libjson-c-dev lcov
```
glimchb, line 81: we can try using this in the BSFishy/meson-build https://github.com/marketplace/actions/meson-build#gcovr-version
```
gcovr-version: 12.1.0
```

.github/workflows/meson.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           path: build/meson-logs/meson-log.txt
 
   code-coverage:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: install libraries
```
glimchb, line 80: or we can remove `lcov` and use `gcovr` instead

igaw:
Resolved this via #397

PR #396, opened 2022-06-22
MI: Add endpoint iterators, and allow scanning from MCTP endpoints on dbus

jk-ozlabs:
The MCTP userspace infrastructure provides the set of discovered endpoints over dbus. This series adds a facility to query that list for endpoints that report support for the NVMe-MI protocol, and add them to the `nvme_root_t`.

With that information, we can add iterators to traverse the results of the scan.

This dbus mechanism is conditional on the low-level sdbus interface library, provided by libsystemd, and is enabled/disabled at `meson setup`.

As always, comments, queries, etc. are most welcome.

### DIFF
--- a/src/nvme/mi-mctp.c
+++ b/src/nvme/mi-mctp.c
@@ -21,6 +21,12 @@
 
 #include <ccan/endian/endian.h>
 
+#ifdef CONFIG_LIBSYSTEMD
+#include <systemd/sd-event.h>
+#include <systemd/sd-bus.h>
+#include <systemd/sd-id128.h>
+#endif
+
 #include "private.h"
 #include "log.h"
 #include "mi.h"
@@ -73,6 +79,10 @@ struct nvme_mi_transport_mctp {
 	__u8	eid;
 	int	sd;
 };
+
+#define MCTP_DBUS_PATH "/xyz/openbmc_project/mctp"
+#define MCTP_DBUS_IFACE "xyz.openbmc_project.MCTP"
+#define MCTP_DBUS_IFACE_ENDPOINT "xyz.openbmc_project.MCTP.Endpoint"
 
 static const struct nvme_mi_transport nvme_mi_transport_mctp;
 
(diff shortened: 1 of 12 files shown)


igaw (changes requested):
- add to all function/definitions in mi.h proper documentation. I just realized it really looks a bleak right now.
 - use consistently `/* */` for comments and not `//` for consistency reasons. 
 - I don't think we should mix the library code and a D-Bus service daemon. If I understood it correctly, the D-Bus parts are not depending on any internal APIs, so it could be placed into a separate file (even directory).

jk-ozlabs:
> * add to all function/definitions in mi.h proper documentation. I just realized it really looks a bit bleak right now.

Sure, can do! I'll add those as a preliminary patch.
 
> * use consistently `/* */` for comments and not `//` for consistency reasons.

No problem, will update.
 
> * I don't think we should mix the library code and a D-Bus service daemon. If I understood it correctly, the D-Bus parts are not depending on any internal APIs, so it could be placed into a separate file (even directory).

The dbus service daemon isn't included here, that's a separate component (part of the general MCTP tools, at https://github.com/CodeConstruct/mctp ). This addition requests the discovered NVMe-capable endpoints from that daemon. Is that still OK?

jk-ozlabs:
> misread it then

On re-reading my PR description, I totally see how it could be read that way - sorry for any ambiguity.

To clarify: `mctpd`, shipped as part of the general MCTP tools, performs MCTP endpoint discovery and enumeration - somewhat like DHCP, but also keeping track of the protocols advertised by each endpoint.

These patches add support for libnvme to query `mctpd` for endpoints that support the NVMe-over-MCTP protocol, and then use that to populate the endpoints list.

I'll update the series with your requested changes tomorrow, might be best to save any further review for then :)

igaw:
Thanks for explaining. I think this explanation could (should?) go into our documentation, e.g. in doc/mi.rst(.in).

jk-ozlabs:
Looks like I hit a difference in the init process from https://github.com/linux-nvme/libnvme/pull/397 ; fixing now.

igaw:
The tests are working! Hurray!

jk-ozlabs:
Any objections from exporting `nvme_mi_init_ep()` in the symbol map, but leaving out of the public headers?

This would allow our tests to use the "actual" mi_init, rather than re-implementing it (and then getting it wrong) within the test code.

igaw:
Ooops, sorry about this. Please go ahead and export it. Not sure why it's not in the header.

jk-ozlabs:
> Ooops, sorry about this. Please go ahead and export it. Not sure why it's not in the header.

It's intentionally not in the (public) header because it's only part of the core <--> transport interface, not the public interface.

We do have it in `src/private.h` so that it's available for transports to use, but they don't need the export in the .map, of course.

I'd still like to keep it in private.h, but it's a little unconventional to have it in the .map but not headers, I guess.

igaw:
Got it. Maybe move the symbol to an extra section in the map? So that we can later remove it if we want to?

jk-ozlabs:
ooh, that's a good idea, I'll check it out.

(the tests define their own fake transport, so the awkwardness here is that the test code is both providing a transport, as well as using the public API...)

jk-ozlabs:
OK, proposed in #400, but I'm happy to come up with a different approach if you prefer.

Once that's sorted I'll rebase this.

igaw:
The whole thing is a tradeoff. I think it's an acceptable compromise for the time being.

jk-ozlabs:
Rebased onto the test library changes, and re-tested on hardware.

src/nvme/mi-mctp.c
```diff
@@ -194,11 +194,28 @@ static void nvme_mi_mctp_close(struct nvme_mi_ep *ep)
...
+	struct nvme_mi_transport_mctp *mctp;
+	char *s;
+
+	if (ep->transport != &nvme_mi_transport_mctp)
+		return NULL;
+	mctp = ep->transport_data;
+
+	s = malloc(100);
+	if (s) {
+		snprintf(s, 100, "net %d eid %d", mctp->net, mctp->eid);
```
igaw, line 218: asnprintf() could be used here.

src/nvme/mi.c
```diff
@@ -616,3 +617,28 @@ void nvme_mi_close_ctrl(nvme_mi_ctrl_t ctrl)
...
+
+	len = 30;
+	len += strlen(ep->transport->name);
+	if (ts)
+		len += strlen(ts);
+
+	s = malloc(len);
+	if (s) {
+		if (ts)
+			snprintf(s, len, "%s: %s", ep->transport->name, ts);
```
igaw, line 648: the `strlen` and `malloc` dance looks a bit awkward. `asprintf` avoids this altogether

igaw (approved):
Looks good.  The documentation bits  about the D-Bus interaction are really helpful.

There is just the `strlen`-`malloc` dance which is a bit awkward and could really make use of `asprintf`

jk-ozlabs:
Sure - if you're OK with using `asprintf`, I'll update that.

I might even make the transport->desc API take a fixed-sized string, avoiding one of those allocations.

igaw:
There are already users of asprintf in nvme-cli and libnvme. This ship has sailed :)

jk-ozlabs:
now with less `malloc`!

Issue #393, opened 2022-06-21
Test Coverage Report is wrong


igaw:
I see we have about 40% coverage in `src/nvme/mi.c`:

```shell
$ meson setup obj -Db_coverage=true --werror && ninja -C obj test coverage
[...]
$ grep nvme/mi.c obj/meson-logs/coverage.txt | cut -c -65
src/nvme/mi.c                                264     107    40%  
```

but the codecov reports are showing 0% for `mi.c`:

https://app.codecov.io/gh/linux-nvme/libnvme/compare/392/tree/src/nvme

Any ideas?

_Originally posted by @jk-ozlabs in https://github.com/linux-nvme/libnvme/issues/392#issuecomment-1161334274_

igaw:
@glimchb

glimchb:
will take a look

jk-ozlabs:
It's entirely possible I'm doing something odd with the `mi` test; let me know if you need me to check anything out.

glimchb:
when using
```
git clean -ffdx  && meson setup obj -Db_coverage=true --werror && meson test -C obj && ninja -C obj coverage --verbose
```
locally, I also don't see coverage in local report with `lcov` here
```
lynx obj/meson-logs/coveragereport/src/nvme/mi.c.gcov.html
```
shows 0%

igaw:
I get

`src/nvme/mi.c                                262     107    40% `

but I had to install gcovr first.

glimchb:
could be related to versions running in azure's ubuntu... 
trying here https://github.com/linux-nvme/libnvme/pull/395

igaw:
I installed lcov and removed gcovr from system and it looks like mi.c is missing the report

glimchb:
@igaw please approve #395 let's see how things with new ubuntu

glimchb:
same @MoshiBin 
```
geninfo: WARNING: GCOV did not produce any data for /home/runner/work/libnvme/libnvme/build/test/9f86d08@@test-mi@exe/.._src_nvme_mi.c.gcda
```
https://github.com/linux-nvme/libnvme/runs/6987377286?check_suite_focus=true

glimchb:
we can try using this in the BSFishy/meson-build
```
gcovr-version: 12.1.0
```

igaw:
Ah, I see. I mixed gcov and gcovr up. It might be worth to try latest greates gcovr in this case.

Anyway, the no data case seems only to happen for the mi.c file.

[...]

PR #397, opened 2022-06-22
test: Remove code dependency for mi test

igaw:
When generating th coverage report, lcov chokes on the '../src/nvme'
paths prefix:

  geninfo: WARNING: GCOV did not produce any data for /home/wagi/work/libnvme-upstream/obj/src/test-mi.p/.._test_mi.c.gcda

We could restructure the project layout, e.g. by moving the test/mi.c
file to the src/nvme direcotry. But that looks ugly and as it turns
out we only really need one function copied over
nvme_mi_crc32_update() to resolve the dependency. The other function
is nvme_mi_init_ep() which we can easily open code.

Obviously, it would be better to fix the path issue for lcov but for
the time being let's go with this.

Fixes: #393

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -39,17 +39,9 @@ zns = executable(
     include_directories: [incdir, internal_incdir]
 )
 
-# The management interface tests don't require hardware, we have a small
-# test-mi endpoint instead.
-mi_sources = [
-    '../src/nvme/mi.c',
-    '../src/nvme/log.c',
-    '../src/nvme/cleanup.c',
-]
-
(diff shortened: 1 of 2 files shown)


igaw:
I've tried to limit the scope of the coverage report, e.g. I don't think we need to include ccan or tests.

`lcov` and `gcovr` can be configured via a config file but this is only available on meson => 0.63 which is currently under development: https://mesonbuild.com/Release-notes-for-0-63-0.html#

So let's life with the coverage report over the whole project.

PR #398, opened 2022-06-22
fabrics: Update controller authentication in nvmf_add_ctrl()

hreinecke:
When calling merge_config() in nvmf_add_ctrl() the controller
authentication key can't be updated reliably as it may already
been set. So update the controller with the correct key
manually.

Suggested-by: Sagi Grimberg <sagi@grimberg.me>

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -596,8 +596,17 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
 					nvme_ctrl_get_host_iface(c),
 					nvme_ctrl_get_trsvcid(c),
 					NULL);
-		if (fc)
+		if (fc) {
 			cfg = merge_config(c, nvme_ctrl_get_config(fc));
+			/*
+			 * An authentication key might already been set
+			 * in @cfg, so ensure to update @c with the correct
+			 * controller key.
+			 */
+			if (fc->dhchap_key)
+				nvme_ctrl_set_dhchap_key(c, fc->dhchap_key);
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -596,8 +596,17 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
...
-		if (fc)
+		if (fc) {
 			cfg = merge_config(c, nvme_ctrl_get_config(fc));
+			/*
+			 * An authentication key might already been set
+			 * in @cfg, so ensure to update @c with the correct
+			 * controller key.
+			 */
+			if (fc->dhchap_key)
+				nvme_ctrl_set_dhchap_key(c, fc->dhchap_key);
```
igaw, line 607: Can't this logic be in merge_config() directly?

hreinecke: As mentioned, not really.
How should merge_config() decide whether to overwrite an existing value?
We have several cases to contend with:
a) Passing in the dhchap key from the command-line for 'nvme connect' _and_ having a value in the config file:
   -> should be overwritten
b) Passing in the dhchap key from the command-line for 'nvme connect-all' _and_ having a value in the config file:
  -> should not be overwritten, as the command line key refers to the discovery controller, not the I/O controller
So moving that into merge_config() will be cumbersome.

igaw: Got it. I neglected the fact merge_config is called twice... Though I think the merge_config function starts to smell... :)

PR #399, opened 2022-06-23
MI: Add kernel-doc documentation for MI API

jk-ozlabs:
This change adds kernel-doc documentation for the new definitions in `mi.h`, and a file-level header for general API information.

We also fix a misnamed argument found while writing the docs, added as a small prerequisite patch.

Let me know if I've missed any conventions from the core libnvme docs.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -14,7 +14,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Check documentation format
         run: |
-          API_FILES="fabrics.h filters.h ioctl.h linux.h tree.h types.h"
+          API_FILES="fabrics.h filters.h ioctl.h linux.h mi.h tree.h types.h"
           for file in $API_FILES; do
             ./doc/kernel-doc -v -none src/nvme/$file 2>&1 | sed -r 's/^([^:]+):([0-9]+): (warning|error):/::\3 file=\1,line=\2::/g'
           done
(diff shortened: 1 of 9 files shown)


jk-ozlabs:
Just pushed an update, with a new `mi` doc base, as `doc/mi.rst.in`. However, I'm not 10% clear whether I need both the `.rst.in` and the `.rst`, so I've only inuded the former for now, as they just seem to be duplicates. Let me know if you need both though.

jk-ozlabs:
Just a bunch of amendments for typos; sorry for the noise, we should be good now!

jk-ozlabs:
> Not completely sure if we actually still need the `.rst` file

I've only added the `mi.rst.in` here, and everything seems to work OK. However, I have added the references to both `index.rst` and `index.rst.in` because I'm being cautious about what might be required :)

> Is this PR independent or is it ontop of our PR #396 ?

This is independent at the moment, so is based directly on the `master` branch. I have an update to #396 which will add the scan documentation to the new `mi.rst.in` (added by this change) as you've requested in that PR.

igaw:
Ok, let's merge this first in this case.

igaw:
It looks like these changes didn't make to readthedocs site:

  https://libnvme.readthedocs.io/en/latest/index.html

although it says the latest build was against this commit. Ah, this might be the reason 
why we have the `.rst` file also in the tree.

  https://readthedocs.org/api/v2/build/17255626.txt

jk-ozlabs:
Yep, that would seem to be the issue:

```
doc/index.rst:9: WARNING: toctree contains reference to nonexisting document 'mi'
```

I'll send an update.

igaw:
Maybe it is possible to trick readthedocs to run the `.rst.in` to `.rst` step. Then we could drop these files from the source tree.

PR #400, opened 2022-06-23
MI: reinstate nvme_mi_init_ep for endpoint initialisation in test

jk-ozlabs:
Ideally, we would be using the actual implementation of
nvme_mi_init_ep() for our tests, rather than open-coding it in the test
init.

This change exports the nvme_mi_init_ep from libnvme-mi.so, but it
remains excluded from the headers, as it's only intended for use in the
transport API. This means we can call it from the tests, but keep is
somewhat-internal.

We put this into a specific _TEST section of the version script, to keep
separate from the public symbols.

This prevents us from diverging the endpoint init process in our
testcases.

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -19,3 +19,11 @@ LIBNVME_MI_1_1 {
 	local:
 		*;
 };
+# API exported only for libnvme-mi internal test functions. These should
+# not be used other than through the in-tree tests, and cannot be considered
+# at all stable.
+LIBNVME_MI_TEST {
+	global:
+		nvme_mi_init_ep;
+		nvme_mi_crc32_update;
+};
--- a/test/mi.c
+++ b/test/mi.c
@@ -30,18 +30,6 @@ struct test_transport_data {
 
 static const int test_transport_magic = 0x74657374;
 
-static __u32 crc32_update(__u32 crc, void *data, size_t len)
-{
-	int i;
-
-	while (len--) {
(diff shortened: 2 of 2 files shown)


igaw:
Do you want to add the crc32 method as well?

igaw:
And please add a fat comment in the map file to warn anyone depending on this symbol!

jk-ozlabs:
OK, will add crc32 too, good plan. Means we'll be testing more of the actual code.

igaw:
Just got another idea: What about just have a separate library linked for testing only? 

```meson
libnvme_mi_test = library(
   [...]
   link_args: ['-Wl,--version-script=' + mi_version_script_arg],
   [...]
)
```
and only have the internal exposed in libnvme_mi_test (updating the mi_version_script_arg)

jk-ozlabs:
Yep, could do - we could even drop `--version-script` for the `libnvme-mi-test.so` build, making all symbols global, perhaps.

This is all conditional on the coverage getting detected across the two builds though, I guess. Any reliable way to test this?

igaw:
I think you run 

`rm -rf obj && meson setup obj -Db_coverage=true --werror && meson test -C obj && ninja -C obj coverage --verbose`

locally you should see the coverage report. I assume it will work as the source files match?

jk-ozlabs:
Yep, that seems to work well. Proposed in https://github.com/linux-nvme/libnvme/pull/403

igaw:
Cool thanks!

PR #402, opened 2022-06-23
doc: check-in processed mi.rst

jk-ozlabs:
It looks like the generated doc output on readthedocs.io requires the
processed .rst output to be present in the repo, so check it in.


igaw:
This did the trick. readthedocs now shows the new chapter: https://libnvme.readthedocs.io/en/latest/

PR #403, opened 2022-06-23
test: define separate library for MI unit tests

jk-ozlabs:
We previously exposed a couple of internal functions for libnvme-mi.so
in ce4a044720. However, Daniel Wagner has suggest instead using a
separate shared library with all symbols visible, just for the unit
test.

This change reverts the symbol visibility changes for the installed
libnvme-mi.so, and defines a separate, non-installed library for the MI
unit tests. This is the same as libnvme-mi.so, but has no linker script
or versioning information.

This means we're not exposing the init_ep and crc32 functions in the
actual .so, but can still allow unit tests to access non-public symbols.
The coverage reports can handle the separate library just fine, as the
coverage data is correlated on the underlying source files.

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -19,11 +19,3 @@ LIBNVME_MI_1_1 {
 	local:
 		*;
 };
-# API exported only for libnvme-mi internal test functions. These should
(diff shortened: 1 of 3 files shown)


igaw:
Yep, let's go with this and see what happens :)

PR #407, opened 2022-06-28
test: Add tests for mi-mctp layer

jk-ozlabs:
This series adds a set of initial tests for the mi-mctp code, by intercepting the socket calls and replacing them with a test-defined set of device interactions.

This means we can test the MI layer on systems with no `AF_MCTP` support too.

Comments welcome - particularly around the way we abstract the socket interface. This means that `socket`, `sendmsg` and `recvmsg` go through an indirect function call, to allow the tests to override.


jk-ozlabs:
Boo:

```
../test/utils.c:63:2: error: ignoring return value of ‘ftruncate’, declared with attribute warn_unused_result [-Werror=unused-result]
   63 |  ftruncate(fileno(logfd), 0);
```

(fix coming right up!)

jk-ozlabs:
> This means that `socket`, `sendmsg` and `recvmsg` go through an indirect function call, to allow the tests to override.

... and maybe we could use something similar for the `ioctl` interface too?

jk-ozlabs:
Just pushed a minor change: we no longer need the `-Wl,wrap=` link arguments - they were for an earlier version that linked directly with the `src/nvme/*.o` files, which we no longer do as of 0f6ee9b8b .

igaw:
I really like this. I was already discussion with @hreinecke about different strategy for testing. One thing we should do is to avoid unnecessary syscalls when possible. E.g. the whole scanning code just needs to walk the sysfs tree. Though we still would need to do a syscall. With your idea to mock the syscalls we could even test those code paths.

PR #410, opened 2022-07-01
MI: Strict response checking, add support for More Processing Required responses

jk-ozlabs:
This series adds some stricter checking on MI response data, and better handling for response error conditions.

We also add support for More Processing Required responses, which mean we need to perform another recvmsg to retrioeve the follow-up response.

As always: comments, queries etc are most welcome.


src/nvme/mi-mctp.c
```diff
@@ -213,15 +213,13 @@ static int nvme_mi_mctp_submit(struct nvme_mi_ep *ep,
 		 * somewhere in the header buf */
 		resp->hdr_len = len - sizeof(mic);
 		resp->data_len = 0;
-		memcpy(&mic, ((uint8_t *)resp->hdr) + resp->hdr_len,
-		       sizeof(mic));
+		mic = *(__le32 *)(((uint8_t *)resp->hdr) + resp->hdr_len);
```
igaw, line 345: Hmm, not particular a big fan of these casts and load strategy. E.g. can we sure that the address is aligned?

src/nvme/mi.h
```diff
@@ -106,6 +106,29 @@ enum nvme_mi_ror {
...
+	NVME_MI_RESP_MPR = 0x01,
+	NVME_MI_RESP_INTERNAL_ERR = 0x02,
+	NVME_MI_RESP_INVALID_OPCODE = 0x03,
+	NVME_MI_RESP_INVALID_PARAM = 0x04,
+	NVME_MI_RESP_INVALID_CMD_SIZE = 0x05,
+	NVME_MI_RESP_INVALID_INPUT_SIZE = 0x06,
+	NVME_MI_RESP_ACCESS_DENIED = 0x07,
+};
+
+
```
igaw, line 131: one new line too much

src/nvme/mi-mctp.c
```diff
@@ -84,10 +85,16 @@ struct nvme_mi_transport_mctp {
 	int	sd;
 };
 
+static int tag_ioctl(int sd, unsigned long req, struct mctp_ioc_tag_ctl *ctl)
```
igaw, line 88: Maybe `ioctl_tag` in case you want to add more `ioctls` then all of them have the same prefix?

igaw (commented):
Looks good. A couple of nitpicks :)

src/nvme/mi.c
```diff
@@ -157,6 +157,24 @@ int nvme_mi_submit(nvme_mi_ep_t ep, struct nvme_mi_req *req,
...
+
+	if (resp->hdr_len < sizeof(struct nvme_mi_msg_hdr))
+		return -EINVAL;
+
+	if (resp->hdr_len & 0x3)
+		return -EINVAL;
+
+	if (resp->data_len & 0x3)
+		return -EINVAL;
+
```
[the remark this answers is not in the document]
jk-ozlabs: First three are `req`, last three are `resp`.

src/nvme/mi-mctp.c
jk-ozlabs: Good question - the header itself *should* be aligned, and we know that the `hdr_len` is 4-byte aligned too, so we should get an aligned result from the add.

However, these loads are all added by a single (small) patch - I'm happy to drop that one if you'd prefer to keep the `memcpy`.

src/nvme/mi.h
jk-ozlabs: Ack, will remove.

src/nvme/mi-mctp.c
jk-ozlabs: I'm happy with either. We shouldn't have any new forms of `ioctl()` there, as that's all the kernel interface provides. If we do need to extend this later, it's an entirely-internal API, so we can just update all at once.

jk-ozlabs: On second thought, the `ioctl_tag` probably makes more sense as we're also prescribing the prototype of the ioctl call there. I'll update this in v2.

src/nvme/mi-mctp.c
igaw: Well, I am just one of those guys who get bitten very often by alignment issues on embedded systems. So any direct load is always kind of a danger zone. It's your call, I am fine if you want to use the direct load. Just be aware, I will not resist saying 'I told you so' if we ever happen to hit this problem :)

src/nvme/mi.c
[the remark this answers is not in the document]
igaw: Ah got it. Didn't see it.

src/nvme/mi-mctp.c
jk-ozlabs: I'll defer to your experience and drop that change :grin:

v2 coming shortly.

PR #413, opened 2022-07-02
types: fix typo on nvme_mi_ccs definitions in public API

jk-ozlabs:
The NVMe-MI spec defines the Composite Controller Status definitions
using the abbreviation CCS. However, the libnvme code (since initial
commit) has used CSS, not CCS (except for CCS_RDY...).

This change fixes the definitions to use the _CCS_ names, and adds a set
of #defines to allow the old names to be used, including for the enum
itself.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -5189,34 +5189,48 @@ struct nvme_mi_nvm_ss_health_status {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
There may be implications when changing these, particularly with the public API. I'm happy to rework this if necessary, or can just use the old names in my upcoming changes.

igaw:
I think those defines should take care with the backwards compatibility. At least I can't see any problems with it...

Issue #408, opened 2022-06-28
Can't build rpm package


lignumqt:
I'm trying to build an rpm package, but I'm getting an error, what am I doing wrong?
also I would like to know can i build deb package instead rpm?

I run: `make rpm`
> [0/1] Installing files.
Installing subprojects/json-c-0.15/libjson-c.so to /root/libnvme/.build/meson-private/dist-install/usr/lib/x86_64-linux-gnu
Installing subprojects/openssl-1.1.1l/libcrypto.so to /root/libnvme/.build/meson-private/dist-install/usr/lib/x86_64-linux-gnu
Installing subprojects/openssl-1.1.1l/libssl.so to /root/libnvme/.build/meson-private/dist-install/usr/lib/x86_64-linux-gnu
Installing src/libnvme.so.1.0.0 to /root/libnvme/.build/meson-private/dist-install/usr/lib/x86_64-linux-gnu
Installing symlink pointing to libnvme.so.1.0.0 to /root/libnvme/.build/meson-private/dist-install/usr/lib/x86_64-linux-gnu/libnvme.so.1
Installing symlink pointing to libnvme.so.1 to /root/libnvme/.build/meson-private/dist-install/usr/lib/x86_64-linux-gnu/libnvme.so
Installing /root/libnvme/.build/meson-private/dist-build/subprojects/json-c-0.15/json_config.h to /root/libnvme/.build/meson-private/dist-install/usr/include/json-c
Installing /root/libnvme/.build/meson-private/dist-unpack/libnvme-1.0/subprojects/json-c-0.15/arraylist.h to /root/libnvme/.build/meson-private/dist-install/usr/include/json-c
Installing /root/libnvme/.build/meson-private/dist-unpack/libnvme-1.0/subprojects/json-c-0.15/debug.h to /root/libnvme/.build/meson-private/dist-install/usr/include/json-c
Installing /root/libnvme/.build/meson-private/dist-unpack/libnvme-1.0/subprojects/json-c-0.15/json_c_version.h to /root/libnvme/.build/meson-private/dist-install/usr/include/json-c
Installing /root/libnvme/.build/meson-private/dist-unpack/libnvme-1.0/subprojects/json-c-0.15/json_inttypes.h to /root/libnvme/.build/meson-private/dist-install/usr/include/json-c
[...]

PR #414, opened 2022-07-04
rpmbuild: Enable 'make rpm' to build rpm pkgs #408

sc108-lee:
- Build will be happen in Makefile using meson
- Using meson build directory instead of Source(tar.gz)
- meson is needed higher version then what rpm repo offering (use pip install)
- Enable building static library (default: shared)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,5 +39,8 @@ test: ${BUILD-DIR}
 	ninja -C ${BUILD-DIR} $@
 
 .PHONY: rpm
-rpm: dist
-	rpmbuild -ba ${BUILD-DIR}/libnvme.spec
+rpm: ${BUILD-DIR}
+	git archive --format=tar HEAD > libnvme.tar
+	tar rf libnvme.tar ${BUILD-DIR}/libnvme.spec
+	gzip -f -9 libnvme.tar
+	rpmbuild -ta libnvme.tar.gz -v
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ a.out
 *.swp
 *.a
 *.so.*
-
+*.tar.gz
 cscope.*
 
 .build
--- a/libnvme.spec.in
+++ b/libnvme.spec.in
@@ -4,13 +4,9 @@ Release: 0
 Summary: Linux-native nvme device management library
 
 License: @LICENSE@
-Source: %{name}-%{version}.tar.gz
+Source: libnvme.tar.gz
 BuildRoot: %{_tmppath}/%{name}-root
 URL: http://github.com/linux-nvme/libnvme
-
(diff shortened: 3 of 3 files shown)


igaw:
This "Build will be happen in Makefile using meson" sounds wrong to me. Why is rpmbuild not able to build from within? The spec file *is* the build script.

sc108-lee:
@igaw since we are making spec file (version) from meson.
I assume that it's already built via meson once.
so I make that work with meson build directory instead of build twice.

igaw:
If the user type `make rpm` it should just work. Don't make the spec file more special than it needs to be.

sc108-lee:
I assume that this is the way you want @igaw :).
(it does meson .build twice~ 1. on git directory to get spec file, 2. rpmbuild dir to build & install)

Makefile
```diff
@@ -39,5 +39,8 @@ test: ${BUILD-DIR}
 	ninja -C ${BUILD-DIR} $@
 
 .PHONY: rpm
-rpm: dist
-	rpmbuild -ba ${BUILD-DIR}/libnvme.spec
+rpm: ${BUILD-DIR}
```
martin-belanger, line 42: I think the dependency here should be `${NAME}` instead of `${BUILD-DIR}`

igaw:
FYI, @jk-ozlabs, I am getting the same warnings since e33cc1782a01 ("doc/meson: provide a full-path version of api_files")

But this is triggering an ERROR.

```shell
$ meson .build --prefix=/tmp/libnvme -Ddocs=all -Ddocs-build=true 
[...]
doc/rst/meson.build:14:8: ERROR: File ../../src/nvme/../src/nvme/filters.h does not exist.
```

jk-ozlabs:
> FYI, @jk-ozlabs, I am getting the same warnings since [e33cc17](https://github.com/linux-nvme/libnvme/commit/e33cc1782a01cb96ecbbfdf913047b52ab429984) ("doc/meson: provide a full-path version of api_files")

OK, I'll check it out.

jk-ozlabs:
https://github.com/linux-nvme/libnvme/pull/427 should fix the warnings introduced by e33cc17, but I can't replicate the ERROR you're seeing there. Will investigate.

jk-ozlabs:
Ah, I see what's happening here. The fix you're proposing in 6093e5ad2 changes api_files to be actual paths, (`../src/nvme`), but this variable is also used in `doc/rst/meson.build`, so the paths are now invalid for the `rst` subdirectory.

I'd suggest just using #427 instead of 6093e5ad2 proposed here.

sc108-lee:
cool just drop the commit

igaw (approved):

PR #415, opened 2022-07-05
mi: Add MI configuration commands

jk-ozlabs:
This change adds an API to send MI Get / Set Configuration commands to
an endpoint, allowing control of endpoint SMBus frequenct, MCTP MTU and
clearing the subsystem health status bits.

We do this with a new couple of functions:

    int nvme_mi_mi_config_get(nvme_mi_ep_t ep, __u32 dw0, __u32 dw1,
                              __u32 *nmresp)

    int nvme_mi_mi_config_set(nvme_mi_ep_t ep, __u32 dw0, __u32 dw1)

The dw0, dw1 and nmresp formats depend on the type of configuration
accessed, which can be a little opaque. Se we add a bunch of helpers
too, to get/set the three configuration params:

    nvme_mi_mi_config_get_smbus_freq(...);
    nvme_mi_mi_config_set_smbus_freq(...);
    nvme_mi_mi_config_get_mctp_mtu(...);
    nvme_mi_mi_config_set_mctp_mtu(...);
    nvme_mi_mi_config_set_health_status_change(...);

[there's no getter for the latter, as this just clears polled bits]

Also, add support in examples/mi-mctp to illustrate some basic
configuration interactions, and some basic tests.

### DIFF
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -507,6 +507,119 @@ int do_security_info(nvme_mi_ep_t ep, int argc, char **argv)
 	return 0;
 }
 
+struct {
+	enum nvme_mi_config_smbus_freq id;
+	const char *str;
+} smbus_freqs[] = {
+	{ NVME_MI_CONFIG_SMBUS_FREQ_100kHz, "100k" },
+	{ NVME_MI_CONFIG_SMBUS_FREQ_400kHz, "400k" },
+	{ NVME_MI_CONFIG_SMBUS_FREQ_1MHz,   "1M" },
+};
+
+static const char *smbus_freq_str(enum nvme_mi_config_smbus_freq freq)
+{
+	unsigned int i;
+
+	for (i = 0; i < ARRAY_SIZE(smbus_freqs); i++) {
+		if (smbus_freqs[i].id == freq)
+			return smbus_freqs[i].str;
+	}
+
+	return NULL;
+}
+
+static int smbus_freq_val(const char *str, enum nvme_mi_config_smbus_freq *freq)
+{
+	unsigned int i;
+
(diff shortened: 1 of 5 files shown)


igaw (changes requested):
I think this should be three patches:
- introduction of new types (mi.h)
- new API functions
- example

Please also address the documentation issues reported by kernel-doc.

Another point which is independent to this patch is the error codes and return values: The libnvme API returns 0 on success on -1 on error with errno set to the error code. I think we should do the same with the mi.h API. Having two different styles is not really good IMO.

jk-ozlabs:
Sure, can do, will send a v2 today.

For the kernel-doc issues, how are you checking those? with a normal docs build, I get a *lot* of errors from existing headers, do you use anything to extract new issues?

jk-ozlabs:
> For the kernel-doc issues, how are you checking those?

oop, nevermind - found it in the CI.

jk-ozlabs:
> Another point which is independent to this patch is the error codes and return values: The libnvme API returns 0 on success on -1 on error with errno set to the error code. I think we should do the same with the mi.h API. Having two different styles is not really good IMO.

So this is a good point, and does highlight one difference in the MI interface: we may have failures that are due to the transport itself, rather than the command.

Because of that, the API so far has (as yet unwritten) semantics for the error return values:

- 0 for success
- negative return value (negative errno if possible) for transport-reported failures
- positive return value (always the MI status value) for MI-device reported failures

I'm happy to document that, but I'd like to first confirm whether that's suitable in general (I'd just propose a small tidy up: for the second case to also set `errno`). If not, we can move discussion to a bug for an alternative convention.

igaw (approved):

PR #416, opened 2022-07-06
doc: Add a kdoc test for API docs

jk-ozlabs:
Even when we're not building docs, it would be handy to have an easy
check that the kernel-doc comments do not produce warnings or errors.

This change adds a test() defintion to docs/, which calls a simple
script, kernel-doc-check. This script invokes kernel-doc, looking for
error and/or warning output.

This requires a bump in the minimum meson version, for the
program.full_path() method.

RFC: This is RFC-only at present. If this is acceptable, I'll submit as
a series that unifies the api_paths array with the actual doc build
process.


igaw:
Sure, looks reasonable but I am very hesitant to update the minimum requirements for meson blindly. This usually triggers a lot of complains. We should at least try to figure out what the consequences are.

jk-ozlabs:
Sure, I can look at doing this without needing `full_path`, which will save the meson update.

Will rework when I have some spare cycles here.

igaw:
Release dates of meson:
- 0.45.0: Mar 04, 2018
- 0.55.0: Jul 12, 2020

I haven't found any definitive web site listening the package versions for the main distros. Only found this one here

- https://repology.org/project/meson/versions

As far I can tell a bunch of LTS distros will be shipping a too old meson version. Usually via backports it should be possible to upgrade to a newer version of meson. It could be that  relying on a two year old release is asking for troubles. 

Then subprojects such as openssl are relying on 0.55.0. Not sure what to do...

jk-ozlabs:
I have a change that removes the need for passing the kernel-doc binary to the check script entirely - this means we can keep the meson version requirements as-is.

Will push those updates soon.

Otherwise, any issue with doing those `find_program()` calls outside of the `want_docs` conditional?

jk-ozlabs:
Yep, I reckon we can do this without too needing to bump meson; new revision attached, and no longer a draft.

The controversy now will be whether we keep the github workflow (patch 3). Let me know if you have any preferences.

igaw:
The workflow integration is not so great anyway as the results are a bit hidden and are hard to find. Also the build doesn't fail which is what we actually want. No complaints from me.

PR #420, opened 2022-07-08
fabrics: Add No CDC Connectivity to EFLAGS

martin-belanger:
TP8010 adds a new "No Connectivity to CDC (NCC)" field to the Discovery Log Page entry flags (EFLAGS). This pull request adds this new definition to `types.h`. We also add the corresponding strings to `nvmf_eflags_str()`.

At the same time, the Python bindings are extended to report the EFLAGS when a discovery command is invoked. This was previously missing. Also, the libnvme python package has gained two new attributes:
1. \_\_version__
2. \_\_git_version__

Note: It is customary for Python packages to provide a \_\_version__ attribute. The \_\_git_version__ is just a bonus!

### DIFF
--- a/libnvme/meson.build
+++ b/libnvme/meson.build
@@ -46,7 +46,7 @@ if have_python_support
     configure_file(
         input:  '__init__.py', 
         output: '__init__.py',
-        copy: true,
+        configuration: conf,
         install_dir: python3.get_install_dir(pure: false, subdir: 'libnvme'),
     )
 
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -269,6 +269,8 @@ static void PyDict_SetItemStringDecRef(PyObject *p, const char *key, PyObject *v
     PyDict_SetItemStringDecRef(entry, "cntlid", val);
     val = PyLong_FromLong(e->asqsz);
     PyDict_SetItemStringDecRef(entry, "asqsz", val);
+    val = PyLong_FromLong(e->eflags);
+    PyDict_SetItemStringDecRef(entry, "eflags", val);
     PyList_SetItem(obj, i, entry); /* steals ref. to entry */
   }
   $result = obj;
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -4635,15 +4635,24 @@ enum nvme_subsys_type {
  *				 this flag set.
  * @NVMF_DISC_EFLAGS_EPCSD:	 Explicit Persistent Connection Support for Discovery (EPCSD):
  *				 Indicates that Explicit Persistent Connections are
- *				 supported for the Discovery controller.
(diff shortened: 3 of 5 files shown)


src/nvme/types.h
```diff
@@ -4635,15 +4635,24 @@ enum nvme_subsys_type {
...
+ * @NVMF_DISC_EFLAGS_NCC:	 No CDC Connectivity (NCC): If set to
+ *      			 '1', then no DDC that describes this entry
+ *      			 is currently connected to the CDC. If
+ *      			 cleared to '0', then at least one DDC that
+ *      			 describes this entry is currently
+ *      			 connected to the CDC. If the Discovery
+ *      			 controller returning this log page is not
+ *      			 a CDC, then this bit shall be cleared to
+ *      			 '0' and should be ignored by the host.
  * @NVMF_DISC_EFLAGS_BOTH:	 Indicates that both the DUPRETINFO and EPCSD
```
igaw, line 4648: The both documentation needs to removed. And for backwards compatibility add a define

```c
/* Backwards compatibility. Will be removed with next major release */
#define NVMF_DISC_EFLAGS_BOTH (NVMF_DISC_EFLAGS_DUPRETINFO | NVMF_DISC_EFLAGS_EPCSD)
```

libnvme/nvme.i
```diff
@@ -269,6 +269,8 @@ static void PyDict_SetItemStringDecRef(PyObject *p, const char *key, PyObject *v
     PyDict_SetItemStringDecRef(entry, "cntlid", val);
     val = PyLong_FromLong(e->asqsz);
     PyDict_SetItemStringDecRef(entry, "asqsz", val);
+    val = PyLong_FromLong(e->eflags);
+    PyDict_SetItemStringDecRef(entry, "eflags", val);
```
igaw, line 273: Could you move this change into a separate patch? It has nothing to do with the type change.

libnvme/meson.build
```diff
@@ -46,7 +46,7 @@ if have_python_support
     configure_file(
         input:  '__init__.py', 
         output: '__init__.py',
-        copy: true,
+        configuration: conf,
```
igaw, line 49: This should be in a separate patch. And even so, this would update the minimum requirement of meson to 0.49.0. I am *very* hesitant to do this. This needs to be properly documented why this is really necessary.

igaw (changes requested):

libnvme/meson.build
eli-schwartz: Why does it bump the minimum requirement?

igaw: I read the documentation incorrect. The support for dictionaries is depending on 0.49.0 but `conf` is a `configuration_data` object.

Issue #412, opened 2022-07-01
Persistent Event Log - Invalid Field Error


tkenbo:
Hello, I'm trying to understand if this is an issue or intended.  Hope anybody could enlighten me.

`[root@xxxxxxxx ~]#nvme persistent-event-log /dev/nvme1 -a 0`
`NVMe status: INVALID_FIELD: A reserved coded value or an unsupported value in a defined field(0x2002)`

I traced the command and here is the trace log for the failed command.
` <...>-552510 [031] .... 172168.299315: nvme_setup_cmd: nvme1: qid=0, cmdid=2, nsid=4294967295, flags=0x0, meta=0x0, cmd=(nvme_admin_get_log_page cdw10=0d 00 ff ff ff ff 00 00 00 90 3e 00 00 00 00 00 00 00 00 00 00 00 00 00)`
`<idle>-0     [061] d.h. 172168.299485: nvme_sq: nvme1: qid=0, head=23, tail=23`
`<idle>-0     [061] ..s. 172168.299492: nvme_complete_rq: nvme1: qid=0, cmdid=2, res=0x0, retries=0, flags=0x2, status=0x2002`

looking at the cdw10 (0xffff000d) and cdw11 (0x0000ffff), NUMD is set to 0xFFFFFFFF.  Per our engineer, this value exceeds MDTS and thus it returns the invalid field error.

Looking at the libnvme code, I can see how numd is calculated.
`int nvme_get_log(struct nvme_get_log_args *args)`
`    _u32 numd = (args->len >> 2) - 1;`

In my case, args->len is set to 2, which causes `numd = 0 - 1 = 0xFFFFFFFF`
if len is anything below 3, it makes numd to be 0xFFFFFFFF.

`int nvme_get_log(int fd, __u32 nsid, __u8 log_id, bool rae, __u8 lsp, __u32 data_len, void *data)`
`{`
`	__u32 offset = 0, xfer_len = data_len;`
`	do {`
`		xfer_len = data_len - offset;`
`		ret = nvme_get_log13(fd, nsid, log_id, lsp, offset, 0, rae, xfer_len, ptr);`

Here 
data_len = 0x3E9002
offset = 0x3E9000

Thanks in advance,

igaw:
Good catch, the expression `(args->len >> 2) - 1` is not correct for all values of `len`. 

From the spec:
```
Number of Dwords (NUMD): This field specifies the number of dwords of the Reservation Status
[...]

PR #425, opened 2022-07-12
ioctl: Fix numd calculation for corner case

igaw:
numd is a 0's based dword value. Thus after dividing the lenght (in
bytes) through 4 we subtract 1. This works for all values in the range
of [3..].

By adding 4 before the division we make sure we always floor on the
next integer. With this, numd get calculated correctly for all len
values in the range [0..].

Fixes: #412

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -395,7 +395,7 @@ int nvme_identify(struct nvme_identify_args *args)
 
 int nvme_get_log(struct nvme_get_log_args *args)
 {
-	__u32 numd = (args->len >> 2) - 1;
+	__u32 numd = ((args->len + 4) >> 2) - 1;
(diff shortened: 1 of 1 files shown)


igaw:
Zero based ranges are usually `[0..n)` and not `[0..n]`. So we should only do a plus 3 but that makes `len = 0` case fail.

igaw:
- `(i + 3) >> 2) - 1`
```
0 -> -1
1 -> 0
2 -> 0
3 -> 0
4 -> 0
5 -> 1
6 -> 1
7 -> 1
8 -> 1
9 -> 2
10 -> 2
11 -> 2
12 -> 2
13 -> 3
14 -> 3
15 -> 3
```

igaw:
Well after thinking a bit more on it, the range [0..3) can't be mapped to a 0's based index as we can't express no data. So, there is nothing we can do here.

PR #427, opened 2022-07-14
doc: Use correct variable in configure_file input

jk-ozlabs:
In e33cc17 we removed an inner loop on the result of files(), but didn't
update the local variable to suit.

This change uses the correct iterator variable for the configure_file()
invocation.

Reported-by: Daniel Wagner <dwagner@suse.de>


igaw:
Ah thanks. I got interrupted by lunch time. Taking your fix then!

PR #428, opened 2022-07-14
Fix doc build

igaw:
Fix two regressions in the doc building process.

See #414


igaw:
Dropping the second fix, as #427 is merged.

PR #433, opened 2022-07-18
Add support for uploading release candidates to PyPI

### LINKED ISSUE #429: pypi workflow should not upload release candidates
The current workflow rule

```yaml
  upload_pypi:
    needs: [build_sdist]
    runs-on: ubuntu-latest
    if: startsWith(github.ref, 'refs/tags/v')
```

uploads release candidates and marks them as normal release. E.g. v1.1-rc0 was uploaded as v1.1. 

@glimchb

igaw:
This changes slightly how releases are done. It's possible now to have the `-rc[0-9]+` suffix added to the project version. 

Fixes #429

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -46,13 +46,21 @@ jobs:
     runs-on: ubuntu-latest
     if: startsWith(github.ref, 'refs/tags/v')
     steps:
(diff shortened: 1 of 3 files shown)


igaw:
Note: this should be merged after the next release. The v1.1 release candidate is already done and we should not change our strategy during this phase

PR #434, opened 2022-07-18
Fix build against LibreSSL

ismaell:
Check the 3.x API for completeness, in case LibreSSL > 3.5.3 implements
the missing bits.

Signed-off-by: Ismael Luceno <ismael@iodev.co.uk>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -68,14 +68,16 @@ openssl_dep = dependency('openssl',
                          required: get_option('openssl'),
                          fallback : ['openssl', 'libssl_dep'])
 if openssl_dep.found()
+  api_v3 = openssl_dep.version().version_compare('>=3.0.0') and not (
+    # LibreSSL v3.x with incomplete OpenSSL v3 APIs
+    cc.has_header_symbol('openssl/opensslv.h',
+                         'LIBRESSL_VERSION_NUMBER',
+                         dependencies: openssl_dep) and
+    not cc.has_header('openssl/core_names.h', dependencies: openssl_dep))
+  api_v1 = not api_v3 and openssl_dep.version().version_compare('>=1.1.0')
   conf.set('CONFIG_OPENSSL', true, description: 'Is OpenSSL available?')
-  conf.set('CONFIG_OPENSSL_1',
-           openssl_dep.version().version_compare('<2.0.0') and
-           openssl_dep.version().version_compare('>=1.1.0'),
-           description: 'OpenSSL version 1.x')
-  conf.set('CONFIG_OPENSSL_3',
-           openssl_dep.version().version_compare('>=3.0.0'),
-           description: 'OpenSSL version 3.x')
+  conf.set('CONFIG_OPENSSL_1', api_v1, description: 'OpenSSL version 1.x')
+  conf.set('CONFIG_OPENSSL_3', api_v3, description: 'OpenSSL version 3.x')
 endif
 
 # Check for libsystemd availability. Optional, only required for MCTP dbus scan


meson.build
```diff
@@ -69,12 +69,17 @@ openssl_dep = dependency('openssl',
                          fallback : ['openssl', 'libssl_dep'])
 if openssl_dep.found()
   conf.set('CONFIG_OPENSSL', true, description: 'Is OpenSSL available?')
+  # XXX Check the 3.x API for completeness; needed for LibreSSL
+  openssl_3_complete = (openssl_dep.version().version_compare('>=3.0.0') and
```
igaw, line 73: Please don't add `XXX` to the comment? No idea what it suppose to mean. 

So this is a bit backwards in my opinion. LibreSSL is trying to be like OpenSSL, hence I prefer your idea with 

`openssl_is_libressl = cc.has_header_symbol('openssl/opensslv.h', 'LIBRESSL_VERSION_NUMBER')`

This explicitly tells us what is happening. We could rename CONFIG_OPENSSL_1 to something like `CONFIG_OPENSSL_API_V1` to make it even more explicit.

igaw (changes requested):

igaw:
I've played with this here but unfortunaly, this doesn't work as expected the 'has_header' seems to find my local installed OpenSSL header file `/usr/include/openssl` and hence the test will pass.

```diff
diff --git a/meson.build b/meson.build
index 472172d4de3f..5e8c7509a825 100644
--- a/meson.build
+++ b/meson.build
@@ -69,13 +69,25 @@ openssl_dep = dependency('openssl',
                          required: get_option('openssl'),
                          fallback : ['openssl', 'libssl_dep'])
 if openssl_dep.found()
+  api_v1 = openssl_dep.version().version_compare('<2.0.0') and
+           openssl_dep.version().version_compare('>=1.1.0')
+  api_v3 = openssl_dep.version().version_compare('>=3.0.0')
+
+  openssl_is_libressl = cc.has_header_symbol('openssl/opensslv.h',
+                                             'LIBRESSL_VERSION_NUMBER',
+                                             dependencies: openssl_dep)
+  if openssl_is_libressl and api_v3
+    if not cc.has_header('openssl/core_names.h', dependencies: openssl_dep)
+      # LibreSSL v3.x without the OpenSSL v3 APIs
+      api_v1 = true
+      api_v3 = false
+    endif
+  endif
+
   conf.set('CONFIG_OPENSSL', true, description: 'Is OpenSSL available?')
-  conf.set('CONFIG_OPENSSL_1',
-           openssl_dep.version().version_compare('<2.0.0') and
-           openssl_dep.version().version_compare('>=1.1.0'),
+  conf.set('CONFIG_OPENSSL_1', api_v1,
            description: 'OpenSSL version 1.x')
-  conf.set('CONFIG_OPENSSL_3',
-           openssl_dep.version().version_compare('>=3.0.0'),
+  conf.set('CONFIG_OPENSSL_3', api_v3,
            description: 'OpenSSL version 3.x')
 endif
```

```
Header <openssl/opensslv.h> has symbol "LIBRESSL_VERSION_NUMBER" with dependency openssl: YES 
Has header "openssl/core_names.h" with dependency openssl: YES 
```

```
Command line:  cc -I/tmp/libressl/include /home/wagi/work/libnvme/.build/meson-private/tmp1e_w2i2l/testfile.c -E -P -D_FILE_OFFSET_BITS=64 -P -O0 -std=gnu99 

Code:
 
        #ifdef __has_include
         #if !__has_include("openssl/core_names.h")
          #error "Header 'openssl/core_names.h' could not be found"
         #endif
        #else
         #include <openssl/core_names.h>
        #endif
Compiler stdout:
 
Compiler stderr:
 
Has header "openssl/core_names.h" with dependency openssl: YES 
```

igaw:
Ah, with '-nostdinc' the test works:

```meson
  openssl_is_libressl = cc.has_header_symbol('openssl/opensslv.h',
                                             'LIBRESSL_VERSION_NUMBER',
                                             dependencies: openssl_dep,
                                             args: '-nostdinc')
  if openssl_is_libressl and api_v3
    if not cc.has_header('openssl/core_names.h',
                          dependencies: openssl_dep,
                          args: '-nostdinc')
```

igaw:
for completes, this works:

```diff
diff --git a/meson.build b/meson.build
index 472172d4de3f..96c50379170f 100644
--- a/meson.build
+++ b/meson.build
@@ -69,13 +69,28 @@ openssl_dep = dependency('openssl',
                          required: get_option('openssl'),
                          fallback : ['openssl', 'libssl_dep'])
 if openssl_dep.found()
+  api_v1 = openssl_dep.version().version_compare('<2.0.0') and
+           openssl_dep.version().version_compare('>=1.1.0')
+  api_v3 = openssl_dep.version().version_compare('>=3.0.0')
+
+  openssl_is_libressl = cc.has_header_symbol('openssl/opensslv.h',
+                                             'LIBRESSL_VERSION_NUMBER',
+                                             dependencies: openssl_dep,
+                                             args: '-nostdinc')
+  if openssl_is_libressl and api_v3
+    if not cc.has_header('openssl/core_names.h',
+                          dependencies: openssl_dep,
+                          args: '-nostdinc')
+      # LibreSSL v3.x without the OpenSSL v3 APIs
+      api_v1 = true
+      api_v3 = false
+    endif
+  endif
+
   conf.set('CONFIG_OPENSSL', true, description: 'Is OpenSSL available?')
-  conf.set('CONFIG_OPENSSL_1',
-           openssl_dep.version().version_compare('<2.0.0') and
-           openssl_dep.version().version_compare('>=1.1.0'),
+  conf.set('CONFIG_OPENSSL_1', api_v1,
            description: 'OpenSSL version 1.x')
-  conf.set('CONFIG_OPENSSL_3',
-           openssl_dep.version().version_compare('>=3.0.0'),
+  conf.set('CONFIG_OPENSSL_3', api_v3,
            description: 'OpenSSL version 3.x')
 endif
```

ismaell:
On 19/Jul/2022 01:47, Daniel Wagner wrote:
<...>
> Please don't add `XXX` to the comment? No idea what it suppose to
> mean. 

Convention of BSD origin; I suspect the idea of using XXX (and
certainly why I use it) is that it's likely to show only in comments,
much like TODO and FIXME.

> `openssl_is_libressl = cc.has_header_symbol('openssl/opensslv.h', 'LIBRESSL_VERSION_NUMBER')`
> 
> This explicitly tells us what is happening. We could rename
> CONFIG_OPENSSL_1 to something like `CONFIG_OPENSSL_API_V1` to make
> it even more explicit.

The problem with this is that we can't predict which version of
LibreSSL will add the missing API, by testing explicitly at least it
becomes obvious when this changes...

ismaell:
On 19/Jul/2022 02:30, Daniel Wagner wrote:
> for completes, this works:
<...> 

+1 but pkg-config seems to remove -I/usr/include, so -nostdinc alone
doesn't work :(. I'll try to handle this tonight.

eli-schwartz:
What's the advantage of supporting libressl at all? The general trend in Linux distros has been to drop support for libressl as the system SSL library in any distro that used to use it.

e.g. https://lwn.net/Articles/841664/

I know of 3 distros that used to support it: Gentoo, Alpine, and Void. All switched back to openssl.

Python also dropped support for "an SSL module with some missing features and broken tests": https://peps.python.org/pep-0644/

igaw:
Yes, I am fully aware of this argument. That's why I would like to see LibreSSL workarounds really clearly separated from OpenSSL code.  

Anyway, what is the general mood on this topic? 

I am sitting on the fence. If the support for LibreSSL is really just a couple meson.build changes I don't mind. Obviously, if we have to start changing source code I would object.

ismaell:
On 19/Jul/2022 13:26, Eli Schwartz wrote:
> What's the advantage of supporting libressl at all? The general
> trend in Linux distros has been to drop support for libressl as
> the system SSL library in any distro that used to use it.
> 
> e.g. https://lwn.net/Articles/841664/
> 
> I know of 3 distros that used to support it: Gentoo, Alpine, and
> Void. All switched back to openssl.
> 
> Python also dropped support for "an SSL module with some missing
> features and broken tests": https://peps.python.org/pep-0644/

Source Mage GNU/Linux still supports LibreSSL, and will continue to
support it in the foreseeable future; at this point mainly because of
OpenSSL's CLA.

Alpine has an open discussion on supporting LibreSSL again:
https://gitlab.alpinelinux.org/alpine/tsc/-/issues/28

meson.build
```diff
@@ -68,14 +68,15 @@ openssl_dep = dependency('openssl',
...
+    not cc.has_header('openssl/core_names.h', dependencies: openssl_dep))
   conf.set('CONFIG_OPENSSL', true, description: 'Is OpenSSL available?')
-  conf.set('CONFIG_OPENSSL_1',
-           openssl_dep.version().version_compare('<2.0.0') and
-           openssl_dep.version().version_compare('>=1.1.0'),
-           description: 'OpenSSL version 1.x')
-  conf.set('CONFIG_OPENSSL_3',
-           openssl_dep.version().version_compare('>=3.0.0'),
-           description: 'OpenSSL version 3.x')
+  conf.set('CONFIG_OPENSSL_1', not api_v3, description: 'OpenSSL version 1.x')
```
igaw, line 78: This drops the condition:

           openssl_dep.version().version_compare('<2.0.0') and
           openssl_dep.version().version_compare('>=1.1.0'),

Especially the lower bound should stay.

igaw:
I am not really a big fan of your solution because it is difficult to follow. I am going to merge my version. Hope this okay.

ismaell:
> I am not really a big fan of your solution because it is difficult to follow. I am going to merge my version. Hope this okay.

See the comment on your PR; your solution is broken on libressl-native systems.

ismaell:
Also, you removed the lower version bound... wasn't that the one you wanted to keep?

igaw:
> Also, you removed the lower version bound... wasn't that the one you wanted to keep?

I realized that we already have the lower bound tests a few lines above.

PR #439, opened 2022-07-22
build: Add support to build against LibreSSL

igaw:
LibreSSL ships with the same API but has a different version number
scheme. That is LibreSSL 3.x is not providing the OpenSSL 3.x
APIs. Hence we need to test if the OpenSSL 3 APIs are available if
LibreSSL is present.

Reported-by: Ismael Luceno <ismael@linux.com>

Fixes: #434

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -68,14 +68,32 @@ openssl_dep = dependency('openssl',
                          required: get_option('openssl'),
                          fallback : ['openssl', 'libssl_dep'])
 if openssl_dep.found()
-  conf.set('CONFIG_OPENSSL', true, description: 'Is OpenSSL available?')
-  conf.set('CONFIG_OPENSSL_1',
-           openssl_dep.version().version_compare('<2.0.0') and
-           openssl_dep.version().version_compare('>=1.1.0'),
-           description: 'OpenSSL version 1.x')
-  conf.set('CONFIG_OPENSSL_3',
-           openssl_dep.version().version_compare('>=3.0.0'),
-           description: 'OpenSSL version 3.x')
+  conf.set('CONFIG_OPENSSL', true,
+           description: 'Is OpenSSL/LibreSSL available?')
+
+  if openssl_dep.version().version_compare('<2.0.0')
+    api_version = 1
+  endif
+
+  if openssl_dep.version().version_compare('>=3.0.0')
+    api_version = 3
+
+    # Test for LibreSSL v3.x with incomplete OpenSSL v3 APIs
+    is_libressl = cc.has_header_symbol('openssl/opensslv.h',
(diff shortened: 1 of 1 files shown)


ismaell:
`-nostdinc` makes the build fail for a system with native libressl.

igaw:
And how so? The pkg-config step should add the right path, no? Without `-nostdinc` it breaks on my build system where I also have OpenSSL in the standard paths.

igaw:
Right, I forgot that pkg-config is not adding the default paths. So on my system I installed libressl under /tmp/libressl:

```bash
$ export PKG_CONFIG_PATH=/tmp/libressl/lib64/pkgconfig 
$ pkg-config --cflags openssl
-I/tmp/libressl/include 
```
without `-nostdinc`

```
Header <openssl/opensslv.h> has symbol "LIBRESSL_VERSION_NUMBER" with dependency openssl: YES 
Has header "openssl/core_names.h" with dependency openssl: YES 
```

So the first test finds LibreSSL `openssl/opensslv.h` but `has_header` is finding the OpenSSL header file.

```
Header <openssl/opensslv.h> has symbol "LIBRESSL_VERSION_NUMBER" with dependency openssl: YES 
Running compile:
Working directory:  /home/wagi/work/libnvme/.build/meson-private/tmp9t2xnyda
Command line:  cc -I/tmp/libressl/include /home/wagi/work/libnvme/.build/meson-private/tmp9t2xnyda/testfile.c -E -P -D_FILE_OFFSET_BITS=64 -P -O0 -std=gnu99 

Code:
 
        #ifdef __has_include
         #if !__has_include("openssl/core_names.h")
          #error "Header 'openssl/core_names.h' could not be found"
         #endif
        #else
         #include <openssl/core_names.h>
        #endif
Compiler stdout:
 
Compiler stderr:
 
Has header "openssl/core_names.h" with dependency openssl: YES 
```

igaw:
I suppose we can drop the `-nostdinc` as it shouldn't break any OpenSSL systems.

eli-schwartz:
Ah yeah, because header includes are the union of all include paths. The header does exist in at least one place, and just because its patent directory is shadowed by another include path, doesn't mean that individual header is.

Maybe that test could be gated on the openssl API version as well as whether the header exists?

igaw:
Usually only one library is present and only if you have both libraries in the include path things will break. So my test is a bit artificial. Let's just drop the `-nostdinc` and if someone really wants to get this setup working, I am happy to review patches.

ismaell:
The problem with using `-nostdinc` is that `pkg-config` drops system directories unless `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` is defined.

PR #445, opened 2022-07-25
libnvme: Change nlbs type to __u32 in copy command

Francis-Pravin:
In copy command, nlbs is declared as __u16. While parsing command line argument,
it is typecast to integer pointer. So nlbs[1], nlbs[3], nlbs[5], ... are stored
as zero after getting parsed. Therefore it is passing wrong value to controller.
Hence, changing the nlbs type to __u32.

Signed-off-by: Francis Pravin Antony Michael Raj <francis.michael@solidigm.com>
Signed-off-by: Jonathan Derrick <jonathan.derrick@solidigm.com>

### DIFF
--- a/doc/rst/util.rst
+++ b/doc/rst/util.rst
@@ -168,7 +168,7 @@ The nvme command status if a response was received or -errno
 otherwise.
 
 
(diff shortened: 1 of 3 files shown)


igaw:
In this case the parser in nvme-cli needs to fixed. First, we can't change the API and second according the spec 16 bit value is correct.

PR #446, opened 2022-07-26
fabrics: Add nvmf_connect_disc_entry2

igaw:
The current nvmf_connect_disc_entry function does not set the DHCHAP
ctrl key. The problem is that we can't really pass a new argument into
the function without breaking the existing API.

The first idea (and probably the technical correct one) was to extend
struct nvme_fabrics_config. Unfortunately, we forgot to add the
args_size member to this data structure which makes it really hard to
extend. Introducing a wrapper struct and adding new function for all
users of struct nvme_fabrics_config (nvmf_default_config,
nvmf_update_config, nvmf_add_ctrl, ...) is a bit excessive.

The main problem we want to solve here is the ability to call setters
on a newly created controller before we do a connect call. The only
place where we have this sequence is nvmf_connect_disc_entry.

Hence, this approach introduces a new nvmf_connect_disc_entry2
function which sets the key between the ctrl creation and the connect
call. We use the struct args approach so that we can extend if
needed.

See https://github.com/linux-nvme/nvme-cli/issues/1618

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -650,10 +650,26 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
 	return nvme_init_ctrl(h, c, ret);
 }
 
+
 nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 				    struct nvmf_disc_log_entry *e,
 				    const struct nvme_fabrics_config *cfg,
 				    bool *discover)
+{
+	struct nvmf_connect_disc_entry2_args args = {
+		.args_size = sizeof(args),
+		.defcfg = cfg,
+		.discover = discover,
+		.dhchap_ctrl_key = NULL,
+	};
+
+	return nvmf_connect_disc_entry2(h, e, &args);
+}
+
+
+nvme_ctrl_t nvmf_connect_disc_entry2(nvme_host_t h,
+				     struct nvmf_disc_log_entry *e,
+				     struct nvmf_connect_disc_entry2_args *args)
 {
 	const char *transport;
 	char *traddr = NULL, *trsvcid = NULL;
@@ -710,7 +726,8 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 		 "(transport: %s, traddr: %s, trsvcid %s)\n",
 		 transport, traddr, trsvcid);
 	c = nvme_create_ctrl(h->r, e->subnqn, transport, traddr,
-			     cfg->host_traddr, cfg->host_iface, trsvcid);
+			     args->defcfg->host_traddr,
+			     args->defcfg->host_iface, trsvcid);
 	if (!c) {
 		nvme_msg(h->r, LOG_DEBUG, "skipping discovery entry, "
 			 "failed to allocate %s controller with traddr %s\n",
@@ -724,8 +741,8 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 		nvme_ctrl_set_discovered(c, true);
 		break;
 	case NVME_NQN_DISC:
-		if (discover)
-			*discover = true;
+		if (args->discover)
+			*args->discover = true;
 		nvme_ctrl_set_discovery_ctrl(c, true);
 		break;
 	default:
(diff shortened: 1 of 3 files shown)


hreinecke:
Are you really sure we need to?
When we're doing a 'connect-all' we should be using a _different_ controller structure from the main config tree.
And that controller structure should have the correct dhchap key already attached.
Why do we need to add the key to the nvmf_fabrics_config structure?

src/nvme/fabrics.c
```diff
@@ -650,10 +650,26 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
 	return nvme_init_ctrl(h, c, ret);
 }
 
+
```
hreinecke, line 653: Pointless newline.

src/nvme/fabrics.c
```diff
@@ -650,10 +650,26 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
...
 nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 				    struct nvmf_disc_log_entry *e,
 				    const struct nvme_fabrics_config *cfg,
 				    bool *discover)
+{
+	struct nvmf_connect_disc_entry2_args args = {
+		.args_size = sizeof(args),
+		.defcfg = cfg,
```
hreinecke, line 661: I'm not happy with the pointer approach; that way we will always have two allocations when using nvmf_connect_disc_entry_2.
I'd rather go for either embed struct nvme_fabrics_config into nvmf_connect_disk_entry2 (and copy data over), or simply have two structures, passing both of them to nvmf_connect_disc_entry2()

src/nvme/fabrics.h
```diff
@@ -237,8 +237,36 @@ char *nvmf_hostid_from_file();
  * Return: Pointer to the new controller
  */
 nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
-	struct nvmf_disc_log_entry *e,
-	const struct nvme_fabrics_config *defcfg, bool *discover);
+				    struct nvmf_disc_log_entry *e,
+			            const struct nvme_fabrics_config *defcfg,
+				    bool *discover);
+
```
hreinecke, line 243: The indentation here look pretty weird...

src/nvme/fabrics.h
```diff
@@ -237,8 +237,36 @@ char *nvmf_hostid_from_file();
...
+/**
+ * struct nvmf_connect_disc_entry2_args - Arguments for the nvmf_connect_disc_entry2 command
+ * @defcfg:	Default configuration to be used for the new controller
+ * @discover:	Set to 'true' if the new controller is a discovery controller
+ * @dhchap_ctrl_key: DH-HMAC-CHAP ctrl key
+ * @args_size:	Size of &struct nvmf_connect_disc_entry2_args
+ */
+struct nvmf_connect_disc_entry2_args
+{
+	const struct nvme_fabrics_config *defcfg;
```
hreinecke, line 253: As mentioned; I'd rather embed 'struct nvme_fabrics_config' here. Or use two different structures altogether.

hreinecke (changes requested):

igaw:
> Are you really sure we need to?

No, not really. I haven't been able to test this yet. It's purely based on the input from @martin-gpy

> When we're doing a 'connect-all' we should be using a _different_ controller structure from the main config tree.

My understanding is that  we don't create a such _different_ controller instance in the connect-all without JSON config file case. The lookup in nvmf_connect_disc_entry will not return a controller. 

> And that controller structure should have the correct dhchap key already attached.

But only if the lookup has a result. Hmm, obviously we could also try to figure out if we can fix this. IIRC we don't expose this internal things to the API yet.

> Why do we need to add the key to the nvmf_fabrics_config structure?

I though it is the logical place to keep all the transport related configuration grouped.

src/nvme/fabrics.c
igaw: We can't avoid the allocation for `struct nvme_fabrics_config` as it already part of the current API. I'm fine with either idea. I have a slight preference for `nvme_fabrics_config` and `nvmf_connect_disk_entry2_args` as separate arguments.

Though first let's figure it out if we need it at all.

igaw:
As discussed, this doesn't really make sense, because we can't tell which controller key to use in a multi controller setup. So let's just drop this one.

PR #449, opened 2022-07-28
fabrics: Fix build_options() return values

tbzatek:
A small issue spotted by reading the code.


igaw:
Should we also set the errno value to EINVAL in case the big 'if' fails?

igaw:
Forget it, I see that add_bool_argument() well set the errno if needed.

Issue #426, opened 2022-07-12
Config JSON file fails to apply dhchap_key at subsystem/port level


martin-gpy:
Turns out specifying dhchap_key in the config JSON file works only if specified at the host level in the config JSON file, and not at the corresponding port level. 

For e.g. this works:

```
# cat /home/config-success.json
[
  {
    "hostnqn":"nqn.2014-08.org.nvmexpress:uuid:6d6b4ca0-8bed-4bdb-a28e-07ea8ed0acf7",
    "hostid":"17503367-6ae0-4b0b-9f9b-113daf169556",
    "dhchap_key":"DHHC-1:00:QaxYTIQLMUZOE+\/dmIForZfP5Mf9OX0Hiy8+uhG8kdOjrVxN:",
    "subsystems":[
      {
        "nqn":"nqn.1992-08.com.netapp:sn.94d3e4c5f3ed11eca5871c666dbe2b2a:subsystem.ss_tcpnvme_2_1",
        "ports":[
          {
            "transport":"tcp",
            "traddr":"10.228.36.89",
            "host_traddr":"10.228.32.190",
            "trsvcid":"4420"
          }
        ]
      }
    ]
  }
]
```

But this fails:

```
# cat /home/config-fail.json
[
  {
    "hostnqn":"nqn.2014-08.org.nvmexpress:uuid:6d6b4ca0-8bed-4bdb-a28e-07ea8ed0acf7",
    "hostid":"17503367-6ae0-4b0b-9f9b-113daf169556",
    "subsystems":[
      {
        "nqn":"nqn.1992-08.com.netapp:sn.94d3e4c5f3ed11eca5871c666dbe2b2a:subsystem.ss_tcpnvme_2_1",
        "ports":[
          {
            "transport":"tcp",
            "traddr":"10.228.36.89",
            "host_traddr":"10.228.32.190",
            "trsvcid":"4420",
            "dhchap_key":"DHHC-1:00:QaxYTIQLMUZOE+\/dmIForZfP5Mf9OX0Hiy8+uhG8kdOjrVxN:"
          }
        ]
      }
    ]
  }
]
```

martin-gpy:
Turns out the dhchap key fails to apply at the subsystem level too. For e.g, this too fails:

```
# cat /etc/nvme/config-subsystem-fail.json
[
  {
[...]

PR #451, opened 2022-07-29
Implement DH-HMAC-CHAP key templating

hreinecke:
Implement templating for dhchap host keys, where the host key can be stored at either the host, subsystem, or controller level. The function nvme_ctrl_get_dhchap_host_key() will retrieve the key, with the setting from the controller taking precedence over both the subsystem and controller settings.


igaw:
Fixes #426

hreinecke:
Superseded by #461.

hreinecke:
Closing.

PR #455, opened 2022-08-12
libnvme: Fix the Reference Tag in copy command

Francis-Pravin:
The elbt is declared as elbt[10] to store 80-bit refernce tag value. For now,
the Kernel supports 32-bit and 48-bit refernce tag only. Therefore the value is
received in the 64-bit variable. So elbt[2] to elbt[9] is used to store that 64-bit value.

Identify the type of copy format using args->format instead of structure size.
Also, use a single 64-bit ilbrt instead of using two ilbrt with different datatypes.

Signed-off-by: Francis Pravin Antony Michael Raj <francis.michael@solidigm.com>
Signed-off-by: Jonathan Derrick <jonathan.derrick@solidigm.com>

### DIFF
--- a/src/nvme/api-types.h
+++ b/src/nvme/api-types.h
@@ -666,8 +666,6 @@ struct nvme_dsm_args {
  * @prinfow:	Protection information field for write
  * @dtype:	Directive type
  * @format:	Descriptor format
- * @ilbrt_u64:	Initial logical block reference tag - 8 byte
- *              version required for enhanced protection info
  */
 struct nvme_copy_args {
 	__u64 sdlba;
@@ -677,7 +675,7 @@ struct nvme_copy_args {
 	int fd;
 	__u32 timeout;
 	__u32 nsid;
-	__u32 ilbrt;
+	__u64 ilbrt;
 	int lr;
 	int fua;
 	__u16 nr;
@@ -688,7 +686,6 @@ struct nvme_copy_args {
 	__u8 prinfow;
 	__u8 dtype;
 	__u8 format;
-	__u64 ilbrt_u64;
 };
 
 /**
--- a/src/nvme/util.c
+++ b/src/nvme/util.c
@@ -401,14 +401,16 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
 			  __u64 *slbas, __u64 *eilbrts, __u32 *elbatms,
 			  __u32 *elbats, __u16 nr)
 {
-	int i;
+	int i, j;
 
 	for (i = 0; i < nr; i++) {
 		copy[i].nlb = cpu_to_le16(nlbs[i]);
 		copy[i].slba = cpu_to_le64(slbas[i]);
-		copy[i].elbt[2] = cpu_to_le64(eilbrts[i]);
 		copy[i].elbatm = cpu_to_le16(elbatms[i]);
 		copy[i].elbat = cpu_to_le16(elbats[i]);
+
+		for (j = 0; j < 8; j++)
+			copy[i].elbt[9 - j] = (eilbrts[i] >> (8 * j)) & 0xff;
 	}  
 }
 
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -1718,11 +1718,9 @@ int nvme_dsm(struct nvme_dsm_args *args)
 
 int nvme_copy(struct nvme_copy_args *args)
 {
-	const size_t size_v1 = sizeof_args(struct nvme_copy_args, format, __u64);
-	const size_t size_v2 = sizeof_args(struct nvme_copy_args, ilbrt_u64, __u64);
 	__u32 cdw3, cdw12, cdw14, data_len;
 
-	if (args->args_size < size_v1 || args->args_size > size_v2) {
+	if (args->args_size < sizeof(*args)) {
 		errno = EINVAL;
 		return -1;
 	}
@@ -1732,18 +1730,18 @@ int nvme_copy(struct nvme_copy_args *args)
 		((args->prinfow & 0xf) << 26) | ((args->fua & 0x1) << 30) |
 		((args->lr & 0x1) << 31);
 
-	if (args->args_size == size_v1) {
+	if (args->format == 0) {
 		cdw3 = 0;
 		cdw14 = args->ilbrt;
-	} else {
-		cdw3 = (args->ilbrt_u64 >> 32) & 0xffffffff;
-		cdw14 = args->ilbrt_u64 & 0xffffffff;
-	}
-
-	if (args->format == 1)
-		data_len = args->nr * sizeof(struct nvme_copy_range_f1);
-	else
 		data_len = args->nr * sizeof(struct nvme_copy_range);
+	} else if (args->format == 1) {
+		cdw3 = (args->ilbrt >> 32) & 0xffffffff;
+		cdw14 = args->ilbrt & 0xffffffff;
+		data_len = args->nr * sizeof(struct nvme_copy_range_f1);
+	} else {
+		errno = EINVAL;
+		return -1;
+	}
 
 	struct nvme_passthru_cmd cmd = {
 		.opcode         = nvme_cmd_copy,


src/nvme/api-types.h
```diff
@@ -677,7 +675,7 @@ struct nvme_copy_args {
 	int fd;
 	__u32 timeout;
 	__u32 nsid;
-	__u32 ilbrt;
+	__u64 ilbrt;
```
igaw, line 678: This breaks the API

src/nvme/api-types.h
```diff
@@ -688,7 +686,6 @@ struct nvme_copy_args {
 	__u8 prinfow;
 	__u8 dtype;
 	__u8 format;
-	__u64 ilbrt_u64;
```
igaw, line 691: and this one too.

igaw (changes requested):
You need to find another way to address this. We can fix this properly when we do the next major version of the library until then we have to find ways around without breaking the API.

src/nvme/api-types.h
Francis-Pravin: I have one more patch for nvme-cli to fix this issue. In that patch, this renamed ilbrt variable is handled. I hope the API will not break once it is merged.

src/nvme/util.c
```diff
@@ -401,14 +401,16 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
...
 
 	for (i = 0; i < nr; i++) {
 		copy[i].nlb = cpu_to_le16(nlbs[i]);
 		copy[i].slba = cpu_to_le64(slbas[i]);
-		copy[i].elbt[2] = cpu_to_le64(eilbrts[i]);
 		copy[i].elbatm = cpu_to_le16(elbatms[i]);
 		copy[i].elbat = cpu_to_le16(elbats[i]);
+
+		for (j = 0; j < 8; j++)
+			copy[i].elbt[9 - j] = (eilbrts[i] >> (8 * j)) & 0xff;
```
bjpaupor, line 413: Isn't this just a more complicated way of storing the eilbrt little-endian order in the last 8 bytes of elbt (as is removed above)?

Francis-Pravin: I have verified the copy command with QEMU emulated NVMe device. On using previous logic, the controller received the eilbrt value as zero via format1. Hence, I stored the value byte by byte. Here observed the controller is receiving the appropriate value.

src/nvme/api-types.h
igaw: As far I can tell, libnvme is not just used by nvme-cli. This is why I keep repeating we can't break the API for the 1.x series of the library. For 2.x sure, I don't have any problems to modify the API. 

BTW, I would suggest to open an in issue with the exact work description and queue it up for the 2.x library. I fear if we write it down, this will be lost.

src/nvme/util.c
igaw: Which version of the nvme-cli and libnvme did you use? This might be caused by my brown bag release where I
crippled the nvme_init_copy_range_f1 function. The result was, calling nvme_init_copy_range_f1 was a no-op.

https://github.com/linux-nvme/nvme-cli/pull/1636/commits/fd582b00ec7fc4b46094dee693dca9ea73024981

Francis-Pravin: Always I use the latest version of nvme-cli and libnvme. The issue was observed even before the above mentioned changes.

igaw: That would have been too easy...

igaw:
If I understand the situation the main issue is that:

`              copy[i].elbt[2] = cpu_to_le64(eilbrts[i]);`

is not working correctly. I suggest you just fix this using the existing API as we can't change it (only extend it).

jderrick:
Hi Daniel,

`copy[i].elbt[2] = cpu_to_le64(eilbrts[i]);`

Was the intent here to fill [2] - [9] ?
I have a question about this:
1. Shouldn't this start at [0]
2. Wouldn't this be non-little endian?

The same thing appears to be taking place in qemu:
[https://elixir.bootlin.com/qemu/latest/source/hw/nvme/ctrl.c#L2658](https://elixir.bootlin.com/qemu/latest/source/hw/nvme/ctrl.c#L2658)

igaw:
I haven't really digged into the details about this feature until now. I think we have two things to address here. 

First, is the handling of the API backwards compatibility, meaning how read the `struct nvme_copy_args`. The current version looks okay but it is awkwardly used. That means we have functional dependency on the version of struct. I think this is what you also tried to resolve and I said no.

```
        if (args->args_size == size_v1) {                                                                     
                cdw3 = 0;                                                                                     
                cdw14 = args->ilbrt;                                                                          
        } else {                                                                                              
                cdw3 = (args->ilbrt_u64 >> 32) & 0xffffffff;                                                  
                cdw14 = args->ilbrt_u64 & 0xffffffff;                                                         
        }         
```

Second we have to handle various sizes of LBST/ELBST/ILBRT/EIBTR.

```
Command Set Specification
Figure 35: Copy – Source Range Entries Descriptor Format 1h

elbt:

This field specifies variable sized Expected Logical Block Storage Tag (ELBST) and
Expected Initial Logical Block Reference Tag (EILBRT) fields, which are defined in section
5.2.1.4.1, to be used for the read portion of the copy operation. If the namespace is not
formatted to use end-to-end protection information, then this field is ignored.
```

As I understand the current version of `nvme_copy()` is assuming a specific STS value. To address this I think we need to extend `struct nvme_copy_args` with a STS field like `struct nvme_io_args` and update `nvme_copy()` and `nvme_init_copy_range_f1()`. 

Comments?

bjpaupor:
@igaw To the second point, there's a few problems with the current implementation and with the proposed fix here. 

- The current implementation only actually sets elbt[2] to eilbrt & 0xFF rather than elbt[2-9] to the full value. This could be fixed by using a 64-bit pointer or by setting it byte-wise like this proposed fix.
- The proposed fix here nearly works as it does set the correct bytes, but the problem is that it's equivalent to properly setting elbt[2] = cpu_to_**be**64(elibrt), as a big-endian value.
- Even adjusting the current implementation to use a 64-bit pointer gives a problem that it will start the value from elbt[2-X] rather than elbt[X-9].

Only way I see to properly set this value is by breaking the API to get the Storage Tag Size and using something like:
`elbt << STS = cpu_to_le64(eilbrt)`

Francis-Pravin:
Thanks for the comments.
I totally agree it. To handle the various sizes of LBST/ELBST/ILBRT/EIBTR, STS and PIF values are required. Because, the size of the reference tag are based on those values. Hence We need to extend the struct nvme_copy_args like in struct nvme_io_args. I think it will breaks the API. As @igaw  commented above we can't break the API for the 1.x series of the library.
Already  an issue has been created in libnvme to fix it in 2.0 version. #459 
Please add your suggestion on closing this PR.

igaw:
I think there is a misunderstanding. I said, we can't break the API but this doesn't mean we cannot extend it. This is where the versioning of the `struct args` comes into play. If a `struct arg` is passed into a function we check the size of it and only access added members if it is possible, hence the size check. So in this case the STS member needs to be added after the last current member (plus pending if needed) and then we can use it in the function.

See also https://github.com/linux-nvme/libnvme/commit/9350e73e302a2dfaa3be9b964e6a9681c6740ff5

bjpaupor:
That should work for the nvme_copy function and set tags for the write portion of the copy command, but the nvme_init_copy_range_f1 function doesn't have a 'struct args' to extend for the read portion tags. 

For that we'd need to update the signature to include PIF and STS, and maybe convert the list of parameters to a struct that could be versioned and extended in the future.

igaw:
Got it. In this case we have to introduce a `nvme_init_copy_range_f1` v2 (which we are going to merge when we do the 2.x release of the library).

igaw:
Not sure about adding `struct args` to the util function though. We had this discussion when we introduce the `struct args` in the `api-types.h` header if we should do the same for the helpers and decided against it. Can't remember the argument though.

src/nvme/api-types.h
Francis-Pravin: Hi @igaw 
I missed to mention a point.
In `struct nvme_copy_args`, `ilbrt` and `ilbrt_u64` are used to store the reference tag values of 32-bit and 64-bit respectively. The `ilbrt_u64` is only assigned with `cfg.ilbrt` value in nvme-cli. Whereas `ilbrt` is not assigned with any value. But, both the variables are handled in `nvme_copy()` libnvme. It seems fishy for me. That's why I tried to use single `ilbrt` instead of two variable with different data type.

Francis-Pravin:
I need some clarification on identifying the copy format type based on the `sizeof_args` value. 
In nvme-cli, `args_size` is assigned with `sizeof(args)` which means `args_size` will be assigned with whole structure size value. But, in libnvme the format type is identified based on the `size_v1` and `size_v2`. Based on my understanding, the control will always go to `else` condition only. Could you please throw light on this issue.

igaw:
A nvme-cli binary build against v1 will only pass in `arg_size = sizeof(v1)`. If you happen to have a newer library installed which has support for v2 it will not attempt to read anything but v1 members. The `arg_size` is only relevant for binaries.

igaw:
Closing due to inactivity.

PR #457, opened 2022-08-14
MI: expand admin command implementation

jk-ozlabs:
This series adds MI implementations for more Admin commands:

 * Get Features
 * Set Features
 * Namespace Management
 * Namespace Attach
 * Format NVM
 * Sanitize NVM

plus a few inline helpers for these and the existing Get Log Page and Identify commands (which `nvme-cli` needs to support the above)

As always, comments/queries/etc most welcome.


igaw:
Thanks!

One thing I start to wonder if we could harness the tests you are adding also for the libnvme part.... just a random thought

PR #458, opened 2022-08-16
example/mi-conf: Add an example for MI transport configuration

jk-ozlabs:
With the Read MI Data Structure and MI Configuration Set functions
available, we can query a device for its maximum MTU, and update the
current MTU to suit.

This also requires telling the local MCTP stack about the new MTU too,
by sending a d-bus message to `mctpd`.

This change adds a little example for this functionality, where we
perform the MI query and configuration, and implements the d-bus
communication for the local stack control.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -39,3 +39,12 @@ executable(
     dependencies: libnvme_mi_dep,
     include_directories: [incdir, internal_incdir]
 )
+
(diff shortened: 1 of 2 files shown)


jk-ozlabs:
Looks like the optional `libsystemd` dependency isn't being handled correctly, will fix.

PR #460, opened 2022-08-23
mi: Allow Admin-message sized More Processing Required responses

jk-ozlabs:
Devices may implement their MPR response as an actual Admin response
message, rather than the simple MI-only message described in 4.1.2.3 of
NVMe-MI v1.2b.

Allow this, but with some fairly stringent header checks. Add a test for
this behaviour too.

### DIFF
--- a/src/nvme/mi-mctp.c
+++ b/src/nvme/mi-mctp.c
@@ -179,13 +179,15 @@ struct nvme_mi_msg_resp_mpr {
  * populate the worst-case expected processing time, given in milliseconds.
  */
 static bool nvme_mi_mctp_resp_is_mpr(struct nvme_mi_resp *resp, size_t len,
-				     unsigned int *mpr_time)
+				     __le32 mic, unsigned int *mpr_time)
 {
+	struct nvme_mi_admin_resp_hdr *admin_msg;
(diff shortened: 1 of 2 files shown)


src/nvme/mi-mctp.c
```diff
@@ -179,36 +179,58 @@ struct nvme_mi_msg_resp_mpr {
...
+	 * If the packet was at the expected response size, then mic will
+	 * be set already; if not, find it within the header/data buffers.
+	 */
+
+	/* Devices may send a MPR response as a full-sized Admin response,
+	 * rather than the minimal MI-only header. Allow this, but only if the
+	 * type indicates admin, and the allocated response header is the
+	 * correct size for an Admin response.
+	 */
+	if (((msg->hdr.nmp >> 3) & 0xf) == NVME_MI_MT_ADMIN &&
```
igaw, line 212: Should we have a  getter function for this? We have partially stuff in place but not everywhere (see also #https://github.com/linux-nvme/libnvme/issues/148)

igaw: Okay got it. Let's go then with this version and keep that in mind.

PR #461, opened 2022-08-25
Allow to get and set DHCHAP host key on controller level

hreinecke:
The json config schema already defines a DHCHAP host key on the controller level, so add the infrastructure for parsing and setting it. And while we're at it fixup the json config schema to define the DHCHAP key on the host level.

Fixes: #426

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -287,7 +287,9 @@ struct nvme_host {
   char *hostnqn;
   char *hostid;
   char *hostsymname;
-  char *dhchap_key;
+  %extend {
+    char *dhchap_key;
+  }
 };
 
 struct nvme_subsystem {
@@ -332,7 +334,9 @@ struct nvme_ctrl {
   char *subsysnqn;
   char *traddr;
   char *trsvcid;
-  char *dhchap_key;
+  %extend {
+    char *dhchap_key;
+  }
   char *cntrltype;
   char *dctype;
   bool discovery_ctrl;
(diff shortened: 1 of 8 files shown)


src/libnvme.map
```diff
@@ -22,6 +22,7 @@ LIBNVME_1_0 {
 		nvme_ctrl_get_address;
 		nvme_ctrl_get_config;
 		nvme_ctrl_get_dhchap_key;
+		nvme_ctrl_get_dhchap_host_key;
```
igaw, line 25: These need to go into LIBNVME_1_2

src/libnvme.map
```diff
@@ -46,6 +47,7 @@ LIBNVME_1_0 {
 		nvme_ctrl_next_path;
 		nvme_ctrl_reset;
 		nvme_ctrl_set_dhchap_key;
+		nvme_ctrl_set_dhchap_host_key;
```
igaw, line 50: And this one as well

hreinecke: Fixed with the new submission.

libnvme/nvme.i
```diff
@@ -335,6 +336,7 @@ struct nvme_ctrl {
   char *traddr;
   char *trsvcid;
   %extend {
+    char *dhchap_host_key:
     char *dhchap_key;
```
igaw, line 340: Wouldn't make sense to rename `dhcap_key` to `dhcp_ctrl_key` as in the C code part as well?

src/nvme/private.h
```diff
@@ -82,6 +82,7 @@ struct nvme_ctrl {
 	char *subsysnqn;
 	char *traddr;
 	char *trsvcid;
+	char *dhchap_key;
```
igaw, line 85: `dhchap_host_key`?

igaw:
Looks good. Just my nitpicking on the naming pattern which seems a bit inconsisten.

libnvme/nvme.i
hreinecke: Compat stuff. When we're renaming it in the python binding all existing python programs will need to be updated. With this they continue working.

src/nvme/private.h
hreinecke: We could, but with this naming we're consistent with the kernel usage. But in the end, I don't mind.

libnvme/nvme.i
igaw: Okay, so it's another todo on the 2.x list.

Issue #466, opened 2022-09-07
C++ Compiler Error on libnvme-mi


drakedog2008:
The C++ compiler doesn't fully compatible with C99 standard. So it gave the errors: 

1.  
```
/usr/include/nvme/mi.h:790:56: error: invalid conversion from 'int' to 'nvme_mi_config_smbus_freq' [-fpermissive]
```
2. 
```
/usr/include/nvme/mi.h:2227:9: error: designator order for field 'nvme_ns_mgmt_args::nsid' does not match declaration order in 'nvme_ns_mgmt_args'
 2227 |         };
```

jk-ozlabs:
@drakedog2008 thanks for the report, I'll take a look.

jk-ozlabs:
Looks like @drakedog has a PR already: https://github.com/linux-nvme/libnvme/pull/468

igaw:
The fixes are in the tree. Closing bug report.
8fd81cdb88bb ("mi: Fix C++ compiler errors")

PR #468, opened 2022-09-08
Fix C++ compiler errors

drakedog2008:
C++20 compiler reported the following errors:

```
error: invalid conversion from 'int' to 'nvme_mi_config_smbus_freq'
```

```
error: designator order for field 'nvme_ns_mgmt_args::nsid' does not match declaration order in 'nvme_ns_mgmt_args'
```

### DIFF
--- a/src/nvme/mi.h
+++ b/src/nvme/mi.h
@@ -787,7 +787,7 @@ static inline int nvme_mi_mi_config_get_smbus_freq(nvme_mi_ep_t ep, __u8 port,
 
 	rc = nvme_mi_mi_config_get(ep, dw0, 0, &tmp);
 	if (!rc)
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
This should fix https://github.com/linux-nvme/libnvme/issues/466 , but appears to break the actual build at the moment.

igaw:
The cast needs a 'enum' I think.

drakedog2008:
Didn't play with C for a while. Apparently it is the `enum` issue. Thanks for the advise. Updated the PR.

jk-ozlabs:
LGTM!

igaw:
I've merged the change directly because I updated the commit message slightly. Thanks for the contribution!

Issue #467, opened 2022-09-08
Compile warning in mi-mctp.c 


igaw:
```shell
$ gcc --version
gcc (SUSE Linux) 12.2.1 20220830 [revision e927d1cf141f221c5a32574bde0913307e140984]
```

```shell
[35/100] Compiling C object subprojects/libnvme/examples/mi-mctp.p/mi-mctp.c.o
../subprojects/libnvme/examples/mi-mctp.c: In function ‘do_action_endpoint’:
../subprojects/libnvme/examples/mi-mctp.c:665:16: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
  665 |         return rc;
      |                ^~
../subprojects/libnvme/examples/mi-mctp.c:637:13: note: ‘rc’ was declared here
  637 |         int rc;
      |             ^~
```

jk-ozlabs:
Will take a look and get a fix done - feel free to assign to me.

jk-ozlabs:
Looks like this requires specific CFLAGS to reproduce:

```sh
CFLAGS=-Wmaybe-uninitialized meson obj --optimization=1
```

However, unlike https://github.com/linux-nvme/nvme-cli/issues/1647 , I can reproduce this on Debian.

PR #469, opened 2022-09-09
examples/mi-mctp: fix warning for possibly uninitialised variable

jk-ozlabs:
gcc-12 with `-Wuninitialized-var` reports a warning for `mi-mctp`:

    examples/mi-mctp.c: In function ‘do_action_endpoint’:
    examples/mi-mctp.c:665:16: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
      665 |         return rc;
          |                ^~
    examples/mi-mctp.c:637:13: note: ‘rc’ was declared here
      637 |         int rc;
          |             ^~

We shouldn't be able to hit this path, but explicitly set rc to an error if we miss an enum case.

Fixes: https://github.com/linux-nvme/libnvme/issues/467


igaw:
Thanks!

PR #474, opened 2022-09-21
fabrics: Add nvmf_get_discovery_log2()

martin-belanger:
TP8010 introduces new values for the Log page Specific field (LSP) in the "Get Discovery Log Page" command. Previously, the LSP field was unused and set to 0. Now, the LSP can carry the following values, which can be OR'ed.
```
enum nvmf_get_log_discovery_lsp {
	NVMF_LOG_DISC_LSP_NONE		= 0,
	NVMF_LOG_DISC_LSP_EXTDLPE	= (1 << 0), // Extended Discovery Log Page Entries
	NVMF_LOG_DISC_LSP_PLEO		= (1 << 1), // Port Local Entries Only
	NVMF_LOG_DISC_LSP_ALLSUBE	= (1 << 2), // All NVM Subsystem Entries
};
```
The existing API used to issue the "Get Discovery Log Page" command, `nvmf_get_discovery_log()`, does not allow specifying the LSP and always sets it to 0. This pull request introduces `nvmf_get_discovery_log2()`, which takes an `args` parameter allowing users to specify every field of the "Get Discovery Log Page" command and providing a more generic API. 

To determine which of the above LSP features a Discovery Controller supports, TP8010 extends the "Get supported log pages" command for LID=70h (Discovery) by adding the following definitions (e.g. `PLEOS` -> `PLEO Supported`). 
```
enum nvmf_get_log_discovery_lid_support {
	NVMF_LOG_DISC_LID_NONE		= 0,
	NVMF_LOG_DISC_LID_EXTDLPES	= (1 << 0),
	NVMF_LOG_DISC_LID_PLEOS		= (1 << 1),
	NVMF_LOG_DISC_LID_ALLSUBES	= (1 << 2),
};
```

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -340,6 +340,7 @@ LIBNVME_1_0 {
 		nvmf_default_config;
 		nvmf_eflags_str;
 		nvmf_get_discovery_log;
+		nvmf_get_discovery_log2;
 		nvmf_hostid_from_file;
 		nvmf_hostnqn_from_file;
 		nvmf_hostnqn_generate;
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -780,30 +780,10 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 	return NULL;
 }
 
-static int nvme_discovery_log(int fd, __u32 len, struct nvmf_discovery_log *log, bool rae)
-{
-	struct nvme_get_log_args args = {
-		.args_size = sizeof(args),
-		.fd = fd,
-		.nsid = NVME_NSID_NONE,
-		.lsp = NVME_LOG_LSP_NONE,
-		.lsi = NVME_LOG_LSI_NONE,
-		.uuidx = NVME_UUID_NONE,
-		.timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
-		.result = NULL,
-		.lid = NVME_LOG_LID_DISCOVER,
-		.log = log,
-		.len = len,
-		.csi = NVME_CSI_NVM,
-		.rae = rae,
-		.ot = false,
-	};
-
-	return nvme_get_log_page(fd, 4096, &args);
-}
-
-int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
-			   int max_retries)
+int nvmf_get_discovery_log2(nvme_ctrl_t c,
+			    struct nvme_get_log_args *args,
+			    struct nvmf_discovery_log **logp,
+			    int max_retries)
 {
 	nvme_root_t r = c->s && c->s->h ? c->s->h->r : NULL;
 	struct nvmf_discovery_log *log = NULL;
@@ -811,6 +791,9 @@ int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
 	const char *name = nvme_ctrl_get_name(c);
 	uint64_t genctr, numrec;
 	unsigned int size;
+	int fd = nvme_ctrl_get_fd(c);
+
+	args->fd = fd;
 
 	do {
 		size = sizeof(struct nvmf_discovery_log);
@@ -826,7 +809,10 @@ int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
 
 		nvme_msg(r, LOG_DEBUG, "%s: get header (try %d/%d)\n",
 			name, retries, max_retries);
-		ret = nvme_discovery_log(nvme_ctrl_get_fd(c), size, log, true);
+		args->rae = true;
+		args->len = size;
+		args->log = log;
+		ret = nvme_get_log_page(fd, 4096, args);
 		if (ret) {
 			nvme_msg(r, LOG_INFO,
 				 "%s: discover try %d/%d failed, error %d\n",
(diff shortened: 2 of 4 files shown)


src/nvme/fabrics.c
```diff
@@ -780,37 +780,20 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
...
-
-	return nvme_get_log_page(fd, 4096, &args);
-}
-
-int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
-			   int max_retries)
+int nvmf_get_discovery_log2(nvme_ctrl_t c,
+			    struct nvme_get_log_args *args,
+			    struct nvmf_discovery_log **logp,
+			    int max_retries)
```
igaw, line 786: Maybe we should just introduce a variant (and keep it even for an 2.x update where we could change the arguments of the original function) with `struct nvme_get_log_args` argument and have a better name (though I don't have a good idea here `nvmf_get_discover_log_extended`?

src/nvme/fabrics.h
```diff
@@ -197,13 +197,52 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
...
+ * 	.args_size = sizeof(args),
+ * 	.fd = 0,			< Set internally
+ * 	.nsid = NVME_NSID_NONE,
+ * 	.lsp = NVMF_LOG_DISC_LSP_NONE,	< See enum nvmf_get_log_discovery_lsp
+ * 	.lsi = NVME_LOG_LSI_NONE,
+ * 	.uuidx = NVME_UUID_NONE,
+ * 	.timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
+ * 	.result = NULL,
+ * 	.lid = NVME_LOG_LID_DISCOVER,
+ * 	.log = NULL,			< Set internally
```
igaw, line 229: Obviously if we drop the log argument, this one needs to be set correctly.

src/nvme/types.h
```diff
@@ -4856,6 +4856,34 @@ enum nvmf_tcp_sectype {
...
+ * @NVMF_LOG_DISC_LSP_EXTDLPE:	Extended Discovery Log Page Entries
+ * @NVMF_LOG_DISC_LSP_PLEO:	Port Local Entries Only
+ * @NVMF_LOG_DISC_LSP_ALLSUBE:	All NVM Subsystem Entries
+ */
+enum nvmf_get_log_discovery_lsp {
+	NVMF_LOG_DISC_LSP_NONE		= 0,
+	NVMF_LOG_DISC_LSP_EXTDLPE	= (1 << 0),
+	NVMF_LOG_DISC_LSP_PLEO		= (1 << 1),
+	NVMF_LOG_DISC_LSP_ALLSUBE	= (1 << 2),
+};
```
igaw, line 4885: Could you introduce this as a separate commit with the commit message containing the information from the PR message?

src/libnvme.map
```diff
@@ -340,6 +340,7 @@ LIBNVME_1_0 {
 		nvmf_default_config;
 		nvmf_eflags_str;
 		nvmf_get_discovery_log;
+		nvmf_get_discovery_log2;
```
igaw, line 343: This part needs to go to the `LIBNVME_1_2` section

igaw (changes requested):

src/nvme/types.h
martin-belanger: Will do.

src/nvme/fabrics.h
martin-belanger: Actually, the function dynamically allocates memory based on the responses it gets from the Discovery Controller (DC). That is, it performs a Get Discovery Log Page (DLP) using just the "minimum memory" required to Get the DLP header which contains the number of DLP Entries (DLPE) that are available at the DC. The number of DLPEs is used to determine how much memory will be required to perform a second Get DLP command. But before allocating a second larger buffer, `nvmf_get_discovery_log()` frees the "minimum memory" buffer previously allocated. The second larger buffer is the one that gets returned in `.log` and `logp`. 

Note that if the DLP changes while being retrieved (indicated by a different `genctr` between the first and second Get DLP), then the algorithm starts all over again freeing and allocating new memory. So, it's not possible for the caller to pre-allocate memory and pass it in using "`.log`".

Maybe passing `struct nvme_get_log_args args` is not the best approach for the Get DLP command. When we look at the structure (below) we see that most of the parameters are not configurable and must be set to specific values. In fact, I believe that only the `lsp`, `timeout`, and possibly the `csi` could be configured by a user when performing a Get DLP command.

```
struct nvme_get_log_args args = {
	.args_size = sizeof(args),
	.fd = 0,			< Set internally
	.nsid = NVME_NSID_NONE,
	.lsp = NVMF_LOG_DISC_LSP_NONE,	< See enum nvmf_get_log_discovery_lsp
	.lsi = NVME_LOG_LSI_NONE,
	.uuidx = NVME_UUID_NONE,
	.timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
	.result = NULL,
	.lid = NVME_LOG_LID_DISCOVER,
	.log = NULL,			< Set internally
	.len = 0,			< Set internally
	.csi = NVME_CSI_NVM,
	.rae = false,			< Set internally
	.ot = false,
};
```

Let me know if I should take a different approach than passing `struct nvme_get_log_args args`. I could, for example, define a new struct containing a smaller number of arguments (only those that are truly configurable).

igaw: The args approach seems indeed a bit wrong in this case. Alright, let's introduce a function with just the args you need. Maybe we find also a matching name along side then. It sounds like the new function is used in the scenario where have this kind of 'nested' lookups.

PR #476, opened 2022-09-22
fabrics: Add new TP8010 definitions

martin-belanger:
Submitting as a separate commit (and pull request) per discussion: https://github.com/linux-nvme/libnvme/pull/474#discussion_r977523864

This is to prepare for additional TP8010 changes related to the Get Discovery Log Page command. More precisely, the ability to set the LSP field (Log page SPecific) of the Get Discovery Log Page command.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -4857,6 +4857,34 @@ enum nvmf_tcp_sectype {
 };
 
 /**
+ * enum nvmf_log_discovery_lid_support - Discovery log specific support
+ * @NVMF_LOG_DISC_LID_NONE:	None
+ * @NVMF_LOG_DISC_LID_EXTDLPES:	Extended Discovery Log Page Entries Supported
+ * @NVMF_LOG_DISC_LID_PLEOS:	Port Local Entries Only Supported
+ * @NVMF_LOG_DISC_LID_ALLSUBES:	All NVM Subsystem Entries Supported
+ */
+enum nvmf_log_discovery_lid_support {
+	NVMF_LOG_DISC_LID_NONE		= 0,
+	NVMF_LOG_DISC_LID_EXTDLPES	= (1 << 0),
+	NVMF_LOG_DISC_LID_PLEOS		= (1 << 1),
+	NVMF_LOG_DISC_LID_ALLSUBES	= (1 << 2),
+};
+
+/**
+ * enum nvmf_log_discovery_lsp - Discovery log specific field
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -4856,6 +4856,34 @@ enum nvmf_tcp_sectype {
...
 };
 
+/**
+ * enum nvmf_get_log_discovery_lid_support - Discovery log specific support
+ * @NVMF_LOG_DISC_LID_NONE:	None
+ * @NVMF_LOG_DISC_LID_EXTDLPES:	Extended Discovery Log Page Entries Supported
+ * @NVMF_LOG_DISC_LID_PLEOS:	Port Local Entries Only Supported
+ * @NVMF_LOG_DISC_LID_ALLSUBES:	All NVM Subsystem Entries Supported
+ */
+enum nvmf_get_log_discovery_lid_support {
```
igaw, line 4866: The '_get_'  part should be dropped to match with the rest. Also the spec doesn't mention anything about 'get' in the `5.16.1.23 Discovery Log Page (Log Identifier 70h)` section

src/nvme/types.h
```diff
@@ -4856,6 +4856,34 @@ enum nvmf_tcp_sectype {
...
+};
+
+/**
+ * enum nvmf_get_log_discovery_lsp - Discovery log specific field
+ * @NVMF_LOG_DISC_LSP_NONE:	None
+ * @NVMF_LOG_DISC_LSP_EXTDLPE:	Extended Discovery Log Page Entries
+ * @NVMF_LOG_DISC_LSP_PLEO:	Port Local Entries Only
+ * @NVMF_LOG_DISC_LSP_ALLSUBE:	All NVM Subsystem Entries
+ */
+enum nvmf_get_log_discovery_lsp {
```
igaw, line 4880: Same here, the '_get_' seems to be wrong here.

igaw (changes requested):

src/nvme/types.h
martin-belanger: Agreed. I will drop the `get`. 
FWIW, I was trying to stay consistent with another "get log" LSP enum (i.e. `nvme_cmd_get_log_telemetry_host_lsp`)

igaw: Ah, good to know. I think we should rename that one for 2.x!

igaw (approved):

PR #479, opened 2022-09-24
github workflows: Add armhf & powerpc64le cross builds

jk-ozlabs:
This RFC change adds a github workflow for a cross-compile of libnvme on armhf & powerpc64le We use a similar setup to the build-distro job, but install a suitable cross compiler and provide a cross config to meson.

We currently setup with -Dpython=false, as the setup-python@v4 seems to assume a native python build. We may be able to remove this later.

The suitability of this will depend on the resources available for actions; doing some test runs on a private repo shows an execution time of about 1 min 15 sec per cross build - a little less than the build-static job.


jk-ozlabs:
I'd like to see how this goes, and if all is well, add to nvme-cli too.

igaw:
Thanks a lot. This was also something I wanted to add for some time!

PR #482, opened 2022-09-26
mi: Add firmware download and commit commands

jk-ozlabs:
This change adds MI implementations for the Firmware Download and Firmware Commit admin commands, as well as a couple of tests.

As usual, these are designed to match the ioctl API.

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -6,6 +6,8 @@ LIBNVME_MI_1_2 {
 		nvme_mi_admin_ns_attach;
 		nvme_mi_admin_format_nvm;
 		nvme_mi_admin_sanitize_nvm;
+		nvme_mi_admin_fw_download;
+		nvme_mi_admin_fw_commit;
 };
 
 LIBNVME_MI_1_1 {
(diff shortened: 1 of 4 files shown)


src/libnvme-mi.map
```diff
@@ -18,6 +18,8 @@ LIBNVME_MI_1_1 {
 		nvme_mi_admin_set_features;
 		nvme_mi_admin_ns_mgmt;
 		nvme_mi_admin_ns_attach;
+		nvme_mi_admin_fw_download;
+		nvme_mi_admin_fw_commit;
```
igaw, line 22: I think these should go into LIBNVME_MI_1_2

igaw (changes requested):
Just the symbol versioning thing needs to be updated.

src/libnvme-mi.map
jk-ozlabs: ah, we should probably move others into `1_2` too then. There have been a few new definitions since 1.1 was released.

I'll do that as a first change, which this would then apply on top of.

jk-ozlabs: That's now up at https://github.com/linux-nvme/libnvme/pull/484, I'll rebase this change.

jk-ozlabs:
Rebased on to https://github.com/linux-nvme/libnvme/pull/484, so we have the symbols added at the correct version.

igaw (approved):

PR #483, opened 2022-09-26
fabrics: Add nvmf_get_discovery_wargs()

martin-belanger:
TP8010 introduces new values for the Log page Specific field (LSP) in the "Get Discovery Log Page" command. Previously, the LSP field was unused and set to 0. 

This pull request adds function `nvmf_get_discovery_wargs()`. This function is similar to `nvmf_get_discovery_log()`, but takes an extensible "`args`" parameter. `nvmf_get_discovery_wargs()` allows setting more options (including the LSP)  than `nvmf_get_discovery_log()`. 

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -4,6 +4,7 @@ LIBNVME_1_2 {
 	global:
 		nvme_ctrl_get_dhchap_host_key;
 		nvme_ctrl_set_dhchap_host_key;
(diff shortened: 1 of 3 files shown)


igaw:
Okay, let's go with this version. Looks reasonable.

I was pondering on the `struct arg` layout (padding) and it happens to work on 32 and 64 bit. The naming scheme is also way better than appending a 2 the original name.

PR #490, opened 2022-09-29
Init ctrl_id within xfer of libnvme-mi

drakedog2008:
The xfer() requires a `nvme_mi_ctrl_t ctrl` as input so the ctrl_id should be included in `ctrl` instead of from
`nvme_mi_admin_req_hdr.ctrl_id`.


jk-ozlabs:
This will mean that `nvme_mi_admin_xfer` cannot be used on a "remote" controller (ie., not the one specified by `ctrl`), but I think that's acceptable here; if we need that, we can introduce one that operates on the `ep` instead of the `ctrl`.

So: LGTM, thanks!

igaw:
Commit merged after tweaking the commit message title

PR #321, opened 2022-03-30
ioctl: Remove attribute packed and alignedof for args structs

igaw:
The attribute packed is usually used to make sure the data structures
is compatible between different compilers in regards of padding rules.

As we have sorted the members of all argument structs according their
naturual size, there are no holes to pad. This makes the packed
attributed superflous as compilers are agree on the data layout in
this case.

The alignedof attribute is used to tell the outer alignmen of the data
structure because the aligment of a packed data structure is 1.

Anyway, both attributes doen't add any benefits to the layout and
pahole agrees on this.

This is an example the diff between the packed/aligneof version and
the plain version:

 struct nvme_dim_args {
        __u32 *                    result;               /*     0     8 */
        void *                     data;                 /*     8     8 */
@@ -776,4 +777,4 @@
        /* size: 40, cachelines: 1, members: 7 */
        /* padding: 7 */
        /* last cacheline: 40 bytes */
-} __attribute__((__aligned__(8)));
+};

Thus, just remove the attributes.

Signed-off-by: Daniel Wagner <dwagner@suse.de>

Fixes: #285

PR #492, opened 2022-10-04
Add packed to all nvme message struct

drakedog2008:
The NVMe messages are self-padding with resv bytes. However, the `alignof` operator will still show a none-byte value (a.k.a > 1) for the structures. When doing type conversion between the struct and byte array. It makes incompatible with using some libraries such as stdplus::raw, given the library will check the alignment of the types. E.g. https://github.com/openbmc/stdplus/blob/master/include/stdplus/raw.hpp#L96

The packed attribute will force the alignment of the struct always 1 byte.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -738,7 +738,7 @@ struct nvme_id_psd {
 	__le16			actp;
 	__u8			apws;
 	__u8			rsvd23[9];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_ctrl - Identify Controller data structure
@@ -1051,7 +1051,7 @@ struct nvme_id_ctrl {
 
 	struct nvme_id_psd	psd[32];
 	__u8			vs[1024];
-};
+} __attribute__((packed));
 
 /**
  * enum nvme_id_ctrl_cmic - Controller Multipath IO and Namespace Sharing
@@ -1649,7 +1649,7 @@ struct nvme_lbaf {
 	__le16			ms;
 	__u8			ds;
 	__u8			rp;
-};
+} __attribute__((packed));
 
 /**
  * enum nvme_lbaf_rp - This field indicates the relative performance of the LBA
@@ -1816,7 +1816,7 @@ struct nvme_id_ns {
 	struct nvme_lbaf	lbaf[64];
 	__le64			lbstm;
 	__u8			vs[3704];
-};
+} __attribute__((packed));
 
 /**
  * enum nvme_id_nsfeat - This field defines features of the namespace.
@@ -2050,7 +2050,7 @@ struct nvme_ns_id_desc {
 	__u8	nidl;
 	__le16	rsvd;
 	__u8	nid[];
-};
+} __attribute__((packed));
 
 /**
  * enum nvme_ns_id_desc_nidt - Known namespace identifier types
@@ -2099,7 +2099,7 @@ struct nvme_nvmset_attr {
 	__u8			tnvmsetcap[16];
 	__u8			unvmsetcap[16];
 	__u8			rsvd48[80];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_nvmset_list - NVM set list
@@ -2111,7 +2111,7 @@ struct nvme_id_nvmset_list {
 	__u8			nid;
 	__u8			rsvd1[127];
 	struct nvme_nvmset_attr	ent[NVME_ID_NVMSET_LIST_MAX];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_independent_id_ns - Identify - I/O Command Set Independent Identify Namespace Data Structure
@@ -2140,7 +2140,7 @@ struct nvme_id_independent_id_ns {
 	__le16	endgid;
 	__u8	nstat;
 	__u8	rsvd15[4081];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_ns_granularity_desc -	 Namespace Granularity Descriptor
@@ -2150,7 +2150,7 @@ struct nvme_id_independent_id_ns {
 struct nvme_id_ns_granularity_desc {
 	__le64			nszegran;
 	__le64			ncapgran;
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_ns_granularity_list - Namespace Granularity List
@@ -2166,7 +2166,7 @@ struct nvme_id_ns_granularity_list {
 	__u8			rsvd5[27];
 	struct nvme_id_ns_granularity_desc entry[NVME_ID_ND_DESCRIPTOR_MAX];
 	__u8			rsvd288[3808];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_uuid_list_entry - UUID List Entry
@@ -2178,7 +2178,7 @@ struct nvme_id_uuid_list_entry {
 	__u8			header;
 	__u8			rsvd1[15];
 	__u8			uuid[16];
-};
+} __attribute__((packed));
 
 /**
  * enum nvme_id_uuid - Identifier Association
@@ -2202,7 +2202,7 @@ enum nvme_id_uuid {
 struct nvme_id_uuid_list {
 	__u8	rsvd0[32];
 	struct nvme_id_uuid_list_entry entry[NVME_ID_UUID_LIST_MAX];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_ctrl_list - Controller List
@@ -2212,7 +2212,7 @@ struct nvme_id_uuid_list {
 struct nvme_ctrl_list {
 	__le16 num;
 	__le16 identifier[NVME_ID_CTRL_LIST_MAX];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_ns_list - Namespace List
@@ -2220,7 +2220,7 @@ struct nvme_ctrl_list {
  */
 struct nvme_ns_list {
 	__le32 ns[NVME_ID_NS_LIST_MAX];
-};
+} __attribute__((packed));
 
 /**
  * struct nvme_id_ctrl_nvm - I/O Command Set Specific Identify Controller data structure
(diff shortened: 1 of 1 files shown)


igaw:
IIRC, this will break the alignment information of all members inside the struct and leads to alignment warnings when passing structs by value. Your argument what the problem is 'conversion between the struct and byte array' sounds already wrong.

drakedog2008:
Hi Daniel, Thanks for the comment. 

> IIRC, this will break the alignment information of all members inside the struct 

The structures are self-padded, meaning all the elements are aligned without compiler inserted padding. The internal memory allocation should not be impacted with or without the packed attribute. A simple test: 

```c++
struct nvme_smart_log_pack
{
    __u8 critical_warning;
    __u8 temperature[2];
    __u8 avail_spare;
    __u8 spare_thresh;
    __u8 percent_used;
    __u8 endu_grp_crit_warn_sumry;
    __u8 rsvd7[25];
    __u8 data_units_read[16];
    __u8 data_units_written[16];
    __u8 host_reads[16];
    __u8 host_writes[16];
    __u8 ctrl_busy_time[16];
    __u8 power_cycles[16];
    __u8 power_on_hours[16];
    __u8 unsafe_shutdowns[16];
    __u8 media_errors[16];
    __u8 num_err_log_entries[16];
    __le32 warning_temp_time;
    __le32 critical_comp_time;
    __le16 temp_sensor[8];
    __le32 thm_temp1_trans_count;
    __le32 thm_temp2_trans_count;
    __le32 thm_temp1_total_time;
    __le32 thm_temp2_total_time;
    __u8 rsvd232[280];
} __attribute__((packed));

struct nvme_smart_log
{
    __u8 critical_warning;
    __u8 temperature[2];
    __u8 avail_spare;
    __u8 spare_thresh;
    __u8 percent_used;
    __u8 endu_grp_crit_warn_sumry;
    __u8 rsvd7[25];
    __u8 data_units_read[16];
    __u8 data_units_written[16];
    __u8 host_reads[16];
    __u8 host_writes[16];
    __u8 ctrl_busy_time[16];
    __u8 power_cycles[16];
    __u8 power_on_hours[16];
    __u8 unsafe_shutdowns[16];
    __u8 media_errors[16];
    __u8 num_err_log_entries[16];
    __le32 warning_temp_time;
    __le32 critical_comp_time;
    __le16 temp_sensor[8];
    __le32 thm_temp1_trans_count;
    __le32 thm_temp2_trans_count;
    __le32 thm_temp1_total_time;
    __le32 thm_temp2_total_time;
    __u8 rsvd232[280];
};
int main(void)
{
    std::cout << "packed size is: " << sizeof(nvme_smart_log_pack) << std::endl;
    std::cout << "unpacked size is: " << sizeof(nvme_smart_log) << std::endl;
}
```
Output: 
```
packed size is: 512
unpacked size is: 512
```

> and leads to alignment warnings when passing structs by value. 

Can you be more specific on the warning? 

> Your argument what the problem is 'conversion between the struct and byte array' sounds already wrong.

The type conversion when using `(struct nvme_struct_type*)(byte_array)` or `reinterpret_cast<nvme_struct_type*>(byte_array)` works ok. The problem is that some library, such as stdplus, checks the alignment before type conversion. It is to prevent the implicit padding. e.g. for a struct of 

```c++
struct test{
unit8_t a;
uint32_t b;
};
```
The address of `b` is actually `&a + sizeof(uint32_t)` instead of `&a + sizeof(a)`. So the library force the caller to use a `uint32_t array[]`, instead `uint8_t array[]` as the underlying buffer in order to force the alignment between the structure elements and the array index. 

Given packed attribute will not change the internal memory of nvme structures (explained above). But it will be much convenient for the client to use a unified char array as buffer to transfer all nvme structure objects and convert the type back and force.

keithbusch:
Let me see if I understand this. You want to define some buffer as a `char buf[LARGE_SIZE]`, and some time later cast that to an nvme struct like `struct nvme_smart_log *log = &buf[offset]`. The problem is that using some library that handles casting has a rule that the address alignment of the source buffer needs to match the aligned size of the type it's being cast to?

drakedog2008:
Hi Keith, 

That is correct.

igaw:
> ```
> packed size is: 512
> unpacked size is: 512
> ```

This is not the issue. The problem is that the alignment information of all members of the struct is byte aligned with the packed attribute.

> > and leads to alignment warnings when passing structs by value.
> 
> Can you be more specific on the warning?

IIRC, the warnings are from `-Waddress-of-packed-member`

https://stackoverflow.com/questions/68640690/arm-gcc-address-of-packed-struct-warning

And it's a real warning we got in our library when we played with the packed attribute. The 
packed attribute removal commit message  does mention this
https://github.com/linux-nvme/libnvme/pull/321/commits/f274eeb2ae6264f32abb919174ee80c1a071a5c4

> The address of b is actually &a + sizeof(uint32_t) instead of &a + sizeof(a). So the library force the caller to use a uint32_t array[], instead uint8_t array[] as the underlying buffer in order to force the alignment between the structure elements and the array index.

None of the structs in ioctl.h suffer from this problem (if they do it's a bug in the data struct definition). All members are naturally aligned with explicit padding bytes. And your very example is the reason why we don't have packed attribute. The member alignment information should not be casted away.

Another thing you need to consider is padding outside of the structure. So if you have an array of data structures with members align on 8bytes boundaries the data structure itself needs to be aligned on 8 byte boundaries. With the packed attribute this information is lost as well.

The C++ library is correct with the warning. None of these data structures are one byte aligned.

drakedog2008:
> > ```
> > packed size is:  #512
> > unpacked size is: 512
> >  #```
> 
> This is not the issue. The problem is that the alignment information of all members of the struct is byte aligned with the packed attribute.
> 
the byte alignment is actually what I am trying to achieve here. 
> > > and leads to alignment warnings when passing structs by value.
> > 
> > 
> > Can you be more specific on the warning?
> 
> IIRC, the warnings are from `-Waddress-of-packed-member`
> 
> https://stackoverflow.com/questions/68640690/arm-gcc-address-of-packed-struct-warning
> 
> And it's a real warning we got in our library when we played with the packed attribute. The packed attribute removal commit message does mention this [f274eeb](https://github.com/linux-nvme/libnvme/commit/f274eeb2ae6264f32abb919174ee80c1a071a5c4)
> 
I saw the contradict :) 

We all know the nvme structures are well padded w/o holes. But what you elaborated is the user may requires the alignment information to construct an aligned buffer type for **a single struct**. 

What I was trying to do is to create a buffer for **all struct** regardless of alignment info. 

Just curious, does some client really need to know the alignment info? If we are allocating the buffer for a specific struct type, it is easy for us to use a std::array<unit8_t, 4> to replace uint32_t? 

> > The address of b is actually &a + sizeof(uint32_t) instead of &a + sizeof(a). So the library force the caller to use a uint32_t array[], instead uint8_t array[] as the underlying buffer in order to force the alignment between the structure elements and the array index.
> 
> None of the structs in ioctl.h suffer from this problem (if they do it's a bug in the data struct definition). All members are naturally aligned with explicit padding bytes. And your very example is the reason why we don't have packed attribute. The member alignment information should not be casted away.
> 
ioctl.h is for internal message definitions, which (maybe) is used only by a single module. What I concerns are the defs in `nvme/types.h`, which is installed and referred across modules, processes, or even machines. And it makes the padding info really untrustful. 

IMHO, all the types are used for serializing/deserializing across boundaries should be packed physically and semantically? 
e.g. https://github.com/openbmc/phosphor-ipmi-blobs/blob/83f9992cc60dedb6a2a964a733fd35c89c8a0cfa/ipmi.hpp

> Another thing you need to consider is padding outside of the structure. So if you have an array of data structures with members align on 8bytes boundaries the data structure itself needs to be aligned on 8 byte boundaries. With the packed attribute this information is lost as well.
> The C++ library is correct with the warning. None of these data structures are one byte aligned.

That is correct. ATM, I have to no plan to transfer multiple structures data within one dbus(an IPC protocol) call. But if I do so, the transmission layer  will do the padding. e.g. dbus do 8 bytes padding for all fields ([ref](https://dbus.freedesktop.org/doc/api/html/structDBusHeader.html#details)).

keithbusch:
Unless I'm missing something, I think this seems okay. The nvme structures are generally padded with aligned fields correctly, though the one exception is the SMART log's temperature field. If you're gonna pack the struct, then I suppose we can change it to the native __le16 type if the endian conversions are unaligned safe (too busy to check right now).

igaw:
> the byte alignment is actually what I am trying to achieve here.

Sure but I am still not convinced it is the right thing to do.

> > IIRC, the warnings are from `-Waddress-of-packed-member`
> > https://stackoverflow.com/questions/68640690/arm-gcc-address-of-packed-struct-warning
> > And it's a real warning we got in our library when we played with the packed attribute. The packed attribute removal commit message does mention this [f274eeb](https://github.com/linux-nvme/libnvme/commit/f274eeb2ae6264f32abb919174ee80c1a071a5c4)
> 
> I saw the contradict :)

It depends as they are fulfilling different requirements. The question is which requirement is the right one. And this we haven't agreed on (yet).

> We all know the nvme structures are well padded w/o holes. But what you elaborated is the user may requires the alignment information to construct an aligned buffer type for **a single struct**.

Yes, the alignment information is important. The compiler issues different load instruction depended on this information. You can't just assume the compiler magically knows what the type is supposed to be. The packed attribute is the sibling of the void pointer.

> What I was trying to do is to create a buffer for **all struct** regardless of alignment info.

I don't understand this part. The alignment information *is* important if you access the data types by it natural size. That's why the compiler complains if you try to access a pointer which is correctly aligned. On some platforms this will terminate the program.

> Just curious, does some client really need to know the alignment info? If we are allocating the buffer for a specific struct type, it is easy for us to use a std::array<unit8_t, 4> to replace uint32_t?

When you access those members directly, the alignment informant is necessary. You can hide by adding getter/setters
but this is also pretty ugly. 

> ioctl.h is for internal message definitions, which (maybe) is used only by a single module. What I concerns are the defs in `nvme/types.h`, which is installed and referred across modules, processes, or even machines. And it makes the padding info really untrustful.

The C/C++ has pretty clear rules on this as far I remember. I really would like to see a proof for your claim. 

Note also many Linux kernel header do not use packed for wired data structures (networking).

> IMHO, all the types are used for serializing/deserializing across boundaries should be packed physically and semantically? e.g. https://github.com/openbmc/phosphor-ipmi-blobs/blob/83f9992cc60dedb6a2a964a733fd35c89c8a0cfa/ipmi.hpp

Good point, I think this is the core of our discussion. Don't get me wrong, I am not saying no to your change per se, I just want to avoid having this discussion in a half a year again with someone arguing the other way.

igaw:
> > Just curious, does some client really need to know the alignment info? If we are allocating the buffer for a specific struct type, it is easy for us to use a std::array<unit8_t, 4> to replace uint32_t?
> 
> When you access those members directly, the alignment informant is necessary. You can hide by adding getter/setters but this is also pretty ugly.

I missed the point that we are actually using the `le*_to_cpu` getters. So this renders this point invalid.

igaw:
During [ALPSS](https://alpss.at) Christoph Hellwig, Keith and I discussed this. Christoph (strongly) pointed out that if one wants to use the `packet` attribute 'the question is wrong'. It should be avoided. 'There is almost no case where it is correct'.

As this is in very much in line with my original feedback I reject this patch.

drakedog2008:
Hi Daniel, 

Thanks for sharing the info. I got your concerns and the discussion was quite informative and meaningful.

PR #493, opened 2022-10-11
libnvme: fix a memory leak in nvme_mi_open_mctp()

maurizio-lombardi:
If the call to ops.socket() fails, the mctp pointer will be leaked.


igaw:
Just modified the subject using 'mi' instead of 'libnvme'

PR #497, opened 2022-10-18
util: Add simple UUID type

igaw:
Provide our own UUID type as there is no point to have a dependency on libuuid just for a couple function we use.


igaw:
FWIW, I've tested if the current version of nvme-cli builds again this change. All seems fine.

PR #499, opened 2022-10-19
mi: fix get_log_page chunked offset check

jk-ozlabs:
In our get_log_page helper, we're incorrectly checking the requested chunk offset against the chunk len, rather than the overall length - this means we can't query anything but the first chunk.

This change fixes this to check against the overall length instead, and adds an additional check to ensure we're not requesting more than the full data size.

We also add a testcase for the chunking code.

Reported-by: Hao Jiang <jianghao@google.com>


jk-ozlabs:
@drakedog2008 : this should fix the `nvme_mi_get_log_page` issue you found, could you give this a test?

PR #505, opened 2022-10-25
mi: add nvme_mi_admin_admin_passthru

jinliangw:
Similar to nvme_admin_passthru, send a customized NVMe Admin command request message and get the corresponding response message. Currently, it only supports data xfer size <= 4096.

With corresponding nvme-cli changes, verified below commands:
```
# non-data
# ./nvme admin-passthru --opcode 0x0A  --cdw10=0x07 mctp:1,16:0
Admin Command Get Features is Success and result: 0x000e000e

# identify control(read 4096B)
#./nvme admin-passthru --opcode 0x06 --cdw10=0x01 --data-len=4096 --read mctp:1,16:0

# smart log(read 512B)
#./nvme admin-passthru --opcode 0x02 --cdw10=0x7f0002 --data-len=512 --read mctp:1,16:0
```
Write command is not tested because I cannot find a device which supports admin commands with data direction host to controller for now.  I think this should be OK as almost all admin command in OOB are non-data or controller to host.

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -8,6 +8,7 @@ LIBNVME_MI_1_2 {
 		nvme_mi_admin_sanitize_nvm;
 		nvme_mi_admin_fw_download;
 		nvme_mi_admin_fw_commit;
+		nvme_mi_admin_admin_passthru;
 		nvme_mi_status_to_string;
 };
 
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -419,6 +419,88 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
 	return 0;
 }
 
+int nvme_mi_admin_admin_passthru(nvme_mi_ctrl_t ctrl, __u8 opcode, __u8 flags,
+				 __u16 rsvd, __u32 nsid, __u32 cdw2, __u32 cdw3,
+				 __u32 cdw10, __u32 cdw11, __u32 cdw12,
+				 __u32 cdw13, __u32 cdw14, __u32 cdw15,
+				 __u32 data_len, void *data, __u32 metadata_len,
+				 void *metadata, __u32 timeout_ms, __u32 *result)
+{
+	/* Input parameters flags, rsvd, metadata, metadata_len are not used */
+	struct nvme_mi_admin_resp_hdr resp_hdr;
+	struct nvme_mi_admin_req_hdr req_hdr;
+	struct nvme_mi_resp resp;
+	struct nvme_mi_req req;
+	int rc;
+	int direction = opcode & 0x3;
+	bool has_write_data = (direction == NVME_DATA_TFR_HOST_TO_CTRL)
+		&& (data != NULL) && (data_len != 0);
+	bool has_read_data = (direction == NVME_DATA_TFR_CTRL_TO_HOST)
+		&& (data != NULL) && (data_len != 0);
+
+	if (direction == NVME_DATA_TFR_BIDIRECTIONAL) {
+		nvme_msg(ctrl->ep->root, LOG_ERR,
+			"nvme_mi_admin_admin_passthru doesn't support bidirectional commands\n");
+		errno = EINVAL;
+		return -1;
+	}
+
+	if (data_len > 4096) {
+		nvme_msg(ctrl->ep->root, LOG_ERR,
+			"nvme_mi_admin_admin_passthru doesn't support data_len over 4096 bytes.\n");
+		errno = EINVAL;
+		return -1;
+	}
+
+	if (timeout_ms > nvme_mi_ep_get_timeout(ctrl->ep)) {
+		/* Set timeout if user needs a bigger timeout */
+		nvme_mi_ep_set_timeout(ctrl->ep, timeout_ms);
+	}
+
+	nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id, opcode);
+	req_hdr.cdw1 = cpu_to_le32(nsid);
+	req_hdr.cdw2 = cpu_to_le32(cdw2);
+	req_hdr.cdw3 = cpu_to_le32(cdw3);
+	req_hdr.cdw10 = cpu_to_le32(cdw10);
+	req_hdr.cdw11 = cpu_to_le32(cdw11);
+	req_hdr.cdw12 = cpu_to_le32(cdw12);
+	req_hdr.cdw13 = cpu_to_le32(cdw13);
+	req_hdr.cdw14 = cpu_to_le32(cdw14);
+	req_hdr.cdw15 = cpu_to_le32(cdw15);
+	req_hdr.doff = 0;
+	if (data_len != 0) {
+		req_hdr.dlen = cpu_to_le32(data_len);
+		/* Bit 0 set to 1 means DLEN contains a value */
+		req_hdr.flags = 0x1;
+	}
+
+	if (has_write_data) {
+		req.data = data;
+		req.data_len = data_len;
+	}
+
+	nvme_mi_calc_req_mic(&req);
+
+	nvme_mi_admin_init_resp(&resp, &resp_hdr);
+
+	if (has_read_data) {
+		resp.data = data;
+		resp.data_len = data_len;
+	}
+
+	rc = nvme_mi_submit(ctrl->ep, &req, &resp);
+	if (rc)
+		return rc;
+
+	rc = nvme_mi_admin_parse_status(&resp, result);
+	if (has_read_data && (resp.data_len != data_len)) {
+		errno = EPROTO;
+		return -1;
+	}
+
+	return 0;
+}
+
 int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
 				   struct nvme_identify_args *args,
 				   off_t offset, size_t size)
--- a/src/nvme/mi.h
+++ b/src/nvme/mi.h
@@ -956,6 +956,46 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
 		       size_t *resp_data_size);
 
 /**
+ * nvme_mi_admin_admin_passthru() - Submit an nvme admin passthrough command
+ * @ctrl: Controller to send command to
+ * @opcode:	The nvme admin command to send
+ * @flags:	NVMe command flags (not used)
+ * @rsvd:	Reserved for future use
+ * @nsid:	Namespace identifier
+ * @cdw2:	Command dword 2
+ * @cdw3:	Command dword 3
+ * @cdw10:	Command dword 10
+ * @cdw11:	Command dword 11
+ * @cdw12:	Command dword 12
+ * @cdw13:	Command dword 13
+ * @cdw14:	Command dword 14
+ * @cdw15:	Command dword 15
+ * @data_len:	Length of the data transferred in this command in bytes
+ * @data:	Pointer to user address of the data buffer
+ * @metadata_len:Length of metadata transferred in this command(not used)
+ * @metadata:	Pointer to user address of the metadata buffer(not used)
+ * @timeout_ms:	How long to wait for the command to complete
+ * @result:	Optional field to return the result from the CQE dword 0
+ *
+ * Send a customized NVMe Admin command request message and get the corresponding
+ * response message.
+ *
+ * This interface supports no data, host to controller and controller to
+ * host but it doesn't support bidirectional data transfer.
+ * Also this interface only supports data transfer size range [0, 4096] (bytes)
+ * so the & data_len parameter must be less than 4097.
+ *
+ * Return: The nvme command status if a response was received (see
+ * &enum nvme_status_field) or -1 with errno set otherwise.
+ */
+int nvme_mi_admin_admin_passthru(nvme_mi_ctrl_t ctrl, __u8 opcode, __u8 flags,
+				 __u16 rsvd, __u32 nsid, __u32 cdw2, __u32 cdw3,
+				 __u32 cdw10, __u32 cdw11, __u32 cdw12,
+				 __u32 cdw13, __u32 cdw14, __u32 cdw15,
+				 __u32 data_len, void *data, __u32 metadata_len,
+				 void *metadata, __u32 timeout_ms, __u32 *result);
+
+/**
  * nvme_mi_admin_identify_partial() - Perform an Admin identify command,
  * and retrieve partial response data.
  * @ctrl: Controller to process identify command


src/nvme/mi.c
```diff
@@ -18,6 +17,12 @@
 #include "mi.h"
 #include "private.h"
 
+/* 01b: host to controller */
+#define IS_DIRECTION_WRITE(op)  (((op)&0x3) == 0x1)
```
igaw, line 21: Use static inline instead macros for this kind of stuff.

src/nvme/mi.h
```diff
@@ -955,6 +955,44 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+ * Return: The nvme command status if a response was received (see
+ * &enum nvme_status_field) or -1 with errno set otherwise.
+ */
+int nvme_mi_admin_admin_passthru(nvme_mi_ctrl_t ctrl, __u8 opcode, __u8 flags,
+				 __u16 rsvd, __u32 nsid, __u32 cdw2, __u32 cdw3,
+				 __u32 cdw10, __u32 cdw11, __u32 cdw12,
+				 __u32 cdw13, __u32 cdw14, __u32 cdw15,
+				 __u32 data_len, void *data, __u32 metadata_len,
+				 void *metadata, __u32 timeout_ms, __u32 *result);
+
```
igaw, line 995: `struct nvme_passthru_cmd`  ?

jinliangw: The in-band API `nvme_admin_passthru` and `nvme_admin_passthru64` are still using the same long argument list.  If we  use `struct nvme_passthru_cmd` here, we need to change the in-band APIs too. I'm not sure whether in-band API change will cause any dependency issues.

src/nvme/mi.c
jinliangw: Removed.

src/nvme/mi.h
igaw: Ah, okay, I thought we were using the `struct nvme_passthru_cmd` for those commands as well. I suppose we just keep it consistent for this interface.

src/libnvme-mi.map
```diff
@@ -30,6 +30,7 @@ LIBNVME_MI_1_1 {
 		nvme_mi_admin_xfer;
 		nvme_mi_admin_security_send;
 		nvme_mi_admin_security_recv;
+		nvme_mi_admin_admin_passthru;
```
igaw, line 33: This should go into the `LIBNVME_MI_1_2`

src/nvme/mi.c
```diff
@@ -419,6 +418,80 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+	req_hdr.cdw15 = cpu_to_le32(cdw15);
+	req_hdr.doff = 0;
+	if (data_len != 0) {
+		req_hdr.dlen = cpu_to_le32(data_len);
+		/* Bit 0 set to 1 means DLEN contains a value */
+		req_hdr.flags = 0x1;
+	}
+
+	/* write data */
+	if((opcode & 0x3) == NVME_DATA_TFR_HOST_TO_CTRL) {
```
igaw, line 464: `NVME_DATA_TFR_HOST_TO_CTRL`  looks really like the wrong enum type. This enum is for the `nvme_get_feature_length2` function.

And why not just introduce static inline function instead the original macro?

igaw (changes requested):

src/nvme/mi.c
jinliangw: The `enum nvme_data_tfr` is defined according to the description: opcode[1:0]"Indicates the data transfer direction of the command. All options to the command shall transfer data as specified or transfer no data. All commands, including vendor specific commands, shall follow this convention: 00b = no data transfer; 01b = host to controller; 10b = controller to host; 11b = bidirectional." in NVMe spec. So this enum apply to all Admin commands, not only set_feature/get_feature.

The `static inline function` looks like:
```
static inline int nvme_get_admin_data_direction(__u8 opcode) {
    return opcode & 0x3;
}
```

The use case looks like:
```
if(nvme_get_admin_data_direction(opcode) == NVME_DATA_TFR_HOST_TO_CTRL) {
```

VS current sytle:
```
if((opcode & 0x3) == NVME_DATA_TFR_HOST_TO_CTRL) {
```

I prefer my current style as it will save me some typing effort.

jk-ozlabs:
> Write command is not tested because I cannot find a device which supports admin commands with data direction host to controller for now. I think this should be OK as almost all admin command in OOB are non-data or controller to host.

I'll give this a test with a Firmware Image Download command, which is host-to-controller.

src/nvme/mi.c
```diff
@@ -419,6 +418,80 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+	struct nvme_mi_req req;
+	int rc;
+	if(data_len > 4096) {
+		nvme_msg(ctrl->ep->root, LOG_ERR, "nvme_mi_admin_admin_passthru"
+			"does not supports data_len over 4096 bytes.\n");
+		errno = EINVAL;
+		return -1;
+	}
+
+	if(timeout_ms > nvme_mi_ep_get_timeout(ctrl->ep)) {
```
jk-ozlabs, line 441: space after the `if` here; can you run this through checkpatch?

src/nvme/mi.c
jk-ozlabs: Typing effort is almost never a good optimisation; this code will be read *many* more times than it is typed. Please optimise for clarity instead.

src/nvme/mi.h
```diff
@@ -955,6 +955,44 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+ * @cdw2:	Command dword 2
+ * @cdw3:	Command dword 3
+ * @cdw10:	Command dword 10
+ * @cdw11:	Command dword 11
+ * @cdw12:	Command dword 12
+ * @cdw13:	Command dword 13
+ * @cdw14:	Command dword 14
+ * @cdw15:	Command dword 15
+ * @data_len:	Length of the data transferred in this command in bytes
+ * @data:	Pointer to user address of the data buffer
```
jk-ozlabs, line 974: You have a single buffer + length for both command and response data. However, it's possible for a command to be bi-directional.

Should we include a separate response buffer + length here? Then we don't need to guess the direction of the API from the command opcode.

(However, this will diverge from the existing direct ioctl API; I'm happy to discuss the better solution here. @igaw , how do you intend to handle bidirectional commands through the existing API?)

src/nvme/mi.c
```diff
@@ -8,7 +8,6 @@
 
 #include <errno.h>
 #include <stdlib.h>
-#include <stdlib.h>
```
jk-ozlabs, line 11: Maybe do this as a separate cleanup.

src/nvme/mi.h
```diff
@@ -955,6 +955,44 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+ * @metadata_len:Length of metadata transferred in this command(not used)
+ * @metadata:	Pointer to user address of the metadata buffer(not used)
```
jk-ozlabs, line 976: I understand these to be present to match the existing API. However, if we break that with the bidirectional support, we could remove these too.

jk-ozlabs (commented):
Looks good - a couple of things, mainly a query on how we can handle the data direction, or whether we want to maintain strict compat with the ioctl-based API.

In the naming-things regard: While `nvme_mi_admin_admin_passthru` is strictly  consistent with the existing naming scheme (`nvme_mi` base prefix + `admin` command set + `admin_passthru` command name), I'm happy to compress this to `nvme_mi_admin_passthru`, which is a little less verbose.

src/nvme/mi.c
igaw: > The enum nvme_data_tfr is defined according to the description:

Yes and no. It's an argument type for the `nvme_get_feature_length2`. That is also the reason why it lives in `utils.h`. But the biggest clue is that we don't define any of the values, we let the compiler choose. So you would first need to update `enum nvme_data_tfr` accordingly. 

I asked to transform `IS_DIRECTION_WRITE` to a function. The resulting code would be something
like `if (nvme_is_direction_write(op)) { ...}` or whatever the function is supposed be named to match
the current coding style.

src/nvme/mi.h
igaw: Hmm, I was not involved when those APIs where designed. Maybe there was specific reason why there is only one buffer. Need to figure this out first.

But at this stage I think we should not start introducing different API styles for the pass through commands. As I understood the mi version of the functions are just using a different transport for the same commands. 

So this sounds like 2.x material.

src/nvme/mi.h
igaw: Ok. maybe @keithbusch can remember why the APIs are designed this way.

igaw: IIRC, the API is following the IOCTL interface. So this might explain why the buffer is reused. If this it the case I don't think we should start trying to change this model in libnvme and hide this. It usually doesn't end to good if one tries to be clever.

src/nvme/mi.c
jinliangw: 1. Defined values in `enum nvme_data_tfr`. No sure if you need to move it from `util.h` to other header file.
2. Changed usage to:
```
       int direction = opcode & 0x3;
       bool has_write_data = (direction == NVME_DATA_TFR_HOST_TO_CTRL)
               && (data != NULL) && (data_len != 0);
       bool has_read_data = (direction == NVME_DATA_TFR_CTRL_TO_HOST)
               && (data != NULL) && (data_len != 0);
       if (direction == NVME_DATA_TFR_BIDIRECTIONAL) {...
```
Please let me know if you still want to add 3 helper functions(`nvme_is_direction_write(op)`, `nvme_is_direction_read(op)`, `nvme_is_bidirection(op)`)

src/nvme/mi.c
jinliangw: Fixed and attached checkpatch.pl result in the comment above.

src/nvme/mi.c
```diff
@@ -419,6 +418,80 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+		req_hdr.dlen = cpu_to_le32(data_len);
+		/* Bit 0 set to 1 means DLEN contains a value */
+		req_hdr.flags = 0x1;
+	}
+
+	/* write data */
+	if((opcode & 0x3) == NVME_DATA_TFR_HOST_TO_CTRL) {
+		req.data = data;
+		req.data_len = data_len;
+	}
```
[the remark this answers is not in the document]
jinliangw: We don't support bidirectional commands. Please see your [comment](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/mi.c#L379):
> /* bidirectional not permitted (see DLEN definition) */

src/nvme/mi.c
jinliangw: Done.

src/nvme/mi.h
jinliangw: The nvme-cli subcommand `admin-passthru` doesn't support bidirectional data transfer for now(both in-band or OOB). So the discussion about bidirectional support is out of scope of this pull request.

src/nvme/mi.h
jinliangw: We may need to add new interface to handle bidirectional commands. Again, it is out of scope of this pull request.

jinliangw:
BTW, we have to use double `admin` in the interface name `nvme_mi_admin_admin_passthru` if we want to keep the wrapper style in nvme-cli:
```
-               err = nvme_admin_passthru(dev_fd(dev), cfg.opcode, cfg.flags,
+               err = nvme_cli_admin_passthru(dev, cfg.opcode, cfg.flags,
```

src/nvme/mi.c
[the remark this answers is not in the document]
jk-ozlabs: Good point, we can't do this over MI anyway!

OK, sounds good, and it's good to keep the API consistent, so I'm happy with this approach.

src/nvme/mi.c
```diff
@@ -419,6 +418,80 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+{
+	/* Input parameters flags, rsvd, metadata, metadata_len are not used */
+	struct nvme_mi_admin_resp_hdr resp_hdr;
+	struct nvme_mi_admin_req_hdr req_hdr;
+	struct nvme_mi_resp resp;
+	struct nvme_mi_req req;
+	int rc;
+	if(data_len > 4096) {
+		nvme_msg(ctrl->ep->root, LOG_ERR, "nvme_mi_admin_admin_passthru"
+			"does not supports data_len over 4096 bytes.\n");
```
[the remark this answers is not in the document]
jk-ozlabs: I have a slight preference for something like:

```c
    nvme_msg(ctrl->ep->root, LOG_ERR,
            "nvme_mi_admin_admin_passthru does not support data_len over 4096 bytes.\n");
```

\- where we keep the string intact. We'll be over 80 cols, but this reduces the line length as much as is sensible.

However, @igaw would have the final say about style matters...

(also, `does not supports` → `does not suppport`)

src/nvme/mi.h
jk-ozlabs: Yep, agreed. Let's keep this as-is.

jk-ozlabs:
> BTW, we have to use double `admin` in the interface name `nvme_mi_admin_admin_passthru` if we want to keep the wrapper style in nvme-cli:

Yes, good point. Let's keep it as you have proposed.

src/nvme/mi.c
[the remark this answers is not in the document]
jk-ozlabs: > (also, `does not supports` → `does not suppport`)

.. and I see you've fixed this in the latest push, thanks.

src/nvme/util.h
```diff
@@ -169,10 +169,10 @@ int nvme_get_feature_length(int fid, __u32 cdw11, __u32 *len);
...
  */
 enum nvme_data_tfr {
-	NVME_DATA_TFR_NO_DATA_TFR,
-	NVME_DATA_TFR_HOST_TO_CTRL,
-	NVME_DATA_TFR_CTRL_TO_HOST,
-	NVME_DATA_TFR_BIDIRECTIONAL
+	NVME_DATA_TFR_NO_DATA_TFR   = 0x0,
+	NVME_DATA_TFR_HOST_TO_CTRL  = 0x1,
+	NVME_DATA_TFR_CTRL_TO_HOST  = 0x2,
+	NVME_DATA_TFR_BIDIRECTIONAL = 0x3
```
igaw, line 175: I also asked to move this to the right header. Don't bother. I'll do this part myself now to avoid that release from later today has this part wrong.

As not all points have been addressed by this update I wont pull it.

jinliangw:
Sync with [linux-nvme:master](https://github.com/linux-nvme/libnvme) to include `enum nvme_data_tfr` change.
Hi @igaw, please let me know if you still want to add 3 helper functions(`nvme_is_direction_write(op)`, `nvme_is_direction_read(op)`, `nvme_is_bidirection(op)`).

jinliangw:
Hi @igaw, do you have more comments about this PR?

src/nvme/mi.c
igaw: This is difficult to read. Fixed it myself.

igaw:
Sorry, for the delay. I was finishing up the 1.2 release and decided to move this part to the 1.3 release. I've updated the change slightly and pushed it directly. thanks.

PR #506, opened 2022-10-25
Rename fallthrough to LIBNVME_FALLTHROUGH

drakedog2008:
`fallthrough` is a well-known name that may reused by other libraries. E.g. https://github.com/nlohmann/json/blob/a3e6e26dc83a726b292f5be0492fcc408663ce55/include/nlohmann/thirdparty/hedley/hedley.hpp#L1704

Rename to a local named macro which follows other name convention. E.g: https://github.com/boostorg/config/blob/902273e7380fedaa15f53ea15d0c3d947915fe4f/include/boost/config/detail/suffix.hpp#L1032


igaw:
Good point. Claiming the 'fallthrough' name is going to be a problem in the long run. Though I wonder if we can't make it this local in config.h and use meson to detect support from the compiler and just add the define if needed. Let my play with this.

PR #510, opened 2022-10-27
 Set log page offset for nvme_{mi}_get_log_telemetry_host

### LINKED ISSUE #472: Unused offset for nvme_mi_admin_get_log_telemetry_host
The offset is not used in both nvme_get_log_telemetry_host() and nvme_mi_admin_get_log_telemetry_host(). 

The log can be large especially for MI transportation. We need offset for partial log transfer.

igaw:
These function do not sett the passed in offset. Fix this.

Fixes #472


jk-ozlabs (approved):
Looks good for the MI part, and I assume the ioctl interface too, given the simplicity!

PR #511, opened 2022-10-27
build: Update min version of meson to 0.48

igaw:
python3.extension_module() is depending on gnu_symbol_visibility. So arguably, this has been from day 0.

Version 0.47 was released on Jul 02, 2018 and 0.48 on Sep 22,
2018. Even Debian oldstable ships 0.49.

Fixes: #496

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@
 #
 project(
     'libnvme', ['c'],
-    meson_version: '>= 0.47.0',
+    meson_version: '>= 0.48.0',
     version: '1.1',
     license: 'LGPL-2.1-or-later',
     default_options: [


eli-schwartz:
While it's perfectly fine to bump the minimum version of Meson, you are not actually setting this kwarg. Meson's python module does this internally, via https://github.com/mesonbuild/meson/pull/10369

The warning is actually bogus, but it's emitted in the backend not in the interpreter, so we do not actually know whether it was set by the python module or by the user's meson.build. Fun! Also way too annoying to fix, it will just fix itself as part of the pending rewrite in https://github.com/mesonbuild/meson/pull/10714.

PR #514, opened 2022-11-01
MI quirk implementation

jk-ozlabs:
We have found that some devices need small workarounds ("quirks") for proper MI-channel functionality.

This series introduces a little infrastructure for these changes, triggered by a probe of the Controller's Identify data. Based on that data, we have a new bitmask (`quirks`) in the endpoint structure:

```diff
@@ -195,6 +196,7 @@ struct nvme_mi_ep {
        bool controllers_scanned;
        unsigned int timeout;
        unsigned int mprt_max;
+       unsigned long quirks;
 };
```

which can be queried via a helper:

```c
static bool nvme_mi_ep_has_quirk(nvme_mi_ep_t ep, unsigned long quirk)
```

Using this, we introduce our first quirk: a delay between commands, required by a specific NVMe device:

```c
/* Set a minimum time between receiving a response from one command and
 * sending the next request. Some devices may ignore new commands sent too soon
 * after the previous request, so manually insert a delay
 */
#define NVME_QUIRK_MIN_INTER_COMMAND_TIME	(1 << 0)
```

The probing can be controlled via the MI API, or a new `LIBNVME_MI_PROBE_ENABLED` environment variable, documented in the `mi.rst` file.

### DIFF
--- a/doc/mi.rst
+++ b/doc/mi.rst
@@ -13,6 +13,17 @@ Most of the MI API is transport-agnostic, except for the endpoint constructor
 functions. Once an endpoint object (``nvme_mi_ep_t``) is created, the generic
 functions can be used to manage it.
 
+When endpoints are created (through one of the transport-specific functions,
+like ``nvme_mi_open_mctp()``), the endpoint hardware will be probed to
+see if any device-specific workarounds ("quirks") are required. This is
+implemented as an Identify Controller command, requesting a small amount of
+data on controller ID 0.
+
+To suppress this probe, the ``LIBNVME_MI_PROBE_ENABLED`` environment var can be
+set. Values of ``0``, ``false`` and ``disabled`` will disable the probe, and no
+quirks will be applied. Other values, or an unset environment variable, will
+enable the probe.
+
 MCTP Transport
 --------------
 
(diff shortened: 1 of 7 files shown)


jk-ozlabs:
Hm, we also need the .map change. Next version will be 1.3? or do you have a preference for signifying non-released symbols in the .map?

igaw:
There are a couple of fallouts from the nvme-cli 2.2 and libnvme 1.2 release. I am pondering of doing a 2.2.1 resp. a 1.2.1 release this week. So I am going to hold off functional changes until then. But yes, this will be then 1.3. material.

igaw:
So there aren't any new complains for the v2.2.1/v1.2 release so far. Let's wait a few more hours before starting the new devel cycle. I started to package up it for TW/openSUSE. Looks all good so far.

jk-ozlabs:
OK, great. No particular rush on the merging.

The somewhat-controversial part here is that we're now issuing a (single, small) NVMe-MI request on endpoint initialisation, rather than all commands being explicitly triggered by API calls. However, that's necessary to establish proper communication with some devices.

igaw:
Alright, let's merge it now and see if there are any complains on this new feature. Somehow it's sad to have to do add quirks in user space as well but I don't see any other way around.

PR #517, opened 2022-11-02
mi: Allow zero as a controller ID in nvme_mi_scan_ep

drakedog2008:
ctrl_id of 0 should be still considered as valid. Remove the condition check correspondingly.


drakedog2008:
@jk-ozlabs 

PTAL.

jk-ozlabs (approved):
Looks good - `0` is indeed a valid controller identifier according to the spec.

Thanks!

jk-ozlabs:
... this might warrant a better title though, so the bug fix can be found in future: I'd suggest:

``` mi: Allow zero as a controller ID in nvme_mi_scan_ep```

drakedog2008:
Thanks. Changed the title.

PR #520, opened 2022-11-04
mi: Fix status extraction from cdw3 field

jk-ozlabs:
Currently, we extract the NVMe status value from the top two bytes of cdw3 of the response.

However, the Status Code (SC) field actually starts at bit 17 (where the lsbit is bit 0), so we need to be shifting down by 17 instead. This makes the status align with the enum nvme_status_field values.

Add an explicit test for this, and clarify that the existing admin_err test is for the MI header status field.

Fixes: 108d696 ("mi: Introduce a helper for response status, unify values with ioctls")


jk-ozlabs:
Looks like a CI failure due to `wrapdb.mesonbuild.com` returning a `503`.

jk-ozlabs:
> Looks like a CI failure due to `wrapdb.mesonbuild.com` returning a `503`.

... aaaand we're good now.

PR #521, opened 2022-11-04
python: Release GIL on blocking calls

martin-belanger:
Release Python's GIL (global interpreter lock) when invoking blocking functions from multithreaded Python applications. This is to prevent long running threads to block all other threads.

Some I/O operations can run for a long time (seconds). For example, when connecting to a remote controller over fabrics, the timeout for a failure to connect is 3 seconds. If we do not release the GIL before calling a blocking function, all threads of a Python application could be blocked for several seconds. 

Ref: https://docs.python.org/3/c-api/init.html?highlight=gil#releasing-the-gil-from-extension-code

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -585,7 +585,11 @@ struct nvme_ns {
       connect_err = 1;
       return;
     }
+
+    Py_BEGIN_ALLOW_THREADS  /* Release Python GIL */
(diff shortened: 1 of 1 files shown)


igaw:
Just updated the comment style to `/* ... */` to stay consistent with (almost all the) rest.

PR #525, opened 2022-11-08
fix a check-after-dereference error in nvmf_discovery_log()

maurizio-lombardi:
logp pointer can't be NULL, because it has already been dereferenced, the function should check *logp instead.


martin-belanger (approved):

igaw:

PR #526, opened 2022-11-08
fix a static checker warning in examples/mi-mctp

maurizio-lombardi:
tmp is an "unsigned long" variable so it can't be less than zero.


igaw:

PR #527, opened 2022-11-08
fix a file handle leak in nvme_uuid_random()

maurizio-lombardi:


igaw:

Issue #524, opened 2022-11-08
test-mi: ../test/mi.c:333: test_scan_ctrl_list_cb: Assertion `hdr[11] == nvme_mi_dtyp_ctrl_list' failed


bdrung:
Running the test suite on s390x (big endian) fails on Debian and Ubuntu:
```
==================================== 2/6 =====================================
test:         mi
start time:   02:35:41
duration:     0.00s
result:       killed by signal 6 SIGABRT
command:      MALLOC_PERTURB_=24 /<<PKGBUILDDIR>>/obj-s390x-linux-gnu/test/test-mi
----------------------------------- stdout -----------------------------------
Running test endpoint_lifetime...  OK
Running test ctrl_lifetime...  OK
Running test read_mi_data...  OK
Running test transport_fail...  OK
Running test transport_describe...  OK
Running test scan_ctrl_list...
----------------------------------- stderr -----------------------------------
test-mi: ../test/mi.c:333: test_scan_ctrl_list_cb: Assertion `hdr[11] == nvme_mi_dtyp_ctrl_list' failed.
==============================================================================
```

Full log: https://buildd.debian.org/status/fetch.php?pkg=libnvme&arch=s390x&ver=1.2-1&stamp=1667874947&raw=0

This looks to an little vs big endian issue to me.

igaw:
@jk-ozlabs

jk-ozlabs:
Nice to have some BE coverage in the tests, thanks @bdrung . I'll get a BE setup going here and replicate; I'm pretty sure I've found the underlying issue but will also need to verify the fix.

igaw:
As previously mentioned we should be able to run the tests also in the cross builds with qemu.

jk-ozlabs:
Thanks! I wonder if we can enforce the current `__leXX` type annotations? that would have caught two of these cases.

igaw:
Maybe it could be done with `sparse` but I think we would need to figure out how to use it correctly:

```shell
[...]

PR #529, opened 2022-11-10
mi: endian fixes

jk-ozlabs:
We have a couple of endian issues in the mi code: one where we're not converting cdw0 for MI commands, and one where we're using the wrong byte length for an existing conversion. There is also an error in the test suite, where we should be converting the `nsze` field of a `nvme_id_ns`.

This change fixes those, so that the test suite passes on a BE platform (ppc64 in my case).

Fixes: https://github.com/linux-nvme/libnvme/issues/524


igaw:
Thanks a lot!

PR #533, opened 2022-11-16
Support static builds

igaw:
Fixes: #509

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -180,24 +180,7 @@ conf.set10(
     description: 'Is linux/mctp.h include-able?'
 )
 
-if meson.version().version_compare('>= 0.48')
-  has_fallthrough = cc.has_function_attribute('fallthrough')
-else
-  has_fallthrough = cc.compiles(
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -91,7 +91,20 @@ endif
...
+                 int main(void) {
+                     struct sd_bus *ret;
+                     return sd_bus_default(&ret);
+                 }
+               ''',
+              dependencies: libsystemd_dep,
+              name: 'Link check')
+       conf.set('CONFIG_LIBSYSTEMD', libsystemd_dep.found(), description: 'Is libsystemd(>219) available?')
+  else
+    libsystemd_dep = dependency('', required: false)
```
martin-belanger, line 105: meson also provides `declare_dependency()`, which allows declaring a dependency to a missing library.

igaw: Ah, didn't see that in the documentation. Looks a bit saner. Right let me fix that up.

Issue #537, opened 2022-11-18
Bad UUID size in nvmf_get_tel()


martin-belanger:
The recent changes to remove a dependency on libuuid has introduced the following bug. A pull request is on the way.

```
static __u32 nvmf_get_tel(const char *hostsymname)
{
	__u32 tel = sizeof(struct nvmf_ext_die);
	__u16 len;

	/* Host ID is mandatory */
	tel += nvmf_exat_size(NVME_UUID_LEN_STRING);  /* <<<<<<<<<<<<<< SHOULD BE NVME_UUID_LEN */

	/* Symbolic name is optional */
	len = hostsymname ? strlen(hostsymname) : 0;
	if (len)
		tel += nvmf_exat_size(len);

	return tel;
}
```

martin-belanger:
https://github.com/linux-nvme/libnvme/pull/538

PR #538, opened 2022-11-18
fabrics: Fix bad UUID size introduced in recent UUID changes

martin-belanger:
Fixes: https://github.com/linux-nvme/libnvme/issues/537

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1127,7 +1127,7 @@ static __u32 nvmf_get_tel(const char *hostsymname)
 	__u16 len;
 
 	/* Host ID is mandatory */
-	tel += nvmf_exat_size(NVME_UUID_LEN_STRING);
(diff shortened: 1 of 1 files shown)


martin-belanger:
@igaw - I don't know why, but some of the checks are failing to build subproject/openssl. 

That has nothing to do with the one line change I made.

igaw:
> @igaw - I don't know why, but some of the checks are failing to build subproject/openssl.
> 
> That has nothing to do with the one line change I made.

The static builds will build the OpenSSL library which produces warnings (new?). With `-werror` set this breaks the build. We have to remove the compile flag for the static builds.

igaw:
Rebased the commit on the current master with a build config change so that we can test if the builds are fixed now. While at also replaced the github reference in the commit message. The reason for this is that I want to avoid to have git commit message depending on an external service. We never know when we have to move the repo.

PR #539, opened 2022-11-18
mi: use correct log page offset

jinliangw:
mi: use correct log page offset

Problem:
The usage of `doff` in `nvme_mi_admin_get_log` is wrong.  It will cause below error:
```
Admin request:
 opcode: 0x02
 flags:  0x03
 ctrl:   0x0000
 nsid:   0x00000000
 cdw2:   0x00000000
 cdw3:   0x00000000
 cdw4:   0x00000000
 cdw5:   0x00000000
 doff:   0x00001000
 olen:   0x00001000
 cdw10:  0x03ff8007
 cdw11:  0x00000000
 cdw12:  0x00000000
 cdw13:  0x00000000
 cdw14:  0x00000000
 cdw15:  0x00000000
Admin response:
 Status: 0x04
 cdw0:   0x677830c7
 cdw1:   0x00000000
 cdw3:   0x00000000
 data_len: 0x00000000

NVMe-MI status: Invalid Parameter(0x4)
``` 
This GetLogPage Admin Command requests to read Log Page 0x07, `lpo` = 0, length = 4KB
However, the `doff` and `olen` requests NVMe controller skip the first 4KB, and return the next 4KB.
We should set `doff` to 0, and increase the `lpo` inside `cdw12` and `cdw13` to get the second 4KB data.
Please refer to the `Figure 121: NVMe Get Log Page Command Response Data Example` in "NVM-Express-Management-Interface-Specification-1.2b"

Changes:

1) Rename `tmp` to `xfer_len` for `nvme_mi_admin_get_log`;
2) Fix log page offset for `__nvme_mi_admin_get_log`:
The LPO(log page offset) of each chunk Admin Request message needs to
be updated.  Also, the DOFST in MI NVMe Admin Command Request is used to select the
window inside the response data of current command. So it's different from
LPO.
3) Fix `test_admin_get_log_split_cb`.

Tested with below commands:
```
./nvme telemetry-log --host-generate=1 --output-file=./host_gen1.data mctp:1,16:0
./nvme telemetry-log --host-generate=0 --output-file=./host_gen0.data mctp:1,16:0
./nvme get-log --log-id=0x07 --log-len=40960 -b mctp:1,16:0  > host.data
./nvme get-log --log-id=0x08 --log-len=40960 -b mctp:1,16:0  > ctrl.data
```

### DIFF
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -772,11 +772,19 @@ int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
 }
 
 /* retrieves a MCTP-messsage-sized chunk of log page data. offset and len are
- * specified within the args->data area */
+ * specified within the args->data area. The `offset` parameter is a relative
+ * offset to the args->lpo !
+ *
+ * What's more, we change the LPO of original command to chunk the request
+ * message into proper size which is allowed by MI interface. One reason is that
+ * this option seems to be supported better by devices.  For more information
+ * about this option, please check https://github.com/linux-nvme/libnvme/pull/539
+ * */
 static int __nvme_mi_admin_get_log(nvme_mi_ctrl_t ctrl,
 				   const struct nvme_get_log_args *args,
 				   off_t offset, size_t *lenp, bool final)
 {
+	__u64 log_page_offset = args->lpo + offset;
 	struct nvme_mi_admin_resp_hdr resp_hdr;
 	struct nvme_mi_admin_req_hdr req_hdr;
 	struct nvme_mi_resp resp;
@@ -808,17 +816,13 @@ static int __nvme_mi_admin_get_log(nvme_mi_ctrl_t ctrl,
 				    (args->lid & 0xff));
 	req_hdr.cdw11 = cpu_to_le32(args->lsi << 16 |
 				    ndw >> 16);
-	req_hdr.cdw12 = cpu_to_le32(args->lpo & 0xffffffff);
-	req_hdr.cdw13 = cpu_to_le32(args->lpo >> 32);
+	req_hdr.cdw12 = cpu_to_le32(log_page_offset & 0xffffffff);
+	req_hdr.cdw13 = cpu_to_le32(log_page_offset >> 32);
 	req_hdr.cdw14 = cpu_to_le32(args->csi << 24 |
 				    (args->ot ? 1 : 0) << 23 |
 				    args->uuidx);
 	req_hdr.flags = 0x1;
 	req_hdr.dlen = cpu_to_le32(len & 0xffffffff);
-	if (offset) {
-		req_hdr.flags |= 0x2;
-		req_hdr.doff = cpu_to_le32(offset);
-	}
 
 	nvme_mi_calc_req_mic(&req);
 
@@ -844,6 +848,11 @@ int nvme_mi_admin_get_log(nvme_mi_ctrl_t ctrl, struct nvme_get_log_args *args)
 	int rc = 0;
 
 	if (args->args_size < sizeof(*args)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	if (args->ot && (args->len > max_xfer_size)) {
 		errno = EINVAL;
 		return -1;
 	}
--- a/test/mi.c
+++ b/test/mi.c
@@ -1776,6 +1776,7 @@ static int test_admin_get_log_split_cb(struct nvme_mi_ep *ep,
 				       struct nvme_mi_resp *resp,
 				       void *data)
 {
+	uint32_t log_page_offset_lower;
 	struct log_data *ldata = data;
 	uint32_t len, off;
 	__u8 *rq_hdr;
@@ -1790,19 +1791,25 @@ static int test_admin_get_log_split_cb(struct nvme_mi_ep *ep,
 	off = rq_hdr[31] << 24 | rq_hdr[30] << 16 | rq_hdr[29] << 8 | rq_hdr[28];
 	len = rq_hdr[35] << 24 | rq_hdr[34] << 16 | rq_hdr[33] << 8 | rq_hdr[32];
 
+	/* From the MI message's Command Dword 12 */
+	log_page_offset_lower = rq_hdr[55] << 24 | rq_hdr[54] << 16 | rq_hdr[53] << 8 | rq_hdr[52];
+
 	/* we should have a full-sized start and middle, and a short end */
 	switch (ldata->n) {
 	case 0:
+		assert(log_page_offset_lower == 0);
 		assert(len == 4096);
 		assert(off == 0);
 		break;
 	case 1:
+		assert(log_page_offset_lower == 4096);
 		assert(len == 4096);
-		assert(off == 4096);
+		assert(off == 0);
 		break;
 	case 2:
+		assert(log_page_offset_lower == 8192);
 		assert(len == 4);
-		assert(off == 4096 * 2);
+		assert(off == 0);
 		break;
 	default:
 		assert(0);
@@ -1836,6 +1843,8 @@ static void test_admin_get_log_split(struct nvme_mi_ep *ep)
 	args.lid = 1;
 	args.log = buf;
 	args.len = sizeof(buf);
+	args.lpo = 0;
+	args.ot = false;
 
 	rc = nvme_mi_admin_get_log(ctrl, &args);
 


jinliangw:
Hi @igaw 

CI static build failed:
```
[273/746] Compiling C object subprojects/openssl-1.1.1l/libcrypto.a.p/crypto_ec_curve448_f_generic.c.o
```

igaw:
> Hi @igaw
> 
> CI static build failed:
> 
> ```
> [273/746] Compiling C object subprojects/openssl-1.1.1l/libcrypto.a.p/crypto_ec_curve448_f_generic.c.o
> FAILED: subprojects/openssl-1.1.1l/libcrypto.a.p/crypto_ec_curve448_f_generic.c.o
> ../subprojects/openssl-1.1.1l/crypto/ec/curve448/f_generic.c:21:27: error: argument 1 of type ‘uint8_t[56]’ {aka ‘unsigned char[56]’} with mismatched bound [-Werror=array-parameter=]
> ```

It looks like the compiler got updated(?) and now triggers build warnings whenever we use the fallbacks which trigger a build warning. Anyway, I've change the build configuration and it should build with the current master again. We can ignore the fallouts here or you rebase is to the current master.

jk-ozlabs:
I'm not clear on what the issue is with the current code, can you expand your commit message to cover that?

You mention:

> The LPO(log page offset) of each chunk Admin Request message needs to be updated

but why is this?

We have two offsets here: the lpo in the args structure, and the offset specified in the MI header. Your patch now combines those into one. If there's an issue with the code, we should fix it, but I'm not convinced this is the correct approach.

jinliangw:
Hi @jk-ozlabs, I updated the pull request comment to explain why this change is needed.

Also, for below two options:
1): increase the `args->lpo` at `nvme_mi_admin_get_log`, 
2): calculate the `__u64 log_page_offset = args->lpo + offset;` at `__nvme_mi_admin_get_log`
I prefer the option 2 as it avoids modifying input paramter `struct nvme_get_log_args *args`, an keep the original information unchanged.

jk-ozlabs:
Hi Jinliang,

Thanks for the updates! I still have questions though. You mentioned:

 > Problem:
> The usage of doff in nvme_mi_admin_get_log is wrong. It will cause below error:
> 
> ```
> Admin request:
>  opcode: 0x02
>  flags:  0x03
>  ctrl:   0x0000
>  nsid:   0x00000000
>  cdw2:   0x00000000
>  cdw3:   0x00000000
>  cdw4:   0x00000000
>  cdw5:   0x00000000
>  doff:   0x00001000
>  olen:   0x00001000
>  cdw10:  0x03ff8007
>  cdw11:  0x00000000
>  cdw12:  0x00000000
>  cdw13:  0x00000000
>  cdw14:  0x00000000
>  cdw15:  0x00000000
> Admin response:
>  Status: 0x04
>  cdw0:   0x677830c7
>  cdw1:   0x00000000
>  cdw3:   0x00000000
>  data_len: 0x00000000
> ```
>
> NVMe-MI status: Invalid Parameter(0x4)

This still doesn't explain _why_ it is wrong. The request data you have posted looks OK on an initial inspection, is something wrong there? Or is it that the drive just doesn't handle non-zero DOFST values (making it not spec-compliant?)? Or something else?

The original intention with this code was that:

 * the `LPO` value is used as-is from the libnvme API caller
 * the `DOFST` value is used for chunking the responses into MI-suitable sizes

You're changing that here, by putting everything into the `LPO` of the request. That may be acceptable, but we would need justification here about why.

Also: the rationale should eventually go into the commit message, not just the PR description - the PR gets separated from the code history, so it's nice to have the full text in the commit log.

jinliangw:
Hi @jk-ozlabs 

I updated the commit message with some extra explanation.
In short:
1) the `LPO` and `NUMD` are used to select a portion of `Get Log Page data` to produce `Get Log Page Command completion data`;
2) the `DOFST` and `DLEN` are used to select a portion of `Get Log Page Command completion data` for host to produce `Response data`
3) `DOFST < NUMD` and `DOFST + DLEN <= NUMD` 
image

jk-ozlabs:
Hi Jinliang,

Thanks, those updates are much more clear. I'm still not sure we want to disable the chunking via DOFST though. I would much rather we keep `lpo` exactly as-is from the original caller, and use DOFST to do the (libnvme-internal) chunking.

(So, relating to Figure 121 that you have included: the caller specifies the ranges on the left-hand side, while libnvme specifies the ranges on the right-hand side)

Your commit log states:

> The original usage of `doff` will create invalid Admin request message
> such as:
> ```
> Admin request:
>  opcode: 0x02
>  flags:  0x03
>  doff:   0x00001000
>  olen:   0x00001000
>  cdw10:  0x03ff8007
>  cdw12:  0x00000000
> ```

That looks like more I have something incorrect in the length calculations. What get_log arguments were used for this?

jinliangw:
Hi @jk-ozlabs 

1) There are two levels of chunk, here we only use first level of chunk, and won't use the second level of chunk `doff`...
2) The first level of chunk happens in `nvme_mi_admin_get_log` of `libnvme`.
3) We may need to  redesign current mi interface to utilize the second level of chunk. 

If you are still confused, we can discuss it at tomorrow's sync meeting.

jk-ozlabs:
Hi Jinliang,

> 1. There are two levels of chunk, here we only use first level of chunk, and won't use the second level of chunk `doff`...

That's what I have the issue with - we should be using the dofst here. The lpo should be left un-altered from the top-level libnvme call, if possible.

> 2. The first level of chunk happens in `nvme_mi_admin_get_log` of `libnvme`.

Yep, and that's available to the caller.

The dofst is not available to the libnvme caller, and we are _currently_ using it to implement the mi-size chunking.

However, you're proposing to remove that chunking and do *everything* at the lpo level. I would rather fix the doft usage if it is causing issues.
 
> If you are still confused, we can discuss it at tomorrow's sync meeting.

We should keep the discussion here, for the benefit of the folks who aren't on that call :)

Can you share the arguments used in `nvme_mi_admin_get_log_page`  that caused the request data that you posted?

jk-ozlabs:
Perhaps this is the fix you're aiming for?:

```diff
diff --git a/src/nvme/mi.c b/src/nvme/mi.c
index e9d86a7..c0c6a18 100644
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -795,7 +795,7 @@ static int __nvme_mi_admin_get_log(nvme_mi_ctrl_t ctrl,
                return -1;
        }
 
-       ndw = (len >> 2) - 1;
+       ndw = (args->len >> 2) - 1;
 
        nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id, nvme_admin_get_log_page);
        req_hdr.cdw1 = cpu_to_le32(args->nsid);
```

jinliangw:
Hi @jk-ozlabs 

>Can you share the arguments used in nvme_mi_admin_get_log_page that caused the request data that you posted?
```
./nvme get-log --log-id=0x07 --log-len=40960 -b mctp:1,16:0  > host.data
```
For these exposed log page API:
1.`nvme_mi_admin_get_log`
2.`nvme_mi_admin_get_log_create_telemetry_host`
3.`nvme_mi_admin_get_log_telemetry_host`
4. `nvme_mi_admin_get_log_telemetry_ctrl`
Caller(`nvme-cli`) won't do chunk by itself. Instead, caller replys on `libnvme` to perform chunking. 


>we should be using the dofst here. The lpo should be left un-altered from the top-level libnvme call, if possible.

I don't think we have to use `DOFST` to chunk log page data here.  Instead, we should use the well defined `lpo` Admin Command level chunk mechanism:
Reason 1: The in-band path [nvme_get_log_page](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/ioctl.c#L433) already uses `lpo` to chunk log page. It's better to keep in line with in-band implementation.
Reason 2: `lpo` is 64bits,  log page length is 34bits(`NUMDU` + `NUMDL` is 32bits). `DOFST` and `DLEN` are only 32 bits. Under extremely conner case, 32bits `DOFST` + `DLEN` cannot be used to represent the full range of `lpo` and log page length.

>you're proposing to remove that chunking and do everything at the lpo level.

Yes, I'm proposing to only use first level chunk(`lpo`) for get log page implement, and only use `DOFST`/`DLEN` chunking for other commands which don't have NVMe Admin command level chunking mechanism(for example, [nvme_mi_admin_identify_partial](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/mi.c#L717))

jk-ozlabs:
Hi Jinliang,

> ```
> ./nvme get-log --log-id=0x07 --log-len=40960 -b mctp:1,16:0  > host.data
> ```

ok, thanks!

> 
> For these exposed log page API: 1.`nvme_mi_admin_get_log` 2.`nvme_mi_admin_get_log_create_telemetry_host` 3.`nvme_mi_admin_get_log_telemetry_host` 4. `nvme_mi_admin_get_log_telemetry_ctrl` Caller(`nvme-cli`) won't do chunk by itself. Instead, caller replys on `libnvme` to perform chunking.

Yes, correct.

> 
> > we should be using the dofst here. The lpo should be left un-altered from the top-level libnvme call, if possible.
> 
> I don't think we have to use `DOFST` to chunk log page data here. Instead, we should use the well defined `lpo` Admin Command level chunk mechanism:

The dofst mechanism is also well-defined.

> Reason 1: The in-band path [nvme_get_log_page](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/ioctl.c#L433) already uses `lpo` to chunk log page.

Sure, but that's the only mechanism it has. This is exactly what the DOFST/DLEN fields are for.

> Reason 2: `lpo` is 64bits, log page length is 34bits(`NUMDU` + `NUMDL` is 32bits). `DOFST` and `DLEN` are only 32 bits. Under extremely conner case, 32bits `DOFST` + `DLEN` cannot be used to represent the full range of `lpo` and log page length.

Are there any log pages that are above 4GB in size?

Keep in mind the scales that we're dealing with the MI interface here - a 4GB log page would take 34 *hours* to transfer over a typical connection. If an API caller is requesting that sort of data, it is going to be doing its own chunking.

> Yes, I'm proposing to only use first level chunk(`lpo`) for get log page implement, and only use `DOFST`/`DLEN` chunking for other commands which don't have NVMe Admin command level chunking mechanism

... and now we need two different chunking mechanisms, depending on the command type. I don't see any benefit for that?

jinliangw:
Hi @jk-ozlabs 

>The dofst mechanism is also well-defined.

Absolutely not. For me, the `DOFST` mechanism is the most confusing/error-prone part of NVM-Express-Management-Interface-Specification.

>Sure, but that's the only mechanism it has.

In-band uses `lpo`, why nvme-mi cannot use `lpo`? Using DOFST/DLEN fields for get log page implement only adds complicity without benefit.

>Are there any log pages that are above 4GB in size?
Keep in mind the scales that we're dealing with the MI interface here - a 4GB log page would take 34 hours to transfer over a typical connection. If an API caller is requesting that sort of data, it is going to be doing its own chunking.

Even it's not practical case in 2022, it may become valid use case in future. (log page size is increasing, and SMBUS speed will increase). And technically`./nvme get-log --log-len=4GB` is still valid use case no matter how much time it needs to complete.  So why not support it right now?

>... and now we need two different chunking mechanisms, depending on the command type. I don't see any benefit for that?

The benefit are that:
1) we remove one level of chunk, which reduce complexity, and make it simpler to model and debug;
2) we use the well defined, durable and stable `lpo` mechanism over newly defined, error-prone  DOFST/DLEN mechanism;
3) we avoid potential 32bit limitation of DOFST/DLEN mechanism;
4) we keep in-band and out-of-band implement unified or similar.

igaw:
> And technically./nvme get-log --log-len=4GB is still valid use case no matter how much time it needs to complete. So why not support it right now?

I might miss something but it really sounds like over engineering to support fetching 4GB log. What is the use case here?

jk-ozlabs:
Hi Jinliang,

> In-band uses `lpo`, why nvme-mi cannot use `lpo`? Using DOFST/DLEN fields for get log page implement only adds complicity without benefit.

Because we can keep the layering consistent if we use DOFST/DLEN.

The design so far has been that:

 * the user provides the actual parameters for the admin command. For Get Log Page, this includes the LPO and NUMD fields.
 * if any chunking is necessary, that's handled as part of the lower-layer: the libnvme-mi transport. Since the DOFST/DLEN fields are defined as part of the MI transport, we're keeping them clearly distinct from the actual Admin command

By using DOFST/DLEN, we will issue the command that the API caller has issued; just repeated with different DOFST/DLEN to suit the transport requirements. If we don't do this, we have to adjust the parameters that the user has supplied, which is a violation of this layering.

This allows us to keep a consistent chunking implementation across all MI commands. Otherwise we would have the  LPO/NUMD chunking just for Get Log page, and DOFST/DLEN chunking for all other admin commands. I'd much rather use the consistent approach across all MI commands, as this is exactly what it is specified for.

The in-band chunking implementation in libnvme for Get Log Page was introduced as a work-around for controllers that can't handle >4kB log page data; I don't think that's a good basis for replicating this in the out-of-band transport.

> >  The dofst mechanism is also well-defined.
>
> Absolutely not.

It is well defined as part of the specification. I do agree that it's difficult having two offsets and lengths when issuing a Get Log Page (ie., having its own offset & length), but the Figure 121 diagram that you have posted does make that quite clear.

Again, the benefit of using the DOFT/DLEN for the Get Log Page chunking is that we can use this across all commands, and not require two separate mechanisms for the one transport.

jinliangw:
Hi @jk-ozlabs 

Sorry that I have to point out that **the intent to use  DOFST/DLEN as chunk mechanism** is inappropriate and even wrong from the beginning.

According to NVM-Express-Management-Interface-Specification-1.2b "6.1 Request and Response Data":
>Some NVMe Admin Commands **specify an offset and length which shall be applied first** to create the NVMe
Admin Command completion data. Then DOFST and DLEN shall be applied to that Admin Command
completion data which may **further reduce the amount of Response Data**. Figure 121 provides an example
for the Get Log Page command.

The DOFST and DLEN is designed to reduce the amount of response data transfered over nvme-mi. It is not intended for general chunk mechanism.
Compared with 1.1d spec, 2.0b spec adds this paragraph the new Figure 121 specially to emphasis the Admin command level chunk mechanism should be used **first**.

**Most importantly**, we need to redesign the `libnvme` mi interface to expose DOFST and DLEN for caller(nvme deamon) at some time in future, so it can be used to select a special field of data for those commands which don't have Admin command offset concept. For example, we may need the SSD FW to :
1) only return  Serial Number(bytes 23:04) of identify controller
2) only return  one telemetry field inside 512 block size of Telemetry Host-Initiated (Log Identifier 07h) or Telemetry Controller-Initiated (Log Identifier 08h).

This is the ideal purpose of usage of DOFST and DLEN - **select/filter an interesting portion of data from Admin command response data**.  Using it as a general chunk mechanism inside `libnmve` implementation is against intuition and against layering principle, and make it complex and hard for further optimization.

igaw:
> > I might miss something but it really sounds like over engineering to support fetching 4GB log. What is the use case here?
> 
> This is not over engineering as this pull request is not meant to support fetching 4GB log.

Okay, I got the impression that's the driving problem. Anyway, I let you guys figure out the solution :)

>  I would say it's just a side- benefit to choose `lpo` level chunking mechanisms so we can keep consistency with NVMe spec definition: `lpo` is 64bits, log page length is 34bits(`NUMDU + NUMDL` is 32bits).

One thing we could do is to do ask the spec folks to explain those parameters if there is anything unclear here.

jinliangw:
Hi @igaw and @jk-ozlabs 

>One thing we could do is to do ask the spec folks to explain those parameters if there is anything unclear here.

Here are thoughts from Myron.Loewen@solidigm.com, a member of managementif@nvmexpress.org:
>I agree with your assessment of DOFST vs LPO in the links and that DOFST is applied after LPO. The reason we put DOFST in NVMe-MI was exactly to reduce traffic on our limited OOB bandwidth and we needed a mechanism consistent for all OOB traffic. I think that the LPO method is the way if you want to chunk writes, or read an entire large field in smaller chunks, or read small pieces of a large data structures since the “effort” and buffering requirements can be reduced earlier in the storage controller’s layered implementation. On commands without chunking or where the smaller size of the admin data transfers impacts storage controller state then use DOFST.

Also, I forwarded the whole email thread to @jk-ozlabs for better context.
Two key points:
1) DOFST in NVMe-MI was exactly to reduce traffic the limited OOB bandwidth.
2) (By using LPO chunk) the “effort” and buffering requirements can be reduced earlier in the storage controller’s layered implementation.

jk-ozlabs:
Hi Jinliang,

> According to NVM-Express-Management-Interface-Specification-1.2b "6.1 Request and Response Data":
> 
> > Some NVMe Admin Commands **specify an offset and length which shall be applied first** to create the NVMe
> > Admin Command completion data. Then DOFST and DLEN shall be applied to that Admin Command
> > completion data which may **further reduce the amount of Response Data**. Figure 121 provides an example
> > for the Get Log Page command.

Yes, and that's exactly what we are currently doing:

 * the caller first specifies the lpo and numd fields to specify which completion data is to be returned
 * libnvme then applies the DOFST and DLEN parameters to fit within the restrictions of the MI transport (ie, a 4096-byte maximum data length)

As a thought experiment: can you imagine trying to do this the other way around? ie, exposing a DOFST/DLEN-like facility to the API caller, then using lpo & numd for MI-specific chunking? that would be extremely awkward, because the lpo & numd is applied first, and then the DOFST/DLEN would need to be removed and re-calculated relative to the lpo chunks.

> It is not intended for general chunk mechanism.

That is your interpretation, not a statement from the specification. The caller may *also* be doing its own chunking with lpo & numd, but that is (as it should be) abstracted from the libnvme transport.

I do agree that the benefit from using lpo & numd (as Myron has mentioned) is that it allows the controller to limit the response size (and effort) "earlier" in its internal NVMe-MI implementation, and indeed, this may help us to avoid firmware bugs. If we're hitting that here, then yes: we should go with the lpo + numd approach to work around those bugs.

(I wouldn't mind investigating what's happening there though, could you send a `.pcap`?)

We may still need to do chunking with DOFST & DLEN for other commands that have > 4096 byte responses, and don't have their own offset & length parameters, so we're going to need both implementations.

Some other notes:

> Most importantly, we need to redesign the libnvme mi interface to expose DOFST and DLEN for caller(nvme deamon) at 
> some time in future, so it can be used to select a special field of data for those commands which don't have Admin
>  command offset concept

We already have this - for example (as you've mentioned earlier), `nvme_mi_admin_identify_partial()`. Ideally, we would have the same facilities available across the command API, rather than per-command. We'll keep this in mind for the API-breaking changes planned :)

[there is a little bit of complexity around that, as we're attempting to unify command access API across both in-band and out-of-band, so we need to be careful about which facilities we expose through that - ie., DLEN and DOFST not being available in-band)]

> only return Serial Number(bytes 23:04) of identify controller

yep, we do something very similar in the new probe code: perform an identify just for the vendor, model, and firmware revision information.

> only return one telemetry field inside 512 block size of Telemetry Host-Initiated (Log Identifier 07h) or Telemetry Controller-Initiated (Log Identifier 08h).

the caller (which knows which response data it needs) can do this already with lpo & numd, right?

(there is some complexity when returning partial data into a typed structure, but I'm sure we can work that out)

> 1. It causes bug in libnvme

Which we have just fixed :)

> 2. It's easier to hit bug in SSD FW

Yep, this is the big one - if we're hitting an actual bug, then we *have* to use lpo + numd.

jinliangw:
Hi @jk-ozlabs 
> Yes, and that's exactly what we are currently doing:
>- the caller first specifies the lpo and numd fields to specify which completion data is to be returned
>- libnvme then applies the DOFST and DLEN parameters to fit within the restrictions of the MI transport (ie, a 4096-byte >maximum data length)

Not 100% true. The current state is that:
1. Caller does specify log page offset and log page length(not `numd` fields, and even cover the whole log page); calller(nvme-cli) doesn't know the concept of chunking, it relies on `libnvme` to implement proper chunking.
2. `libnvme` uses the DOFST and DLEN parameters do chunking. -> which I don't think it's a good option.

>As a thought experiment: can you imagine trying to do this the other way around? ie, exposing a DOFST/DLEN-like facility to the API caller,

My opinion is that we will only expose DOFST/DLEN to a very small amount of APIs to improve performance - generally supporting DOFST/DLEN for all MI API is not necessary.  For now I don't see the need to expose DOFST/DLEN for get log pageAPI. So, LPO chunking only is enough for getting log page for now.

>that would be extremely awkward, because the lpo & numd is applied first, and then the DOFST/DLEN would need to be removed and re-calculated relative to the lpo chunks.

If we really really have to support a special read log page API such as `get_log_page(offset, length, dofset, dlen)` in future, the `dofset` and `dlen` would be within the range of `offset` and `length`. It may be a little complex to chunk by `lpo` and handle `dofset` and `dlen` correctly, but it's still possible.  

>> It is not intended for general chunk mechanism. 

> That is your interpretation, not a statement from the specification.

This is my interpretation, and I guess Myron also support it to some extent. The introducing of DOFST is to reduce data xfered on SMBus, not for general chunk mechanism.  
As you can see from the email thread, he is suggesting us to add it into ECN to give more explanation about the usage of DOFST. I will draft the ECN so we can make it more clear during the review and discussion of the ECN.

>(I wouldn't mind investigating what's happening there though, could you send a .pcap?)

I will open a bug for our controller team and collect pcap for the bug and send it to you. (Please remind me if I forget it)

>[there is a little bit of complexity around that, as we're attempting to unify command access API across both in-band and out-of-band, so we need to be careful about which facilities we expose through that - ie., DLEN and DOFST not being available in-band)]

As I said above: "we will only expose DOFST/DLEN to a very small amount of APIs to improve performance - generally supporting DOFST/DLEN for all MI API is not necessary. " This small amount of APIs won't have in-band counterparts.

>>only return one telemetry field inside 512 block size of Telemetry Host-Initiated (Log Identifier 07h) or Telemetry Controller-Initiated (Log Identifier 08h).

>the caller (which knows which response data it needs) can do this already with lpo & numd, right?

No, there is a limitation for `lpo`/`numd` regarding Log page 0x07 and log page 0x08: they must be 512 bytes aligned.
`DOFST` and `DLEN` is the perfect candidate for **selecting** a portion of data that is not 512 bytes aligned.  However, there is no strong use case to request this feature for now.


In summary:
Use case 1: Use Admin command level **chunking** mechanism if it exists; 
Use case 2: If it doesn't exist, turn to `DOFST` and `DLEN`. I think only security send/ security receive commands may need this use case. Will investigate whether we really need it.
Use case 3: Use `DOFST` and `DLEN` to **select** the portion of data which needs to be returned host. Again, this only apply to a limited set of API for performance improvement.

Even the discussion we have here is super complicated, the actually code change(like this pull request) may be minor and more clear than the discussion itself.

jk-ozlabs:
Hi Jinliang,

>  1. Caller does specify log page offset and log page length(not `numd` fields, and even cover the whole log page); calller(nvme-cli) doesn't know the concept of chunking, it relies on `libnvme` to implement proper chunking.

Yes, and that's the core of my objection to using lpo for everything: it crosses the boundary of what's part of the command, and what is part of the transport.

The caller has access to the lpo & numd fields (even though it's not direct, it still specifies the values used in numd & lpo through the libnvme API). The caller may be using these for any purpose: either for selecting a subset of data, or to implement its own chunking.

For a MI transport, libnvme then needs to enforce the transport-specific limit of 4096 bytes. We have an effective¹ mechanism, designed exactly at the MI transport level, for working with the transport-specific communication parameters, without having to alter the caller-provided lpo values.

¹: effective until bugs in the drive firmware make it not-effective, of course :)

By using DOFST and DLEN, we keep the "layering" intact: transport-specific behaviour can stay in the (MI) transport layer, and we can keep the Admin Command parameters unaltered at the command layer.

**However**: if we're forced to work only with lpo because of drive firmware bugs, or even if it makes things more efficient on the drive implementation, then that's a reasonable justification for using the lpo approach.

This isn't a matter of what the standard provides (because the standard provides both mechanisms - your diagram is a nice illustration of both working in unison), it's a matter of the API abstractions provided by libnvme, and how those abstractions map to the underlying communication. The NVMe-MI spec (rightly so) does not dictate those API abstractions.

I'm happy to proceed with your approach, I'd just like to figure out the semantics of what's failing at the moment (with that `len` fix in place), so we can establish a full rationale for doing so. Hence the request for your pcap.

But yes, this has been a lot of discussion on selecting which of two values needs to carry an offset :)

jk-ozlabs:
Jinliang has sent me a pcap, and I've been doing some experimentation here with a couple of different drives, using large Get Log page requests.

From that testing: it looks like some hardware cannot handle *any* Admin Command response data being larger than 4096 bytes, even when "cropped" to a 4096-byte section using DLEN. The drive I have reports a much larger supported response size in `MDTS`, so it's not like we're violating the spec with that request.

In the implementations that I have tested (which don't handle the >4096 byte response), I see an error response at least, and it looks like Jinlian's hardware just fails to send any response.

This means have no choice but to implement the libnvme-internal chunking with numd & lpo. Hopefully there are no other commands without their own len/offset fields that also return a >4096 byte response!

Jinliang: so, let's go ahead with this change, but I have a few requests:

 - we'll need a fairly substantial comment on how & why we're setting the `lpo` field there
 - we'll need to reject any Get Log Page which requires chunking, and also has `OT` set in cdw14, as we cannot adjust these offsets. I would suggest an `EINVAL` return.
 - if you want to do the `tmp` variable rename, please split that out into a separate patch, but I would suggest a different name (with your current proposal, we would have `xfer_len` and `xfer_size`, which may be confusing?)

jinliangw:
Hi @jk-ozlabs 

Thank you for your tests and the suggestions.

>- we'll need a fairly substantial comment on how & why we're setting the lpo field there

Added more comment for `__nvme_mi_admin_get_log`

>- we'll need to reject any Get Log Page which requires chunking, and also has OT set in cdw14, as we cannot adjust these offsets. I would suggest an EINVAL return.

I will handle `OT` in subsequent pull request. It looks like:
```
if (args->ot && (args->len > 4096)) {
    errno = EINVAL;
    return -1;
}
```

>- if you want to do the tmp variable rename, please split that out into a separate patch, but I would suggest a different name (with your current proposal, we would have xfer_len and xfer_size, which may be confusing?)

My original initiative is that: `length` is dynamic and variable, `size` is more like constant.
Anyway, I updated it in another pull request: https://github.com/linux-nvme/libnvme/pull/545

src/nvme/mi.c
```diff
@@ -782,6 +789,7 @@ static int __nvme_mi_admin_get_log(nvme_mi_ctrl_t ctrl,
 	struct nvme_mi_resp resp;
 	struct nvme_mi_req req;
 	size_t len;
+	__u64 log_page_offset = args->lpo + offset;
```
jk-ozlabs, line 792: Reverse-christmas-tree for these.

test/mi.c
```diff
@@ -1798,11 +1798,11 @@ static int test_admin_get_log_split_cb(struct nvme_mi_ep *ep,
 		break;
 	case 1:
 		assert(len == 4096);
-		assert(off == 4096);
+		assert(off == 0);
 		break;
 	case 2:
 		assert(len == 4);
-		assert(off == 4096 * 2);
+		assert(off == 0);
```
jk-ozlabs, line 1805: We should probably check the `lpo` field here too, right?

jk-ozlabs (changes requested):
Please also include the `ot` check in this change too - otherwise we are introducing potentially invalid behaviour between this and the (future) invalidation of calling with `ot` set.

test/mi.c
jinliangw: Done

src/nvme/mi.c
jinliangw: Done

jinliangw:
Update code according to the suggestion. 
Hi @igaw please help to squash my commits when you perform the merge.

jinliangw:
Hi @jk-ozlabs 
Any update about this pull request?

jk-ozlabs:
> Hi @jk-ozlabs Any update about this pull request?

No updates from me, are you waiting on something from my side?

In case it's a review: the changes you've pushed 18 hours ago (ie., at 666221837) look good to me.

jinliangw:
Thanks @jk-ozlabs !

Hi @igaw please help to merge this pull request.

igaw:
I was waiting for @jk-ozlabs okay. Good, so everyone is happy now.

jk-ozlabs:
👍

we should probably add the same error return for the in-band Get Log Page if `ot` is set. I can do that if you like, @igaw.

igaw:
> +1
> 
> we should probably add the same error return for the in-band Get Log Page if `ot` is set. I can do that if you like, @igaw.

sound reasonable. yes please :)

PR #542, opened 2022-11-23
MI: switch from libsdbus to libdbus

jk-ozlabs:
libsystemd is not generally available as a static library, and we would like to produce a static nvme-cli. This change switches to libdbus instead.

This requires slightly more boilerplate code for the dbus marshalling/unmarshalling as part of the MCTP endpoint scan, but means we have a simpler upstream lib dependency.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/1734

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -40,11 +40,11 @@ executable(
     include_directories: [incdir, internal_incdir]
 )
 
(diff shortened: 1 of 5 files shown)


igaw:
Thanks a lot!

PR #545, opened 2022-11-25
mi: rename tmp variable inside nvme_mi_admin_get_log

jinliangw:
rename the `tmp` variable into `xfered_size`.
It is hard to figure out a name which can imply that this variable is used as both input and output parameter, so I choose `xfered_size` to emphasize the output functionality.


jk-ozlabs (approved):
LGTM, thanks!

PR #547, opened 2022-11-29
ioctl: fix log page offset

jinliangw:
The original `lpo` info is lost in `nvme_get_log_page`. 
Save it into a local variable and append `offset` after it.


igaw:
Thanks. I've just renamed the local variable to match the code style.

PR #549, opened 2022-12-06
Pr/bug fix

drakedog2008:
Fixed some bugs in mi.c

### DIFF
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -1055,7 +1055,10 @@ int nvme_mi_admin_get_features(nvme_mi_ctrl_t ctrl,
 	int rc;
 
 	if (args->args_size < sizeof(*args))
-		return -EINVAL;
+	{
+		errno = EINVAL;
+		return -1;
+	}
 
 	nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id,
 			       nvme_admin_get_features);
@@ -1094,7 +1097,10 @@ int nvme_mi_admin_set_features(nvme_mi_ctrl_t ctrl,
 	int rc;
 
 	if (args->args_size < sizeof(*args))
-		return -EINVAL;
+	{
+		errno = EINVAL;
+		return -1;
+	}
 
 	nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id,
 			       nvme_admin_set_features);
@@ -1138,7 +1144,10 @@ int nvme_mi_admin_ns_mgmt(nvme_mi_ctrl_t ctrl,
 	int rc;
 
 	if (args->args_size < sizeof(*args))
-		return -EINVAL;
+	{
+		errno = EINVAL;
+		return -1;
+	}
 
 	nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id,
 			       nvme_admin_ns_mgmt);
@@ -1174,7 +1183,10 @@ int nvme_mi_admin_ns_attach(nvme_mi_ctrl_t ctrl,
 	int rc;
 
 	if (args->args_size < sizeof(*args))
-		return -EINVAL;
+	{
+		errno = EINVAL;
+		return -1;
+	}
 
 	nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id,
 			       nvme_admin_ns_attach);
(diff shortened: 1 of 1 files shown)


drakedog2008:
@jk-ozlabs 

PTAL.

src/nvme/mi.c
```diff
@@ -582,12 +582,39 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
 	}
 
 	/* bidirectional not permitted (see DLEN definition) */
-	if (req_data_size && *resp_data_size) {
+	int direction = admin_req->opcode & 0x3;
```
jk-ozlabs, line 585: declarations typically at the top of the function.

src/nvme/mi.c
```diff
@@ -582,12 +582,39 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+	{
+		nvme_msg(ctrl->ep->root, LOG_ERR,
+				 "nvme_mi_admin_xfer invalid offset value\n");
+		errno = EINVAL;
+		return -1;
+	}
+	else if (direction == NVME_DATA_TFR_NO_DATA_TFR && req_data_size && *resp_data_size)
+	{
+		nvme_msg(ctrl->ep->root, LOG_ERR,
+				 "nvme_mi_admin_xfer doesn't support data on NONE_DATA_TRANSFER commands\n");
```
jk-ozlabs, line 617: You have these conditionals on the direction here, and then another set below to set up the request/response fields.

I would suggest simplifying this logic by having a single `switch (direction) { ... }` statement, and checking for the errors, plus setting up the req & resp fields in each `case`.

But if you don't decide to use a `switch`: since all of the statements return, you don't need to make this a chain of `else if`; each check can be its own independent statement. This makes it slightly easier to reason about the logic.

src/nvme/mi.c
```diff
@@ -611,9 +638,42 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+		}
+		admin_req->dlen = cpu_to_le32(resp.data_len & 0xffffffff);
+		admin_req->doff = cpu_to_le32(resp_data_offset & 0xffffffff);
+	}
+	else
+	{
+		admin_req->flags = 0;
+		admin_req->dlen = 0;
+		admin_req->doff = 0;
+	}
```
jk-ozlabs, line 676: I'm not sure how much of this change is required; it's perfectly valid to have the flags at `0x3`, and the offsets still zero. In fact, the flags are deprecated in NVMe-MI 1.2, and the dlen/dofst values are always considered valid, so we'll need to do that anyway.

src/nvme/mi.c
```diff
@@ -582,12 +582,39 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+		return -1;
+	}
+	else if (resp_data_offset && (direction != NVME_DATA_TFR_CTRL_TO_HOST || !*resp_data_size))
+	{
+		nvme_msg(ctrl->ep->root, LOG_ERR,
+				 "nvme_mi_admin_xfer invalid offset value\n");
+		errno = EINVAL;
+		return -1;
+	}
+	else if (direction == NVME_DATA_TFR_NO_DATA_TFR && req_data_size && *resp_data_size)
```
jk-ozlabs, line 614: should be `direction == NVME_DATA_TFR_NO_DATA_TFR && (req_data_size || *resp_data_size)`, I think?

src/nvme/mi.c
```diff
@@ -611,9 +638,42 @@ int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
...
+	{
+		if (req.data_len)
+		{
+			admin_req->flags = 0x1;
+		}
+		else
+		{
+			admin_req->flags = 0;
+		}
+		admin_req->dlen = cpu_to_le32(resp.data_len & 0xffffffff);
```
jk-ozlabs, line 651: looks like this should be `req.data_len` ?

jk-ozlabs (changes requested):
Looks good overall, just a few comments inline.

In general though: you'll want to match the coding style of the existing project; in this case, keep the opening braces on the same lines as the `if` statements, and put all of the variable declarations at the top of the function.

igaw:
Closing due to inactivity.

PR #552, opened 2023-01-04
types: add fdp data structures and helpers

birkelund:
Add required data structures and helper functions for TP4146 ("Flexible Data Placement").

These data structures and helpers are used by the associated [changes for nvme-cli](https://github.com/linux-nvme/nvme-cli/pull/1756), to be PR'ed separately.


igaw:
I did some changes:

- drop packed attribute
- reorder struct args members
- move new functions to next version linker section

PR #555, opened 2023-01-13
build: Add support to build libdbus as fallback

igaw:
The dbus project supports to be consumed via Meson subproject mechanism.  Update the build configuration accordingly.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -89,8 +89,24 @@ if openssl_dep.found()
            description: 'OpenSSL/LibreSSL API version @0@'.format(api_version))
 endif
 
-# Check for libdus availability. Optional, only required for MCTP dbus scan
-libdbus_dep = dependency('dbus-1', required: false)
+if get_option('libdbus').disabled()
+    libdbus_dep = dependency('', required: false)
+else
+    # Check for libdus availability. Optional, only required for MCTP dbus scan
+    libdbus_dep = dependency(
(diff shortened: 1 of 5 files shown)


meson.build
```diff
@@ -89,8 +89,24 @@ if openssl_dep.found()
            description: 'OpenSSL/LibreSSL API version @0@'.format(api_version))
 endif
 
-# Check for libdus availability. Optional, only required for MCTP dbus scan
-libdbus_dep = dependency('dbus-1', required: false)
+if get_option('libdbus').disabled()
+    libdbus_dep = dependency('', required: false)
```
martin-belanger, line 93: Could we use `declare_dependency` instead?

igaw: Sounds like a good idea according the docs. We have some more of those. I add a separate commit to change all in one go

igaw: Nope, doesn't work. The `declare_dependecy` dependency object will return true `.found()`.

martin-belanger: Good to know...

igaw: And one thing I don't understand is that the documentation talks about internal build but we have to use declare_dependency to make the dependency being announced to be used in a wrap... I am too stupid too understand all these things :)

martin-belanger: I read the documentation and did not understand as well... :smirk:

PR #558, opened 2023-01-19
TP4146 updates

birkelund:
A couple of fixes for the TP4146 Flexible Data Placement support.

### DIFF
--- a/src/nvme/ioctl.h
+++ b/src/nvme/ioctl.h
@@ -1672,13 +1672,12 @@ static inline int nvme_get_log_fdp_configurations(int fd, __u16 egid,
 /**
  * nvme_get_log_reclaim_unit_handle_usage() - Get reclaim unit handle usage
  * @fd:		File descriptor of nvme device
- * @nsid:	Namespace identifier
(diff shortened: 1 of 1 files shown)


src/libnvme.map
```diff
@@ -6,6 +6,10 @@ LIBNVME_1_3 {
 		nvme_io_mgmt_send;
 		nvme_fdp_reclaim_unit_handle_status;
 		nvme_fdp_reclaim_unit_handle_update;
+		nvme_get_log_fdp_configurations;
+		nvme_get_log_fdp_events;
+		nvme_get_log_fdp_stats;
+		nvme_get_log_reclaim_unit_handle_usage;
```
igaw, line 12: Inline function don't have to be listed in the linker script. While at it, please sort the whole section :)

birkelund: Ah, I see.

In that case, we can drop the commit, or replace it by one removing `nvme_fdp_reclaim_unit_handle_status,update` as well since those are also inline.

igaw: All good, I got a PR which does the cleanup pending :)

PR #562, opened 2023-01-24
fabrics: Always sanitize discovery log entries

igaw:
Do not just sanitize the traddr and trsvcid when unsing the auto discovery/connect path, instead also do it for all users of nvmf_get_discovery_log{_args}() per default.

See also #561

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -683,8 +683,6 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 		switch (e->adrfam) {
 		case NVMF_ADDR_FAMILY_IP4:
 		case NVMF_ADDR_FAMILY_IP6:
-			strchomp(e->traddr, NVMF_TRADDR_SIZE - 1);
-			strchomp(e->trsvcid, NVMF_TRSVCID_SIZE - 1);
 			traddr = e->traddr;
 			trsvcid = e->trsvcid;
 			break;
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -677,14 +705,14 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 	nvme_ctrl_t c;
 	int ret;
 
+	strchomp_discovery_log_entry(e);
```
igaw, line 708: I think we can even drop this one here. The passed entry should already be sanitized.

src/nvme/fabrics.c
```diff
@@ -903,8 +931,16 @@ int nvmf_get_discovery_log(nvme_ctrl_t c, struct nvmf_discovery_log **logp,
...
 		.ot = false,
 	};
-	*logp = nvme_discovery_log(c, &args, max_retries);
-	return *logp ? 0 : -1;
+
+	log = nvme_discovery_log(c, &args, max_retries);
+	if (!log)
+		return -1;
+
+	for (int i = 0; i < le64_to_cpu(log->numrec); i++)
```
martin-belanger, line 939: We may need to move this inside `nvme_discovery_log()` so that it also applies to  `nvmf_get_discovery_wargs()`.

igaw: Blech, I did read that `nvmf_get_discovery_wargs()` is actually calling `nvmf_get_discovery_log()`... let me fix this.

igaw: I decided against moving this to `nvme_discovery_log()` because I wanted to avoid cluttering the 'get' function with this.

Testing took a bit longer because the software target of Linux is kind of funky when it comes to persistent discovery controllers :)

src/nvme/fabrics.c
igaw: Drop this.

Issue #563, opened 2023-01-25
libnvme-mi test crash


igaw:
The test fails on SLE15-SP4 x86_64 installation:

```gdb
(gdb) r
Starting program: /local/dwagner/nvme-cli/subprojects/libnvme/.build/test/test-mi 
Missing separate debuginfos, use: zypper install glibc-debuginfo-2.31-150300.37.1.x86_64
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Running test endpoint_lifetime...  OK
Running test ctrl_lifetime...  OK
Running test read_mi_data...  OK
Running test transport_fail...  OK
Running test transport_describe...  OK
Running test scan_ctrl_list...  OK
Running test invalid_crc...  OK
--- begin test output
crc mismatch
--- end test output
Running test admin_id...  OK
Running test admin_err_mi_resp...  OK
Running test admin_err_nvme_resp...  OK
Running test admin_invalid_formats...  OK
Running test resp_req...  OK
Running test resp_hdr_small...  OK
Running test resp_invalid_type...  OK
Running test resp_csi...  OK
--- begin test output
Command slot mismatch: req 0, resp 1
--- end test output
Running test mi_config_get_mtu...  OK
Running test mi_config_set_freq...  OK
Running test mi_config_set_freq_invalid...  OK
Running test get_features_nodata...  OK
Running test get_features_data...  OK
Running test set_features...  OK
Running test admin_id_alloc_ns_list...  OK
Running test admin_id_active_ns_list...  OK
Running test admin_id_alloc_ns...  OK
Running test admin_id_active_ns...  OK
Running test admin_id_nsid_ctrl_list...  OK
Running test admin_id_secondary_ctrl_list...  OK
Running test admin_ns_mgmt_create...  OK
Running test admin_ns_mgmt_delete...  OK
Running test admin_ns_attach...  OK
Running test admin_ns_detach...  OK
Running test admin_fw_download...  OK
Running test admin_fw_commit...  OK
Running test admin_format_nvm...  OK
Running test admin_sanitize_nvm...  OK
Running test admin_get_log_split...
Program received signal SIGSEGV, Segmentation fault.

gdb) bt -full
#0  0x00007ffff7bcacca in nvme_mi_admin_parse_status (resp=0x7fffffffbac0, result=0x7ffff7de43e4 <_dl_name_match_p+86>) at ../src/nvme/mi.c:545
[...]

PR #565, opened 2023-01-26
test: mi: ensure we're providing full initialised buffers to the libnvme-mi API

jk-ozlabs:
We have a couple of cases in the MI test code where we're either not providing an initialised argument buffer, or a request with no data allocation.

These changes fix those cases, so we're a closer match to actual API usage, and avoid issues (like #563) from uninit arg data.


igaw:
Thanks!

PR #567, opened 2023-01-27
fabrics: Avoid buffer overrun in strchomp

igaw:
strchomp() has an off-by-one error and starts stripping spaces at the byte AFTER the end of the buffer.

See also https://github.com/linux-nvme/nvme-cli/pull/1780

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -47,13 +47,19 @@ const char *nvmf_dev = "/dev/nvme-fabrics";
 
 /**
  * strchomp() - Strip trailing white space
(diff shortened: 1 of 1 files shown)


calebsander:
It actually looks like the existing code was correct here because the value of `max` passed in is 1 less than the length of the buffer:
```c
strchomp(e->traddr, NVMF_TRADDR_SIZE - 1);
strchomp(e->trsvcid, NVMF_TRSVCID_SIZE - 1);
```
This change means that if there is only a single character of padding (which is unlikely), it won't be removed.

PR #572, opened 2023-02-02
nbft: added NBFT table support

johnmeneghini:
Added support for parsing and printing the contents of the NBFT table (per NVMe-oF boot specification v1.0).

Signed-off-by: Stuart Hayes <stuart_hayes@dell.com>
Signed-off-by: Martin Wilck <mwilck@suse.com>
Signed-off-by: Tomas Bzatek <tbzatek@redhat.com>
Signed-off-by: John Meneghini <jmeneghi@redhat.com>

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -9,14 +9,15 @@
 #
 
 api_files = [
+  'fabrics.h',
   'filters.h',
   'ioctl.h',
   'linux.h',
   'log.h',
   'mi.h',
+  'nbft.h',
   'tree.h',
   'types.h',
-  'fabrics.h',
   'util.h'
 ]
 
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -7,6 +7,8 @@ LIBNVME_1_4 {
 		nvme_lookup_key;
 		nvme_set_keyring;
 		nvme_insert_tls_key;
+		nvme_nbft_read;
+		nvme_nbft_free;
 };
 
 LIBNVME_1_3 {
--- a/src/meson.build
+++ b/src/meson.build
@@ -7,6 +7,7 @@
 #
 sources = [
     'nvme/cleanup.c',
+    'nvme/nbft.c',
     'nvme/fabrics.c',
     'nvme/filters.c',
     'nvme/ioctl.c',
@@ -125,6 +126,7 @@ install_headers([
         'nvme/ioctl.h',
         'nvme/linux.h',
         'nvme/log.h',
+        'nvme/nbft.h',
         'nvme/tree.h',
         'nvme/types.h',
         'nvme/util.h',
--- a/src/nvme/nbft.c
+++ b/src/nvme/nbft.c
@@ -1,0 +1,739 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2021-2022, Dell Inc. or its subsidiaries.  All Rights Reserved.
+ *
+ * Authors: Stuart Hayes <Stuart_Hayes@Dell.com>
+ *
+ */
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <arpa/inet.h>
+#include <ccan/endian/endian.h>
+
+#include "private.h"
+#include "nbft.h"
+#include "log.h"
+
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+static __u8 csum(const __u8 *buffer, ssize_t length)
+{
+	int n;
+	__u8 sum = 0;
+
+	for (n = 0; n < length; n++)
+		sum = (__u8)(sum + ((__u8 *)buffer)[n]);
+	return sum;
+}
+
+static void format_ip_addr(char *buf, size_t buflen, __u8 *addr)
+{
+	struct in6_addr *addr_ipv6;
+
+	addr_ipv6 = (struct in6_addr *)addr;
+	if (addr_ipv6->s6_addr32[0] == 0 &&
+	    addr_ipv6->s6_addr32[1] == 0 &&
+	    ntohl(addr_ipv6->s6_addr32[2]) == 0xffff)
+		/* ipv4 */
+		inet_ntop(AF_INET, &(addr_ipv6->s6_addr32[3]), buf, buflen);
+	else
+		/* ipv6 */
+		inet_ntop(AF_INET6, addr_ipv6, buf, buflen);
+}
+
+static bool in_heap(struct nbft_header *header, struct nbft_heap_obj obj)
+{
+	if (le16_to_cpu(obj.length) == 0)
+		return true;
+	if (le32_to_cpu(obj.offset) < le32_to_cpu(header->heap_offset))
+		return false;
+	if (le32_to_cpu(obj.offset) >
+	    le32_to_cpu(header->heap_offset) + le32_to_cpu(header->heap_length))
+		return false;
+	if (le32_to_cpu(obj.offset) + le16_to_cpu(obj.length) >
+	    le32_to_cpu(header->heap_offset) + le32_to_cpu(header->heap_length))
+		return false;
+	return true;
+}
+
+/*
+ *  Return transport_type string (NBFT Table 2)
+ */
+static char *trtype_to_string(__u8 transport_type)
+{
+	switch (transport_type) {
+	case 3:
+		return "tcp";
+	default:
+		return "invalid";
+	}
+}
+
+#define verify(condition, message)					\
+	do {								\
+		if (!(condition)) {					\
+			nvme_msg(NULL, LOG_DEBUG, "file %s: " message "\n", \
+				 nbft->filename);			\
+			return -EINVAL;					\
+		}							\
+	} while (0)
+
+static int __get_heap_obj(struct nbft_header *header, const char *filename,
+			  const char *descriptorname, const char *fieldname,
+			  struct nbft_heap_obj obj, bool is_string,
+			  char **output)
+{
+	if (le16_to_cpu(obj.length) == 0)
+		return -ENOENT;
+
+	if (!in_heap(header, obj)) {
+		nvme_msg(NULL, LOG_DEBUG,
+			 "file %s: field '%s' in descriptor '%s' has invalid offset or length\n",
+			 filename, fieldname, descriptorname);
+		return -EINVAL;
+	}
+
+	/* check that string is zero terminated correctly */
+	*output = (char *)header + le32_to_cpu(obj.offset);
+
+	if (is_string) {
+		if (strnlen(*output, le16_to_cpu(obj.length) + 1) < le16_to_cpu(obj.length))
+			nvme_msg(NULL, LOG_DEBUG,
+				"file %s: string '%s' in descriptor '%s' is shorter (%zd) than specified length (%d)\n",
+				filename, fieldname, descriptorname,
+				strnlen(*output, le16_to_cpu(obj.length) + 1),
+					le16_to_cpu(obj.length));
+		else if (strnlen(*output, le16_to_cpu(obj.length) + 1) >
+			le16_to_cpu(obj.length)) {
+			nvme_msg(NULL, LOG_DEBUG,
+				 "file %s: string '%s' in descriptor '%s' is not zero terminated\n",
+				 filename, fieldname, descriptorname);
+			return -EINVAL;
+		}
+	}
+
+	return 0;
+}
+
+#define get_heap_obj(descriptor, obj, is_string, output)	\
+	__get_heap_obj(header, nbft->filename,			\
+		       stringify(descriptor), stringify(obj),	\
+		       descriptor->obj, is_string,		\
+		       output)
+
+static struct nbft_info_discovery *discovery_from_index(struct nbft_info *nbft, int i)
+{
+	struct nbft_info_discovery **d;
+
+	for (d = nbft->discovery_list; d && *d; d++) {
+		if ((*d)->index == i)
+			return *d;
+	}
+	return NULL;
+}
+
+static struct nbft_info_hfi *hfi_from_index(struct nbft_info *nbft, int i)
+{
+	struct nbft_info_hfi **h;
+
+	for (h = nbft->hfi_list; h && *h; h++) {
+		if ((*h)->index == i)
+			return *h;
+	}
+	return NULL;
+}
+
+static struct nbft_info_security *security_from_index(struct nbft_info *nbft, int i)
+{
+	struct nbft_info_security **s;
+
+	for (s = nbft->security_list; s && *s; s++) {
+		if ((*s)->index == i)
+			return *s;
+	}
+	return NULL;
+}
+
+static int read_ssns_exended_info(struct nbft_info *nbft,
+				  struct nbft_info_subsystem_ns *ssns,
+				  struct nbft_ssns_ext_info *raw_ssns_ei)
+{
+	struct nbft_header *header = (struct nbft_header *)nbft->raw_nbft;
+
+	verify(raw_ssns_ei->structure_id == NBFT_DESC_SSNS_EXT_INFO,
+	       "invalid ID in SSNS extended info descriptor");
+	verify(raw_ssns_ei->version == 1,
+	       "invalid version in SSNS extended info descriptor");
+	verify(le16_to_cpu(raw_ssns_ei->ssns_index) == le16_to_cpu(ssns->index),
+	       "SSNS index doesn't match extended info descriptor index");
+
+	if (!(le32_to_cpu(raw_ssns_ei->flags) & NBFT_SSNS_EXT_INFO_VALID))
+		return -EINVAL;
+
+	if (le32_to_cpu(raw_ssns_ei->flags) & NBFT_SSNS_EXT_INFO_ADMIN_ASQSZ)
+		ssns->asqsz = le16_to_cpu(raw_ssns_ei->asqsz);
+	ssns->controller_id = le16_to_cpu(raw_ssns_ei->cntlid);
+	get_heap_obj(raw_ssns_ei, dhcp_root_path_str_obj, 1, &ssns->dhcp_root_path_string);
+
+	return 0;
+}
+
+static int read_ssns(struct nbft_info *nbft,
+		     struct nbft_ssns *raw_ssns,
+		     struct nbft_info_subsystem_ns **s)
+{
+	struct nbft_header *header = (struct nbft_header *)nbft->raw_nbft;
+	struct nbft_info_subsystem_ns *ssns;
+	__u8 *ss_hfi_indexes = NULL;
+	__u8 *tmp = NULL;
+	int i, ret;
+
+	if (!(le16_to_cpu(raw_ssns->flags) & NBFT_SSNS_VALID))
+		return -EINVAL;
+
+	verify(raw_ssns->structure_id == NBFT_DESC_SSNS,
+	       "invalid ID in SSNS descriptor");
+
+	ssns = calloc(1, sizeof(*ssns));
+	if (!ssns)
+		return -ENOMEM;
+
+	ssns->index = le16_to_cpu(raw_ssns->index);
+
+	/* transport type */
+	verify(raw_ssns->trtype == NBFT_TRTYPE_TCP,
+	       "invalid transport type in SSNS descriptor");
+	strncpy(ssns->transport, trtype_to_string(raw_ssns->trtype), sizeof(ssns->transport));
+
+	/* transport specific flags */
+	if (raw_ssns->trtype == NBFT_TRTYPE_TCP) {
+		if (le16_to_cpu(raw_ssns->trflags) & NBFT_SSNS_PDU_HEADER_DIGEST)
+			ssns->pdu_header_digest_required = true;
+		if (le16_to_cpu(raw_ssns->trflags) & NBFT_SSNS_DATA_DIGEST)
+			ssns->data_digest_required = true;
+	}
+
+	/* primary discovery controller */
+	if (raw_ssns->primary_discovery_ctrl_index) {
+		ssns->discovery = discovery_from_index(nbft,
+						       raw_ssns->primary_discovery_ctrl_index);
+		if (!ssns->discovery)
+			nvme_msg(NULL, LOG_DEBUG,
+				 "file %s: namespace %d discovery controller not found\n",
+				 nbft->filename, ssns->index);
+	}
+
+	/* subsystem transport address */
+	ret = get_heap_obj(raw_ssns, subsys_traddr_obj, 0, (char **)&tmp);
+	if (ret)
+		goto fail;
+
+	format_ip_addr(ssns->traddr, sizeof(ssns->traddr), tmp);
+
+	/* subsystem transport service identifier */
+	ret = get_heap_obj(raw_ssns, subsys_trsvcid_obj, 1, &ssns->trsvcid);
+	if (ret)
+		goto fail;
+
+	/* subsystem port ID */
+	ssns->subsys_port_id = le16_to_cpu(raw_ssns->subsys_port_id);
+
+	/* NSID, NID type, & NID */
+	ssns->nsid = le32_to_cpu(raw_ssns->nsid);
+	ssns->nid_type = raw_ssns->nidt;
+	ssns->nid = raw_ssns->nid;
+
+	/* security profile */
+	if (raw_ssns->security_desc_index) {
+		ssns->security = security_from_index(nbft, raw_ssns->security_desc_index);
+		if (!ssns->security)
+			nvme_msg(NULL, LOG_DEBUG,
+				 "file %s: namespace %d security controller not found\n",
+				 nbft->filename, ssns->index);
+	}
+
+	/* HFI descriptors */
+	ret = get_heap_obj(raw_ssns, secondary_hfi_assoc_obj, 0, (char **)&ss_hfi_indexes);
+	if (ret)
+		goto fail;
+
+	ssns->hfis = calloc(le16_to_cpu(raw_ssns->secondary_hfi_assoc_obj.length) + 2,
+			    sizeof(*ssns->hfis));
+	if (!ssns->hfis) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+	ssns->hfis[0] = hfi_from_index(nbft, raw_ssns->primary_hfi_desc_index);
+	if (!ssns->hfis[0]) {
+		nvme_msg(NULL, LOG_DEBUG,
+			 "file %s: SSNS %d: HFI %d not found\n",
+			 nbft->filename, ssns->index, raw_ssns->primary_hfi_desc_index);
+		ret = -EINVAL;
+		goto fail;
+	}
+	for (i = 0; i < le16_to_cpu(raw_ssns->secondary_hfi_assoc_obj.length); i++) {
+		ssns->hfis[i + 1] = hfi_from_index(nbft, ss_hfi_indexes[i]);
+		if (ss_hfi_indexes[i] && !ssns->hfis[i + 1])
+			nvme_msg(NULL, LOG_DEBUG,
+				 "file %s: SSNS %d HFI %d not found\n",
+				 nbft->filename, ssns->index, ss_hfi_indexes[i]);
+		else
+			ssns->num_hfis++;
+	}
+
+	/* SSNS NQN */
+	ret = get_heap_obj(raw_ssns, subsys_ns_nqn_obj, 1, &ssns->subsys_nqn);
+	if (ret)
+		goto fail;
+
+	/* SSNS extended info */
+	if (raw_ssns->flags & NBFT_SSNS_EXTENDED_INFO_IN_USE) {
+		struct nbft_ssns_ext_info *ssns_extended_info;
+
+		if (!get_heap_obj(raw_ssns, ssns_extended_info_desc_obj, 0,
+				  (char **)&ssns_extended_info))
+			read_ssns_exended_info(nbft, ssns, ssns_extended_info);
+	}
+
+	*s = ssns;
+	return 0;
+
+fail:
+	free(ssns);
+	return ret;
+}
+
+static int read_hfi_info_tcp(struct nbft_info *nbft,
+			     struct nbft_hfi_info_tcp *raw_hfi_info_tcp,
+			     struct nbft_info_hfi *hfi)
+{
+	struct nbft_header *header = (struct nbft_header *)nbft->raw_nbft;
+
+	if ((raw_hfi_info_tcp->flags & NBFT_HFI_INFO_TCP_VALID) == 0)
+		return -EINVAL;
+
+	verify(raw_hfi_info_tcp->structure_id == NBFT_DESC_HFI_TRINFO,
+	       "invalid ID in HFI transport descriptor");
+	verify(raw_hfi_info_tcp->version == 1,
+	       "invalid version in HFI transport descriptor");
+	if (le16_to_cpu(raw_hfi_info_tcp->hfi_index) != hfi->index)
+		nvme_msg(NULL, LOG_DEBUG,
+			 "file %s: HFI descriptor index %d does not match index in HFI transport descriptor\n",
+			 nbft->filename, hfi->index);
+
+	hfi->tcp_info.pci_sbdf = le32_to_cpu(raw_hfi_info_tcp->pci_sbdf);
+	memcpy(hfi->tcp_info.mac_addr, raw_hfi_info_tcp->mac_addr,
+	       sizeof(raw_hfi_info_tcp->mac_addr));
+	hfi->tcp_info.vlan = le16_to_cpu(raw_hfi_info_tcp->vlan);
+	hfi->tcp_info.ip_origin = raw_hfi_info_tcp->ip_origin;
+	format_ip_addr(hfi->tcp_info.ipaddr, sizeof(hfi->tcp_info.ipaddr),
+		       raw_hfi_info_tcp->ip_address);
+	hfi->tcp_info.subnet_mask_prefix = raw_hfi_info_tcp->subnet_mask_prefix;
+	format_ip_addr(hfi->tcp_info.gateway_ipaddr, sizeof(hfi->tcp_info.ipaddr),
+		       raw_hfi_info_tcp->ip_gateway);
+	hfi->tcp_info.route_metric = le16_to_cpu(raw_hfi_info_tcp->route_metric);
+	format_ip_addr(hfi->tcp_info.primary_dns_ipaddr,
+		       sizeof(hfi->tcp_info.primary_dns_ipaddr),
+		       raw_hfi_info_tcp->primary_dns);
+	format_ip_addr(hfi->tcp_info.secondary_dns_ipaddr,
+		       sizeof(hfi->tcp_info.secondary_dns_ipaddr),
+		       raw_hfi_info_tcp->secondary_dns);
+	if (raw_hfi_info_tcp->flags & NBFT_HFI_INFO_TCP_DHCP_OVERRIDE) {
+		hfi->tcp_info.dhcp_override = true;
+		format_ip_addr(hfi->tcp_info.dhcp_server_ipaddr,
+			       sizeof(hfi->tcp_info.dhcp_server_ipaddr),
+			       raw_hfi_info_tcp->dhcp_server);
+	}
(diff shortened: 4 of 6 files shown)


src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+ * Copyright (c) 2021-2022, Dell Inc. or its subsidiaries.  All Rights Reserved.
+ *
+ * Authors: Stuart Hayes <Stuart_Hayes@Dell.com>
+ *
+ */
+#ifndef _NBFT_H
+#define _NBFT_H
+
+#include <uuid/uuid.h>
+#include <ccan/list/list.h>
```
tbzatek, line 13: `ccan` headers are not shipped as part of `libnvme` public headers nor pulled as a dependency externally. This will cause compilation errors when included from third party.

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	bool dhcp_override;
+};
+
+struct nbft_hfi {
+	int index;
+	char transport[8];
+	//uuid_t *host_id;
+	//char *host_nqn;
+	struct nbft_hfi_info_tcp tcp_info;
+	struct list_node node;
```
tbzatek, line 53: See above, `ccan` types cannot be used in public headers.

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+
+#include <uuid/uuid.h>
+#include <ccan/list/list.h>
+
+enum primary_admin_host_flag {
+	not_indicated,
+	unselected,
+	selected,
+	reserved,
+};
```
tbzatek, line 20: Code style suggestion: add namespace/prefix and convert to uppercase?

src/libnvme.map
```diff
@@ -370,6 +370,8 @@ LIBNVME_1_0 {
 		nvmf_treq_str;
 		nvmf_trtype_str;
 		nvmf_update_config;
+		nbft_read;
+		nbft_free;
```
tbzatek, line 374: These need to go to the `LIBNVME_1_4` section.

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	uuid_t *id;
+	char *nqn;
+	bool host_id_configured;
+	bool host_nqn_configured;
+	enum primary_admin_host_flag primary;
+};
+
+struct nbft_hfi_info_tcp {
+	__u32 pci_sbdf;
+	__u8 *mac_addr;
```
tbzatek, line 32: Is this supposed to be a pointer? Everything else is fixed-size.

tbzatek (commented):
Mostly forwarding my previous remarks from the private repo.

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	struct nbft_security *security;
+	struct nbft_hfi *hfi;
+	char *uri;
+	char *nqn;
+};
+
+struct nbft_security {
+	int index;
+	struct list_node node;
+	/* TODO add fields */
```
igaw, line 68: Please add those if they are missing

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	int asqsz;
+	char *dhcp_root_path_string;
+};
+
+struct nbft_info {
+	struct list_node node;
+	const char *filename;
+	__u8 *raw_nbft;
+	ssize_t raw_nbft_size;
+	/* host info... should match other NBFTs */
```
igaw, line 111: Use proper kdoc style to document these member.s

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	NBFT_HFI,
+	NBFT_SSNS,
+	NBFT_SECURITY,
+	NBFT_DISCOVERY,
+	NBFT_HFI_TRANSPORT,
+	RESERVED_8,
+	NBFT_SSNS_EXTENDED_INFO,
+};
+
+typedef struct __attribute__((__packed__)) heap_obj_s {
```
igaw, line 39: - No typedefs, just use struct heap_obj as name throughout the code. 
- remove `__attribute__((__packed__))`, it doesn't do what you expect it does. When all members of the struct are correctly padded, the attribute will remove all alignment information as all accesses are on byte level. `pahole` is your friend

I only mention it here once. There are more of these.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	__u32 offset;
+	__u16 length;
+} heap_obj;
+
+/*
+ * HEADER (Figure 8)
```
igaw, line 45: Add a self containing documentation and only reference for additional information to external resources. I just mention it once here. There are more of these comments.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	int n;
+	__u8 sum = 0;
+
+	for (n = 0; n < length; n++) {
+		sum = (__u8)(sum + ((__u8 *)buffer)[n]);
+	}
+	return sum;
+}
+
+static void format_ip_addr(char *buf, size_t buflen, __u8 *addr) {
```
igaw, line 306: For function the '{' goes on a new line.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+		/* ipv6 */
+		inet_ntop(AF_INET6, addr_ipv6, buf, buflen);
+}
+
+static int in_heap(struct raw_nbft_header *header, heap_obj obj)
+{
+	if (obj.length == 0)
+		return 1;
+	if (obj.offset < header->heap_offset)
+		goto bad;
```
igaw, line 325: return directly, no goto needed.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+
+/*
+ *  Return transport_type string (NBFT Table 2)
+ */
+static char *trtype_to_string(__u8 transport_type)
+{
+	switch (transport_type) {
+		case 3:
+			return "tcp";
+			break;
```
igaw, line 343: break is unnecessary after return.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+		return -EINVAL;								\
+	}
+
+static int __get_heap_obj(struct raw_nbft_header *header, const char *filename,
+			  const char *descriptorname, const char *fieldname,
+			  heap_obj obj, bool is_string,
+			  char **output)
+{
+	if (obj.length == 0) {
+		*output = NULL;
```
igaw, line 362: Usually, if an error is returned the callee is not supposed to change anything. The caller is supposed to handle the error correctly.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+			 filename, fieldname, descriptorname);
+		return -EINVAL;
+	}
+
+	/* check that string is zero terminated correctly */
+	*output = (char *)header + obj.offset;
+
+	if (is_string) {
+		if (strnlen(*output, obj.length + 1) < obj.length)
+			nvme_msg(NULL, LOG_DEBUG, "file %s: string '%s' in descriptor '%s' is shorter (%ld) than specified length (%d)\n",
```
igaw, line 377: Overlong lines.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	__get_heap_obj(header, nbft->filename,			\
+		       stringify(descriptor), stringify(obj),	\
+		       descriptor->obj, is_string,				\
+		       output)
+
+static struct nbft_discovery *discovery_from_index(struct nbft_info *nbft, int i)
+{
+	struct nbft_discovery *d;
+
+	list_for_each(&nbft->discovery_list, d, node)
```
igaw, line 399: Please add '{}' when it's more than one line.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+{
+	struct nbft_security *s;
+
+	list_for_each(&nbft->security_list, s, node)
+		if (s->index == i)
+			return s;
+	return NULL;
+}
+
+static int read_ssns_exended_info(struct nbft_info *nbft, struct nbft_subsystem_ns *ssns, struct raw_nbft_ssns_extended_info *ssns_ei)
```
igaw, line 425: overlong lines.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+		return -EINVAL;
+	verify(raw_ssns->structure_id == NBFT_SSNS, "invalid ID in SSNS descriptor");
+
+	ssns = calloc(1, sizeof(*ssns));
+	if (!ssns) {
+		return -ENOMEM;
+	}
+
+	/* index */
+	ssns->index = raw_ssns->index;
```
igaw, line 461: What does the comment try to tell me? Trivial comments should be avoided

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+discovery_fail:
+	free(discovery);
+	return ret;
+}
+
+static int read_security(struct nbft_info *nbft, struct raw_nbft_security *raw_security, struct nbft_security **s)
+{
+	/*
+	 *  TO DO add security stuff
+	 */
```
igaw, line 688: Please either add the code or just leave those empty functions away.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	while ((subtable = list_pop(&nbft->discovery_list, struct nbft_discovery, node)))
+		free(subtable);
+	while ((subtable = list_pop(&nbft->security_list, struct nbft_security, node)))
+		free(subtable);
+	while ((ns = list_pop(&nbft->subsystem_ns_list, struct nbft_subsystem_ns, node))) {
+		free(ns->hfis);
+		free(ns);
+	}
+	free(nbft->raw_nbft);
+	free((void *)nbft->filename);
```
igaw, line 865: Why is the cast needed?

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	}
+
+	raw_nbft_size = ftell(raw_nbft_fp);
+	rewind(raw_nbft_fp);
+
+	raw_nbft = malloc(raw_nbft_size);
+	if (!raw_nbft) {
+		nvme_msg(NULL, LOG_ERR, "Failed to allocate memory for NBFT table");
+		ret = -ENOMEM;
+		goto fail_1;
```
igaw, line 906: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#centralized-exiting-of-functions:

> Choose label names which say what the goto does or why the goto exists. An example of a good name could be out_free_buffer: if the goto frees buffer. Avoid using GW-BASIC names like err1: and err2:, as you would have to renumber them if you ever add or remove exit paths, and they make correctness difficult to verify anyway.

igaw (changes requested):
I think we should have a unit test for the parser. My idea would be to grab a memory dump and use it to verify that the parser works. 

`nbft_read` could accept a descriptor instead of a filename and we could run this code with valgrind without having to setup any hardware.

Furthermore, if we modify stuff it could be easily verified nothing breaks.

igaw:
The `checkpatch.pl` script the Linux tree helps to spot a lot of the style issues.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+ */
+struct __attribute__((__packed__)) raw_nbft_ssns {
+	__u8 structure_id;
+	__u16 index;
+	__u16 flags;
+	__u8 transport_type;
+	__u16 transport_specific_flags;
+	__u8 primary_discovery_ctrl_index;
+	__u8 reserved1;
+	heap_obj subsystem_transport_address;
```
martin-belanger, line 181: Just nitpicking - The naming convention for "transport address" is `traddr` and for "transport service id" is `trsvcid`.

This comment applies to both `nbft.c` and `nbft.h`. External code such as `nvme-cli` that refers to these definitions will obviously be affected as well.

tbzatek:
Alright, I'm going to slowly start addressing the points raised above to get things moving. As this will change the API, I'll be posting fixes to https://github.com/linux-nvme/nvme-cli/pull/1791 accordingly.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	__u8 reserved1;
+	heap_obj discovery_ctrl_addr;
+	heap_obj discovery_controller_nqn;
+	__u8 reserved2[14];
+};
+
+#define DISCOVERYFLAG_VALID	0x01
+
+/*
+ *  End of NBFT ACPI table definitions
```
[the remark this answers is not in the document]
tbzatek: Agree, I also think these `nbft_raw_*` definitions should be public as this is essentially translation of the TP8012.

tbzatek:
TODO:
- [ ] fix endianness

tbzatek:
I'm posting my fixes to side branches as the official timberland one is used as a source for PoC and I still need to properly test my changes. I do plan to add some tests as well. There's still some work to do, for now the code lives at:
https://github.com/timberland-sig/libnvme/tree/timberlandmaster_final_v13-fixes
https://github.com/timberland-sig/nvme-cli/tree/timberlandmaster_final_v23-fixes

src/nvme/nbft.c
tbzatek: Hmm, something to look at more closely, for now I left the attribute there. The thing is that e.g. the `heap_obj_s` struct is 32+16 bits large and that doesn't align well, when embedded in other structs. As the ACPI NBFT table binary format is fixed I don't see a room for adding extra padding or reorganizing the elements. Or maybe I'm missing something...

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	int controller_id;
+	int asqsz;
+	char *dhcp_root_path_string;
+};
+
+struct nbft_info {
+	struct list_node node;
+	const char *filename;
+	__u8 *raw_nbft;
```
tbzatek, line 109: Code style question: I was also wondering if there's any reason for keeping the whole binary table in memory (from public API point of view). I.e. once the table is parsed it doesn't change and all the information are (or will be) provided in the respective struct elements.

Technically most of the strings are referenced to this chunk of memory so this would have to change.

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	not_indicated,
+	unselected,
+	selected,
+	reserved,
+};
+
+struct nbft_host {
+	uuid_t *id;
+	char *nqn;
+	bool host_id_configured;
```
tbzatek, line 25: Code style question: now that I proposed the TP8012 enums to be public, does it make sense to provide all the respective flags as bools or better to reuse the original enums (with the need to apply bitmasks)? Thinking about possible future expansion here - it's tough to extend public structs with new elements, enums would absorb some burden.

src/nvme/nbft.c
igaw: It might not work for all data types, but at least for the nvme data types it doesn't seem to be necessary, as the members are naturally aligned. At least I haven't found one which it wasn't the case.

src/nvme/nbft.c
tbzatek: Looking at `struct nbft_security` there's substantial number of elements and objects to be parsed and translated. Better be addressed in a separate PR.

src/nvme/nbft.h
tbzatek: Fixed.

src/nvme/nbft.h
tbzatek: Fixed.

src/libnvme.map
tbzatek: Fixed.

src/nvme/nbft.h
tbzatek: Fixed.

src/nvme/nbft.h
```diff
@@ -0,0 +1,141 @@
...
+	char *host_name;
+	bool this_hfi_is_default_route;
+	bool dhcp_override;
+};
+
+struct nbft_hfi {
+	int index;
+	char transport[8];
+	//uuid_t *host_id;
+	//char *host_nqn;
```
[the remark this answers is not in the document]
tbzatek: Removed.

src/nvme/nbft.c
tbzatek: Fixed.

src/nvme/nbft.c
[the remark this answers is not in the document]
tbzatek: Fixed.

src/nvme/nbft.c
tbzatek: Fixed.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+	    && addr_ipv6->s6_addr32[1] == 0
+	    && (ntohl(addr_ipv6->s6_addr32[2]) == 0xffff) )
+		/* ipv4 */
+		inet_ntop(AF_INET, &(addr_ipv6->s6_addr32[3]), buf, buflen);
+	else
+		/* ipv6 */
+		inet_ntop(AF_INET6, addr_ipv6, buf, buflen);
+}
+
+static int in_heap(struct raw_nbft_header *header, heap_obj obj)
```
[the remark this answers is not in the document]
tbzatek: Fixed.

src/nvme/nbft.c
tbzatek: Fixed.

src/nvme/nbft.c
tbzatek: Fixed.

src/nvme/nbft.c
tbzatek: Fixed.

src/nvme/nbft.c
tbzatek: Not applicable anymore.

src/nvme/nbft.c
tbzatek: Fixed.

src/nvme/nbft.c
tbzatek: Refactored.

src/nvme/nbft.c
tbzatek: Refactored.

src/nvme/nbft.h
tbzatek: Postponed - see the corresponding `nbft.c` note below.

tbzatek:
Thanks @johnmeneghini for merging this. One more fix in #6 to be tested & merged plus this branch needs to be rebased against upstream libnvme master.

src/nvme/nbft.c
```diff
@@ -0,0 +1,722 @@
...
+
+	/*
+	 * header
+	 */
+	header = (struct nbft_header *)raw_nbft;
+
+	verify(strncmp(header->signature, NBFT_HEADER_SIG, 4) == 0, "invalid signature");
+	verify(header->length <= raw_nbft_size, "length in header exceeds table length");
+	verify(header->major_revision == 1, "unsupported major revision");
+	verify(header->minor_revision == 0, "unsupported minor revision");
```
tbzatek, line 531: One more thing: generally NVMe specs are made to be future-proof with _reserved_ fields extended with new stuff, typically maintaining backwards compatibility. In case a new UEFI firmware comes out and a customer would stick to an old OS release (and believe me, it's not an exception), this might fail.

src/nvme/nbft.c
johnmeneghini: NVMe security support is not are part of NVMe Boot 1.0.  There is more work that needs to be done in specification and in the EDK2 code to support NVMe security. I think we can ignore this stuff for now.  I will remove the TO DO.

johnmeneghini:
Next steps for this pull request:

1.  rebase this pull request to master
2.  rebase https://github.com/linux-nvme/nvme-cli/pull/1791 to master
3. retest everything with https://github.com/dracutdevs/dracut/pull/2184
 
Following this I propose we mere this into libnvme. and continue to work on https://github.com/linux-nvme/nvme-cli/pull/1791.

igaw:
@johnmeneghini thanks for the progress report. So I'll wait for the rebase of this PR. BTW, are you also squashing the commits too? I don't think we need to see the development history in upstream.

johnmeneghini:
> @johnmeneghini thanks for the progress report. So I'll wait for the rebase of this PR. BTW, are you also squashing the commits too? I don't think we need to see the development history in upstream.

OK. Will do.  I will squash and rebase now.

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+	int i, ret = 0;
+
+	/*
+	 * read in raw nbft file
+	 */
+	raw_nbft_fp = fopen(filename, "rb");
+	if (raw_nbft_fp == NULL) {
+		nvme_msg(NULL, LOG_ERR, "Failed to open %s: %s\n",
+			 filename, strerror(errno));
+		return -EINVAL;
```
martin-belanger, line 684: After seeing comments by @igaw on an unrelated PR, I was wondering if we should set the `errno` and return `1` on error instead of returning an error code.

src/nvme/nbft.c
johnmeneghini: this has been fixed

src/nvme/nbft.c
johnmeneghini: This To Do comment still remains.

src/nvme/nbft.c
johnmeneghini: Please propose a patch for this in the Timberland-sig libnvme repository.

johnmeneghini:
OK. The good news is:  I have squashed and rebased both https://github.com/linux-nvme/libnvme/pull/572 and  https://github.com/linux-nvme/nvme-cli/pull/1791 onto the HEAD of the master branch.  After some work on the libnvme.spec file in my rh_linux_poc [here]( https://github.com/timberland-sig/rh-linux-poc/pull/7/commits/f86161912acc126208d55197576c9f91110cc442) I was able to get my fedora RPMs to compile and install.  However, testing these new rebased bits in my QEMU POC fails.  After rebasing everything to the HEAD of the master branch in libnvme and nvme-cli, these changes no longer work with https://github.com/dracutdevs/dracut/pull/2184 and with https://github.com/timberland-sig/edk2/commit/9e63dc0cdfa052c3185d9451f2a4fb67886b959e. I am no longer able to boot my QEMU host with NVMe/TCP using these bits.

mwilck:
> I am no longer able to boot my QEMU host with NVMe/TCP using these bits.

Hm, that's weird. Have you tried just running `nvme show-nbft` on a system booted with the previous releases?

tbzatek:
I've tested the updated branches and `nvme show-nbft` seems to be working fine, haven't noticed any difference in the output from the state of a week ago.

johnmeneghini:
On 3/29/23 05:27, Martin Wilck wrote:
>     I am no longer able to boot my QEMU host with NVMe/TCP using these bits.
> 
> Hm, that's weird. Have you tried just running |nvme show-nbft| on a system booted with the previous releases?

Yes,  I reported the same thing last night.  Apparently the rebase process broke something.

I was working on this till 1:00am lsst night and finally decided to push the broken branch upstream.

A bug has been introduced somewhere... maybe an incompatibility with the dracut code.

I'm looking for help to fix the problem.

mwilck:
> I'm looking for help to fix the problem.
I'd love to help but I fear I won't have time for it today.

tbzatek:
Another tiny change in PR https://github.com/timberland-sig/libnvme/pull/7 should fix the github actions checks.

johnmeneghini:
I have redone my work and retested. All tests are passing my nvme/tcp boot test on the rh-linux-poc and this branch is ready for final review.  Once the final few issues are resolved, I think this branch can be merged.

tbzatek:
> I have redone my work and retested. All tests are passing my nvme/tcp boot test on the rh-linux-poc and this branch is ready for final review. Once the final few issues are resolved, I think this branch can be merged.

The patch from https://github.com/timberland-sig/libnvme/pull/7 doesn't seem to be included in your last rebase. The rest of the code looks okay.

src/nvme/nbft.c
johnmeneghini: So what's recommendation here.  Are you saying you want to take both of these verify statements  out?  Do you want to remove these?  I don't think it makes sense to remove just one w/out removing the other.  Both of these verify statements will result in a run time error, not a core, so I think it's safe to leave them.

mwilck: I agree. I would actually suggest to try to parse the NBFT regardless of the version, and just spit out an error message, even if there's major version mismatch.

src/nvme/nbft.c
```diff
@@ -193,14 +200,19 @@ static int read_ssns(struct nbft_info *nbft,
...
 		return -EINVAL;
+        }
+
 	verify(raw_ssns->structure_id == NBFT_DESC_SSNS,
 	       "invalid ID in SSNS descriptor");
 
 	ssns = calloc(1, sizeof(*ssns));
-	if (!ssns)
+	if (!ssns) {
+                errno = ENOMEM;
```
mwilck, line 213: The libnvme convention would be to return -1 after setting errno. Returning the the negative error number *and* setting errno seems strange to me.

johnmeneghini: Does it matter?  Either way the function returns a negative value.  BTW, `-1` is `EPRM`.

```
#define EPERM            1      /* Operation not permitted */
```

src/nvme/nbft.c
johnmeneghini: Fixed

src/nvme/nbft.c
johnmeneghini: I hesitate to change this to `-1` because I have no idea if dracut will care.   If you think it's safe to do, I'll go ahead and change everything to -1.  BTW, we have the same issue in nvme-cli.

src/nvme/nbft.c
```diff
@@ -0,0 +1,947 @@
...
+			break;
+		default:
+			return "invalid";
+			break;
+	}
+}
+
+#define verify(condition, message)							\
+	if (!(condition)) {								\
+		nvme_msg(NULL, LOG_DEBUG, "file %s: " message "\n", nbft->filename);	\
```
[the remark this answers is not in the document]
johnmeneghini: How would it spam stderr? This macro immediately returns and error so all functions calling this macro would exist after the first error.

src/nvme/nbft.c
mwilck: > Does it matter?

@igaw or the other maintainers would need to decide. IMO it matters.

Dracut only distinguishes between zero (success) and non-zero (failure). Shell scripts don't usually rely on any `errno.h` values.

igaw: The libnvme follows (unfortunatly) the POSIX API design. That means it supposed to  return 1 on error with errno set. We have parts of the API though which returns -1 and sets errno. So my take is that within a subset of the libnvme API (header file) we should aim to be consistent. 

Note, internally we use negative returning negative error codes, this is just for the public API where we
do a return 1, errno=CODE.

Anyway for v2, we have plans to change this part to just return the error codes directly with the option to have more structured error messages means if necessary. Similar to the GLib APIs.

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+ *
+ * @nbft:     Parsed NBFT table data.
+ * @filename: Filename of the raw NBFT table to read.
+ *
+ * Read and parse the specified NBFT file into a struct nbft_info.
+ * Free with nbft_free().
+ *
+ * Return: 0 on success, errno otherwise.
+ */
+int nbft_read(struct nbft_info **nbft, const char *filename)
```
igaw, line 693: Prefix this function with `nvme_`

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+	}
+
+	return 0;
+}
+
+/**
+ * nbft_free() - Free the struct nbft_info and its contents
+ * @nbft: Parsed NBFT table data.
+ */
+void nbft_free(struct nbft_info *nbft)
```
igaw, line 656: Prefix this function with `nvme_`

src/meson.build
```diff
@@ -120,6 +121,7 @@ install_headers('libnvme.h', install_mode: mode)
 install_headers('libnvme-mi.h', install_mode: mode)
 install_headers([
         'nvme/api-types.h',
+        'nvme/nbft.h',
```
igaw, line 124: This is also supposed to be alphabetical sorted.

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+#include <ccan/endian/endian.h>
+
+#include "private.h"
+#include "nbft.h"
+#include "log.h"
+
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+static __u8 csum(void *buffer, int length)
```
igaw, line 24: buffer should propebly be `const __u8 *` type, this avoids the cast below and the caller is actually passing in `__u8 *`

length should be type `ssize_t` because of the `raw_nbft_size` definition.

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+static void format_ip_addr(char *buf, size_t buflen, __u8 *addr)
+{
+	struct in6_addr *addr_ipv6;
+
+	addr_ipv6 = (struct in6_addr *)addr;
+	if (addr_ipv6->s6_addr32[0] == 0 &&
+	    addr_ipv6->s6_addr32[1] == 0 &&
+	    ntohl(addr_ipv6->s6_addr32[2]) == 0xffff)
+		/* ipv4 */
+		inet_ntop(AF_INET, &(addr_ipv6->s6_addr32[3]), buf, buflen);
```
igaw, line 43: No need to check if the `inet_ntop` is successful?

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+		return "tcp";
+	default:
+		return "invalid";
+	}
+}
+
+#define verify(condition, message)					\
+	do {								\
+		if (!(condition)) {					\
+			nvme_msg(NULL, LOG_DEBUG, "file %s: " message "\n", \
```
igaw, line 80: I am not really happy with printing on stdout in the library. In this case it's just for debugging. The fabrics parts of the library we avoided this problem by attaching a logging fd to the `nvme_root_t` object. Obviously, here it's a bit difficult as there is no context we could use to pass in a fd.

src/nvme/nbft.c
```diff
@@ -0,0 +1,739 @@
...
+static int __get_heap_obj(struct nbft_header *header, const char *filename,
+			  const char *descriptorname, const char *fieldname,
+			  struct nbft_heap_obj obj, bool is_string,
+			  char **output)
+{
+	if (le16_to_cpu(obj.length) == 0)
+		return -ENOENT;
+
+	if (!in_heap(header, obj)) {
+		nvme_msg(NULL, LOG_DEBUG,
```
igaw, line 98: Trying to print when we can't allocate memory is almost certainly also failing. Just return the error code.

igaw (changes requested):
Sorry for the review delay. Was busy playing fire fighter.

johnmeneghini:
Thanks for your review Daniel.

I will work on getting these updates done tonight.

src/nvme/nbft.c
johnmeneghini: Thanks for explaining Daniel.  So all of the internal return -ERROR functions can stay, it's only the public APIs `nbft_free` and `nbft_read` that need to honor the Posix API by using `return 1, errno=ERROR;`  This makes the change much simpler.

src/nvme/nbft.c
johnmeneghini: I agree. But it will take some work to get rid of this one.  Even if I fix this here, there are many more places where nvme_msg is called.  Maybe we can open an issue for this and move forward. 

```
Functions calling this function: nvme_msg

  File   Function          Line
0 nbft.c verify             84 nvme_msg(NULL, LOG_DEBUG, "file %s: " message "\n", \
1 nbft.c __get_heap_obj     99 nvme_msg(NULL, LOG_DEBUG,
2 nbft.c __get_heap_obj    110 nvme_msg(NULL, LOG_DEBUG,
3 nbft.c __get_heap_obj    117 nvme_msg(NULL, LOG_DEBUG,
4 nbft.c read_ssns         231 nvme_msg(NULL, LOG_DEBUG,
5 nbft.c read_ssns         264 nvme_msg(NULL, LOG_DEBUG,
6 nbft.c read_ssns         282 nvme_msg(NULL, LOG_DEBUG,
7 nbft.c read_ssns         291 nvme_msg(NULL, LOG_DEBUG,
8 nbft.c read_hfi_info_tcp 335 nvme_msg(NULL, LOG_DEBUG,
9 nbft.c read_hfi          412 nvme_msg(NULL, LOG_DEBUG,
a nbft.c read_discovery    454 nvme_msg(NULL, LOG_DEBUG,
b nbft.c read_discovery    460 nvme_msg(NULL, LOG_DEBUG,
c nbft.c nvme_nbft_read    695 nvme_msg(NULL, LOG_ERR, "Failed to open %s: %s\n",
d nbft.c nvme_nbft_read    703 nvme_msg(NULL, LOG_ERR, "Failed to read from %s: %s\n",
e nbft.c nvme_nbft_read    715 nvme_msg(NULL, LOG_ERR, "Failed to allocate memory for NBFT table");
f nbft.c nvme_nbft_read    723 nvme_msg(NULL, LOG_ERR, "Failed to read from %s: %s\n",
g nbft.c nvme_nbft_read    737 nvme_msg(NULL, LOG_ERR, "Could not allocate memory for NBFT\n");
h nbft.c nvme_nbft_read    748 nvme_msg(NULL, LOG_ERR, "Failed to parse %s\n", filename);

```

src/nvme/nbft.c
johnmeneghini: fixed with my latest update

src/libnvme.map
```diff
@@ -379,6 +379,8 @@ LIBNVME_1_0 {
...
```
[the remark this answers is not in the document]
johnmeneghini: fixed with my latest update

src/nvme/fabrics.h
```diff
@@ -12,6 +12,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include "tree.h"
+#include "nbft.h"
```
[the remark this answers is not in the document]
johnmeneghini: fixed with my latest update

src/nvme/nbft.c
johnmeneghini: This change broke compatibility with the current dracut patches.  So I guess we're moving a head.  My first goal is to get these libnvme changes upstream.  Along the way we are going to break some thing.  We'll have to deal with those next.

src/meson.build
johnmeneghini: fixed with my latest update.

src/nvme/nbft.c
johnmeneghini: fixed with my latest update.

src/nvme/nbft.c
johnmeneghini: fixed with my latest update.

doc/meson.build
```diff
@@ -16,6 +16,7 @@ api_files = [
   'mi.h',
   'tree.h',
   'types.h',
+  'nbft.h',
```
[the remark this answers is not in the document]
johnmeneghini: fixed with my latest update.

src/nvme/nbft.c
johnmeneghini: fixed with my latest update

src/nvme/nbft.c
johnmeneghini: I just fixed the compatibility problem with my last patch to nvme-cli.  See: https://github.com/linux-nvme/nvme-cli/pull/1791/commits/9cfa0ec51b55c355ee507cbd2171427fb894211a

src/nvme/nbft.c
johnmeneghini: > No need to check if the `inet_ntop` is successful?

The answer to this question is: NO we don't want to check if `inet_ntop` is successful.  Doing so will cause the `read_hfi_info_tcp` function to inappropriately fail because many of the different ip addressed in the nbft are blank.  I've reverted patch https://github.com/linux-nvme/libnvme/pull/572/commits/8411348419752d0dd511bd32bdee7fe398796129 from this pull request as testing with this change failed.  I am not able to boot with NVMe/TCP with this patch.

johnmeneghini:
I have squashed everything and re based onto v1.4.

src/nvme/nbft.c
igaw: Ah okay, I am looking at the diff right now and wondered what happened with this code.

src/nvme/nbft.c
igaw: Yes, let's move forward and address this later.

PR #574, opened 2023-02-06
build: Make json-c dependency optional again

igaw:
Make support for JSON related functionality optional again. Some embedded configuration do not want to have a lot of dependencies.


igaw:
FWIW, nvme-cli handles `ENOTSUP` gracefully.

PR #578, opened 2023-02-14
types: fix endians

keithbusch:
All the nvme defined payloads are in little endian. All the known users are already using the appropriate le*_to_cpu() accessors, so just fix the type annotations.

Cc: Klaus Jensen <its@irrelevant.dk>
Signed-off-by: Keith Busch <kbusch@kernel.org>

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -2239,8 +2239,8 @@ struct nvme_id_ctrl_nvm {
 	__u8     wzsl;
 	__u8     wusl;
 	__u8     dmrl;
-	__u32    dmrsl;
-	__u64    dmsl;
+	__le32   dmrsl;
+	__le64   dmsl;
 	__u8     rsvd16[4080];
 };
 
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -4779,10 +4779,12 @@ struct nvme_id_directives {
  * enum nvme_directive_types - Directives Supported or Enabled
  * @NVME_ID_DIR_ID_BIT: Identify directive is supported
  * @NVME_ID_DIR_SD_BIT: Streams directive is supported
+ * @NVME_ID_DIR_DP_BIT: Direct Placement directive is supported
  */
 enum nvme_directive_types {
 	NVME_ID_DIR_ID_BIT	= 0,
 	NVME_ID_DIR_SD_BIT	= 1,
+	NVME_ID_DIR_DP_BIT	= 2,
```
igaw, line 4787: This is not an endian fix :)

igaw:
Looks good, I'd just add a prefix to the commit title "types: ...".

src/nvme/types.h
keithbusch: oops, slipped in from another pending change!

PR #582, opened 2023-02-22
Tp4146

sc108-lee:


src/nvme/types.h
```diff
@@ -5930,6 +5930,13 @@ struct nvme_mi_vpd_hdr {
  *				      Lockdown: The command was aborted due to
  *				      command execution being prohibited by
  *				      the Command and Feature Lockdown.
+ * @NVME_SC_FDP_DISABLED: Command is not allowed when Flexible Data Placement
```
igaw, line 5933: Align 'Command' with the above lines.

src/nvme/types.h
```diff
@@ -6276,7 +6283,9 @@ enum nvme_status_field {
 	NVME_SC_CMD_INTERRUPTED			= 0x21,
 	NVME_SC_TRAN_TPORT_ERROR		= 0x22,
 	NVME_SC_PROHIBITED_BY_CMD_AND_FEAT	= 0x23,
-	NVME_SC_ADMIN_CMD_MEDIA_NOT_READY	= 0x24,
+	NVME_SC_FDP_DISABLED				= 0x24,
+	NVME_SC_INVALID_PLACEMENT_HANDLE_LIST	= 0x29,
+	NVME_SC_ADMIN_CMD_MEDIA_NOT_READY	= 0x2A,
```
igaw, line 6288: These values look wrong. The copy of TP4146 says

- FDP Disabled: 0x29
- Invalid Placement Handle List: 0x2a

And the base spec says for 

- Admin Command Media Not Ready: 0x24

src/nvme/api-types.h
```diff
@@ -199,7 +199,7 @@ struct nvme_format_nvm_args {
  */
 struct nvme_ns_mgmt_args {
 	__u32 *result;
-	struct nvme_id_ns *ns;
+	struct nvme_ns_mgmt_host_sw_specified *data;
```
igaw, line 202: As far I can tell, this breaks the API, although the types have the same types (pointer). The caller might hand in a `ns` pointer and the library expects a `nvme_ns_mgmt_host_sw_specified` type.

src/nvme/types.h
```diff
@@ -7643,4 +7643,95 @@ enum nvme_io_mgmt_send_mo {
...
+ * @zns:      rsvd499( Zoned Namespace Command Set specific field )
+ *        see &struct nvme_ns_mgmt_host_sw_specified_zns.
+ * @phndl:    Placement Handle Associated RUH : This field specifies the Reclaim
+ *        Unit Handle Identifier to be associated with the Placement Handle
+ *        value. If the Flexible Data Placement capability is not supported or
+ *        not enabled in specified Endurance Group, then the controller shall
+ *        ignore this field.
+ * @rsvd768:   Reserved
+ */
+struct nvme_ns_mgmt_host_sw_specified {
```
igaw, line 7712: As far I understood TP4146 extends the existing `struct nvme_id_ns` data structure. That means you should actually update the `struct nvme_id_ns` accordingly, no? That would also solve the API breakage as you don't have to change the type.

igaw (changes requested):
I think the existing `struct nvme_id_ns` should be extended with the new data members from TP4146 and not introducing a new type.

src/nvme/types.h
sc108-lee: Spec mentioned that As "Fields that are not a subset of the Identify Namespace data structure."
That's why I updated as this, do you think it's better to be included in id_ns?

src/nvme/types.h
sc108-lee: mistake(I've pushed wrong branch that was first version of this.) :( 
thanks for the checking, !!

fixed

src/nvme/types.h
sc108-lee: Aligned :)

src/nvme/types.h
igaw: Oh, I missed this part. Need to read up on this. I really hope they didn't come up with polymorphism...

igaw: In the base spec `Figure 300: Namespace Management`,it says that the first 512 bytes are CSI specific (see I/O command set spec). I understand that this is for CSI zns and previous type was for CSI NVM. Anyway, it looks like the type `struct nvme_id_ns` is not completely correct either. It should be something for `Figure 105: Namespace Management – Host Software Specified Fields`. I suspect that `struct nvme_ns_id` was used because it is very close to the type we should use here.

Right, so I suggest that you just extend the `struct nvme_ns_mgmt_args` in a backwards compatible way. I don't think we should fix the existing type at this point for CSI NVM. This boat has sailed.

@birkelund just to keep you in the loop here. You added the other bits for FDP.

sc108-lee: I understand what you mean. (remain API format for backwards compatibility)
Then in this case, What's better ?
1. define struct nvme_ns_mgmt_host_sw_specified and send as (struct nvme_ns_id *)
2. define I/O Command Set specific structure only, and copy to 384 addr on (struct nvme_ns_id *)

igaw: I was thinking on doing something like this here: 9350e73e302a ("libnvme: 64-bit Reference Tags and TP-4068 changes")

igaw (changes requested):
see my inline comments.

src/nvme/api-types.h
```diff
@@ -206,6 +215,22 @@ struct nvme_ns_mgmt_args {
...
+ * enum nvme_ns_mgmt_args_data_offset - Namespace management args data offsets
+ *        in byte size for Create operation 
+ * @NVME_NS_MGMT_NPHNDLS:				Number of Placement Handles
+ * @NVME_NS_MGMT_IO_CMD_SET_SPECIFIC:	io_cmd_set_specific
+ * @NVME_NS_MGMT_PHNDL:					Placement Handles
+ */
+enum nvme_ns_mgmt_args_data_offset {
+	NVME_NS_MGMT_NPHNDLS = 392,
+	NVME_NS_MGMT_IO_CMD_SET_SPECIFIC = 499,
+	NVME_NS_MGMT_PHNDL = 512,
```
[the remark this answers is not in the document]
sc108-lee: I don't understand what you mean.

It's added on pointer of __u8, so at once it move 8bit(1 byte)
NPHNDLS addressed [393:392]Bytes in data fields
pointer moves size of actual data(__u8), so it always move 1byte,
That's why I put (__u8 *)

please let me know if I m wrong.

src/nvme/ioctl.c
```diff
@@ -1237,7 +1237,20 @@ int nvme_ns_mgmt(struct nvme_ns_mgmt_args *args)
...
 	__u32 cdw10    = NVME_SET(args->sel, NAMESPACE_MGMT_CDW10_SEL);
 	__u32 cdw11    = NVME_SET(args->csi, NAMESPACE_MGMT_CDW11_CSI);
-	__u32 data_len = args->ns ? sizeof(*args->ns) : 0;
+	__u32 data_len = 0;
+
+	if (args->ns) {
+		data_len = sizeof(*args->ns);
+		memcpy((__u8 *)args->ns + NVME_NS_MGMT_NPHNDLS,
+			&args->nphndls, sizeof(args->nphndls));
+		if (args->io_cmd_set_specific)
```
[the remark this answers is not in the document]
sc108-lee: so do you recommend using just number instead of using enum?
like below?
		memcpy((__u8 *)args->ns + 392,
			&args->nphndls, sizeof(args->nphndls));

src/nvme/types.h
```diff
@@ -7643,4 +7643,34 @@ enum nvme_io_mgmt_send_mo {
...
+ *	      by the controller.
+ * @rar:       ZNS Requested Active Resources specifies the number of active
+ *	      resources to be allocated to the created namespace.
+ * @ror:       ZNS Requested Open Resources specifies the number of open
+ *	      resources to be allocated to the created namespace.
+ * @rnumzrwa:  ZNS Requested Number of ZRWA Resources specifies the number of
+ *	      ZRWA resources to be allocated to the created namespace.
+ */
+union nvme_ns_mgmt_io_cmd_set_specific {
+	// __u8 rsvd[NVME_NS_MGMT_IO_CMD_SET_SPECIFIC_SIZE];
```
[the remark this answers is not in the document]
sc108-lee: I was thinking that remove rsvd or not, so I forgot to remove this. ( it's 13 bytes )
Any recommandation for this?

only ZNS command set is available for this for the moment.

src/nvme/ioctl.c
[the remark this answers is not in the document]
igaw: The idea is to use `sizeof` to figure out if the user passes in v1 or v2 of the data structure. If it is v2 than you just use the correct data types and avoid using `memcpy` with offsets.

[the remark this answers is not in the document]
sc108-lee: To not doing that, I modified commit as you recommended
Check it out

sc108-lee:
how can I see armhf log?
[cross armhf](https://github.com/linux-nvme/libnvme/actions/runs/4371413890/jobs/7647267054#logs)

igaw:
> how can I see armhf log? [cross armhf](https://github.com/linux-nvme/libnvme/actions/runs/4371413890/jobs/7647267054#logs)

Only the build logs are uploaded if the fail. The test logs are not uploaded. But SIGABT on ARM is almost certainly an invalid load address, e.g. not properly aligned reads or writes.

sc108-lee:
I'm sorry for lots for pushes, I don't have any system to build armhf.
but I don't understand the reason test failed from system. that's why I'm testing it.

I just found out, armhf does not goes into 
   if (args->args_size == size_v2) {

so It's send no data from int nvme_mi_admin_ns_mgmt
that's why the test failed from the system.

Any Ideas?

igaw:
> I'm sorry for lots for pushes, I don't have any system to build armhf.

I've spend the time to figure out this for you. Not cool.
- download a toolchain from https://toolchains.bootlin.com/ (forget the toolchains from your distro, they all suck)
- write a cross file (`~.local/share/meson/cross/armhf.txt`)
```ini
[constants]
tpath = '/home/wagi/work/toolchains/armv7-eabihf--glibc--stable-2022.08-1/'

[binaries]
c = tpath + 'bin/arm-linux-gcc'
ar = tpath + 'bin/arm-linux-ar'
strip = tpath + 'bin/arm-linux-strip'
ld = tpath + 'bin/arm-linux-ld'
exe_wrapper = tpath + 'bin/qemu-arm'

[properties]
sys_root = tpath + 'arm-buildroot-linux-gnueabihf/sysroot'

[host_machine]
system = 'linux'
cpu_family = 'arm'
cpu = 'armv7'
endian = 'little'
```
- Install the qemu binary wrapper (in my case it was in the `qemu-linux-user` package)
- You might need to write a wrapper around qemu as in my case the `sys_root` path was not correctly propagated to qemu (this is the `tpath + 'bin/qemu-arm'` file):
```shell
#!/bin/sh

/usr/bin/qemu-arm -l /home/wagi/work/toolchains/armv7-eabihf--glibc--stable-2022.08-1/arm-buildroot-linux-gnueabihf/sysroot $@
```
- build and run the test
```shell
$ meson setup .build-armhf --cross-file armhf.txt -Dlibnvme:openssl=disabled -Dlibnvme:python=false
$ meson compile -C .build-armhf
$ meson test -C .build-armhf
```

igaw:
> I guess It would be nice to print meson-logs/testlog.txt when test failed
> since not many people has all the systems.

I've updated the workflows but now I run out of time for reviewing your PR.

sc108-lee:
@igaw Thanks a lot for the details.
I've tried some distro and some others (but firewall blocked from company :( )
I will try your recommendation to regenerate issues when I face any later.

igaw (commented):
Almost there. The packed attributes looks wrong though

src/nvme/types.h
```diff
@@ -7643,4 +7643,94 @@ enum nvme_io_mgmt_send_mo {
...
+ *	      to be allocated to the created namespace.
+ * @rnumzrwa:  Requested Number of ZRWA Resources specifies the number of ZRWA
+ *	      resources to be allocated to the created namespace.
+ */
+struct nvme_ns_mgmt_host_sw_specified_zns {
+	__u8			znsco;
+	__u32			rar;
+	__u32			ror;
+	__u32			rnumzrwa;
+} __attribute__((packed));
```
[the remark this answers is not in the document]
sc108-lee: yeah, __le32 is right way to define it. aligned :)

TP4146 - Figure 105: Namespace Management – Host Software Specified Fields
[511:499] Reserved for I/O Command Sets that extend this specification. Refer to the applicable
I/O Command Set specification (e.g., Zoned Namespace Command Set
Specification).

TP4115 - Figure TBDX: Namespace Management – Host Software Specified Fields
Here is definition of struct nvme_ns_mgmt_host_sw_specified_zns
[499] : Zoned Namespace Create Options
[503:500] : Requested Active Resources (RAR)
[507:504] : Requested Open Resources (ROR)
[511:508] : Requested Number of ZRWA Resources (RNUMZRWA)

so I put this structure on struct nvme_ns_mgmt_host_sw_specified as below
	__u8			rsvd394[105];
	union {
		__u8		rsvd499[13];
		struct nvme_ns_mgmt_host_sw_specified_zns zns;
	};
	__le16			phndl[128];

sc108-lee:
Yes I considered this too, but I think there is no other place to use this structure(nvme_ns_mgmt_host_sw_specified_zns) because this structure defined inside of structure.

Since rsvd499(13 byte) would be extended from other I/O Command Sets in future,
I think we need to remain union. 

so how about just move defined structure as left side of your picture to not used anywhere else?

sc108-lee:
To apply TP4146 raised new PR #614.
This thread will remain for TP4115 for current problem of alignment.

igaw:
So the problem with TP4115 it actually cares about the alignment inside of `struct nvme_ns_mgmt_host_sw_specified`. I don't know but maybe we need to expand those types and avoid the union of struct.  E.g.

```c
struct nvme_ns_mgmt_host_sw_specified {
	__le64			nsze;
	__le64			ncap;
	__u8			rsvd16[10];
	__u8			flbas;
	__u8			rsvd27[2];
	__u8			dps;
	__u8			nmic;
	__u8			rsvd31[61];
	__le32			anagrpid;
	__u8			rsvd96[4];
	__le16			nvmsetid;
	__le16			endgid;
	__u8			rsvd104[280];
	__le64			lbstm;
	__le16			nphndls;
	__u8			rsvd394[105];
	union {
		__u8		rsvd499[13];
		struct {
			__u8			znsco;
			__le32			rar;
			__le32			ror;
			__le32			rnumzrwa;
		} __attribute__((packed)) zns;
	};
	__le16			phndl[128];
	__u8			rsvd768[3328];
};
```

This gives me
```
struct nvme_ns_mgmt_host_sw_specified {
        /* typedef __le64 */ long unsigned int          nsze;                            /*     0     8 */
        /* typedef __le64 */ long unsigned int          ncap;                            /*     8     8 */
        /* typedef __u8 */ unsigned char              rsvd16[10];                        /*    16    10 */
        /* typedef __u8 */ unsigned char              flbas;                             /*    26     1 */
        /* typedef __u8 */ unsigned char              rsvd27[2];                         /*    27     2 */
        /* typedef __u8 */ unsigned char              dps;                               /*    29     1 */
        /* typedef __u8 */ unsigned char              nmic;                              /*    30     1 */
        /* typedef __u8 */ unsigned char              rsvd31[61];                        /*    31    61 */
        /* --- cacheline 1 boundary (64 bytes) was 28 bytes ago --- */
        /* typedef __le32 */ unsigned int               anagrpid;                        /*    92     4 */
        /* typedef __u8 */ unsigned char              rsvd96[4];                         /*    96     4 */
        /* typedef __le16 */ short unsigned int         nvmsetid;                        /*   100     2 */
        /* typedef __le16 */ short unsigned int         endgid;                          /*   102     2 */
        /* typedef __u8 */ unsigned char              rsvd104[280];                      /*   104   280 */
        /* --- cacheline 6 boundary (384 bytes) --- */
        /* typedef __le64 */ long unsigned int          lbstm;                           /*   384     8 */
        /* typedef __le16 */ short unsigned int         nphndls;                         /*   392     2 */
        /* typedef __u8 */ unsigned char              rsvd394[105];                      /*   394   105 */
        /* --- cacheline 7 boundary (448 bytes) was 51 bytes ago --- */
        union {
                /* typedef __u8 */ unsigned char      rsvd499[13];                       /*   499    13 */
                struct {
                        /* typedef __u8 */ unsigned char znsco;                          /*   499     1 */
                        /* typedef __le32 */ unsigned int rar;                           /*   500     4 */
                        /* typedef __le32 */ unsigned int ror;                           /*   504     4 */
                        /* typedef __le32 */ unsigned int rnumzrwa;                      /*   508     4 */
                } __attribute__((__packed__)) zns;                                       /*   499    13 */
        };                                                                               /*   499    13 */
        /* --- cacheline 8 boundary (512 bytes) --- */
        /* typedef __le16 */ short unsigned int         phndl[128];                      /*   512   256 */
        /* --- cacheline 12 boundary (768 bytes) --- */
        /* typedef __u8 */ unsigned char              rsvd768[3328];                     /*   768  3328 */

        /* size: 4096, cachelines: 64, members: 19 */
} __attribute__((__packed__));
```

I couldn't get rid of the packed attributed though. So I think we need to do something like this. What do you think?

sc108-lee:
I thought same thing, I would not used for other place.
So I mentioned left side of your pic. (My comment might be ambiguous last time)
if it can be avoid arm problem, I couldn't agree more :)

modified as you mentioned and rebased

Joelgranados:
latest solution LGTM. I have a question on what the original problem was:

I've been trying to reproduce the ARM issue @igaw :
I have cross compiled your (@igaw) example in https://github.com/linux-nvme/libnvme/pull/582#issuecomment-1488171683 and have run it with the setup described in https://github.com/linux-nvme/libnvme/pull/582#issuecomment-1467924872 and everything runs nicely with qemu-arm. Are you seeing a failure/crash?

igaw:
> I've been trying to reproduce the ARM issue @igaw : I have cross compiled your (@igaw) example in [#582 (comment)](https://github.com/linux-nvme/libnvme/pull/582#issuecomment-1488171683) and have run it with the setup described in [#582 (comment)](https://github.com/linux-nvme/libnvme/pull/582#issuecomment-1467924872) and everything runs nicely with qemu-arm. Are you seeing a failure/crash?

Some archs do not support not correctly aligned loads. E.g the load address is  `0x13` and the instruction is `ldw` it will round down load address to `0x10`. See for example Address alignment for word transfers section in https://developer.arm.com/documentation/dui0068/b/ARM-Instruction-Reference/ARM-memory-access-instructions/LDR-and-STR--words-and-unsigned-bytes

qemu is not emulating hardware in all details, I can imagine the arm instruction will be translated to x86 instruction where the load works.

igaw:
Thanks a lot for your patience. This one was a bit tricky and I didn't know the right solution either. But the final result looks right.

Joelgranados:
Thx for taking the time to expand on the solution. It makes sense that it might not emmulate that rounding on the address.

PR #583, opened 2023-02-27
python: Fix segmentation fault during garbage collection

martin-belanger:
With SWIG we create proxy classes to wrap the C structures (struct). These classes are used to create Python objects matching the C structures such as:

`nvme.root` is used to wrap `struct nvme_root`
`nvme.host` is used to wrap `struct nvme_host`
`nvme.ctrl` is used to wrap `struct nvme_ctrl`
etc...

One thing that SWIG cannot do is figure out the dependencies between the different objects. For example, it cannot tell that when deleting a host object that all the subsystems, controllers, namespaces under that host need to be deleted as well. That's an implicit property of the libnvme driver and users need to know to stop using objects after their parent has been deleted.

Unfotunately, with Python the Gargage Collector (GC) decides which objects to delete and it can delete obects in any order it sees fit. This can result in objects being deleted in the wrong order. For example, the GC may delete a host object (`nvme_free_host`) before deleting any of the controller objects under that host. And when the GC finally deletes a controller by calling `nvme_free_ctrl()`, the memory for that controller has already been freed and a segmentation fault will result.

To enforce that objects get deleted in the right order, we need to set dependencies between objects at the Python level. This can be achived by having children objects maintain a reference to their parents. This way a parent cannot be deleted before all its children have been deleted.

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -417,6 +417,9 @@ struct nvme_ns {
   }
 }
 
+%pythonappend nvme_host::nvme_host(struct nvme_root *r, const char *hostnqn,
+				   const char *hostid, const char *hostsymname) {
+    self.__parent = r  # Keep a reference to parent to ensure garbage collection happens in the right order}
 %extend nvme_host {
   nvme_host(struct nvme_root *r, const char *hostnqn = NULL,
 	    const char *hostid = NULL, const char *hostsymname = NULL) {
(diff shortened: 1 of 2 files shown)


igaw:
Thanks a lot!

BTW, is this something we could unit test somehow? I think it would be good to have a test for this.

martin-belanger:
It can be partially tested. Some objects (e.g. controller) require a connection to a subsystem to be fully tested, but I can at least add a test for the root and host objects.

PR #584, opened 2023-02-28
Python test: Check there is no SIGSEGV during garbage collection

martin-belanger:
This is to add a unit test to make sure we don't get a segmentation fault during garbage collection. This is to test that the changes made in PR https://github.com/linux-nvme/libnvme/pull/583 are working as expected and preventing SIGSEGV.

Another commit in this PR makes the `__str__()` methods more Pythonic by returning a `PyObject*` instead of a pointer to a `static char` array.


igaw:
Nice! Thanks a lot!

PR #585, opened 2023-02-28
python: Reformat SWIG input file

martin-belanger:
This is mainly a cosmetic change.

I've been wanting to do this for a while, but could never find the time. The SWIG file contains a lot of C code that was not following the standard formatting (e.g. indentation with tabs). I reformatted the code and moved a few things around to make it more readable.


igaw:
Thanks!

PR #587, opened 2023-03-02
update the Expected Logical Block Application Tag and Reference Tag in nvme copy command

zwxdg:
1、The elbatm is behind elbat fields, whitch are defined in section 3.2.2 Copy command.
 2、Storage and Reference Space is declared as an big-endian order, the elbt[10] fields need to reassigned.

### DIFF
--- a/src/nvme/util.c
+++ b/src/nvme/util.c
@@ -403,15 +403,17 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
 			  __u64 *slbas, __u64 *eilbrts, __u32 *elbatms,
 			  __u32 *elbats, __u16 nr)
 {
-	int i;
+	int i, j;
 
 	for (i = 0; i < nr; i++) {
 		copy[i].nlb = cpu_to_le16(nlbs[i]);
 		copy[i].slba = cpu_to_le64(slbas[i]);
-		copy[i].elbt_l = cpu_to_le64(eilbrts[i]);
-		copy[i].elbt_h = 0;
 		copy[i].elbatm = cpu_to_le16(elbatms[i]);
 		copy[i].elbat = cpu_to_le16(elbats[i]);
+		for (j = 0; j < 8; j++)
+			copy[i].elbt[9 - j] = (eilbrts[i] >> (8 * j)) & 0xff;
+		copy[i].elbt[1] = 0;
+		copy[i].elbt[0] = 0;
 	}  
 }
 
--- a/src/nvme/util.c
+++ b/src/nvme/util.c
@@ -408,7 +408,8 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
 	for (i = 0; i < nr; i++) {
 		copy[i].nlb = cpu_to_le16(nlbs[i]);
 		copy[i].slba = cpu_to_le64(slbas[i]);
-		copy[i].elbt[2] = cpu_to_le64(eilbrts[i]);
+		copy[i].elbt_l = cpu_to_le64(eilbrts[i]);
+		copy[i].elbt_h = 0;
 		copy[i].elbatm = cpu_to_le16(elbatms[i]);
 		copy[i].elbat = cpu_to_le16(elbats[i]);
 	}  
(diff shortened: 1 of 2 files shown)


igaw (changes requested):
- The commit subject should explain what changes and it should use a prefix "types: update ..."
- Please also add a commit message which explains the change. We aim for having all relevant information in the git tree.
- Updating the 'struct nvme_copy_range' struct sound reasonable because it is  a real bug fix and doesn't introduce an API breakage
- Not so sure about `struct nvme_copy_range_f1`. It is breaking API and I haven't really figured out yet if we should this.

igaw:
See also #455

src/nvme/util.c
```diff
@@ -403,14 +403,17 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
...
+	int i, j;
 
 	for (i = 0; i < nr; i++) {
 		copy[i].nlb = cpu_to_le16(nlbs[i]);
 		copy[i].slba = cpu_to_le64(slbas[i]);
-		copy[i].elbt[2] = cpu_to_le64(eilbrts[i]);
 		copy[i].elbatm = cpu_to_le16(elbatms[i]);
 		copy[i].elbat = cpu_to_le16(elbats[i]);
+		for (j = 0; j < 8; j++)
+			copy[i].elbt[9 - j] = (eilbrts[i] >> (8 * j)) & 0xff;
```
bjpaupor, line 414: I think the issue here is that the EILBRT values are stored in big-endian order, like the linked PR which was closed.

Storing them in little-endian order requires the Storage Tag Size and Protection Information Format to determine where the EILBRT should start in the shared range.

zwxdg: Since both EILBRT and ELBST values are store in  big-endian order, and the device software know the Storage Tag Size and what kind of Protection Information Format used, so i think the location of  EILBRT in the shared range is handle by device software, as is the ELBST.
For nvme-cli，we only need to deliver EILBRT and ELBST  in big-end mode to the device.

igaw: > Since both EILBRT and ELBST values are store in big-endian order

Most of the spec is using little endian and I can't find anything EILBRT and ELBST being big endian. The EILBRT and ELBST definition is, let's say a 'challenging' definition. Where do you find the info on big-endian?

zwxdg: image
 in section 5.2.1 Protection Information Formats. for example, the screenshot show the 16b guard protection information format, EILBRT and ELBST re store in big-endian order

igaw: Ah, I didn't look at the Figure :)

I think you should use the be64 helpers as the argument type for eilbert is u64 which is in host endian.

But as I said pointed out, this change breaks the API. We can't just change data structures in this way. This was the reason why I didn't merge #455.

bjpaupor: Yeah I'd thought that the MSB/LSB in that figure might be "Most/Least Significant Bits" rather than defining each of these fields as big-endian byte ordering, particularly based on figure 122 in the next section on the shared space and the consistent mention of significant bits rather than bytes:

image

Definitely see how it could be describing the byte-ordering, but it feels odd without a more explicit call-out like NGUID or EUI64. Worth mentioning this would mean similar ordering changes are likely needed for the shared field and for application tag/guard fields across all IO commands.

igaw: Thanks for the feedback. I looked again with fresh eyes I figured that the github webui confused me. So the two commits should be obviously merged and the final result is just:

```diff
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -4640,8 +4640,8 @@ struct nvme_copy_range {
        __le16                  nlb;
        __u8                    rsvd18[6];
        __le32                  eilbrt;
-       __le16                  elbatm;
        __le16                  elbat;
+       __le16                  elbatm;
 };
 
 /**
@@ -4661,8 +4661,8 @@ struct nvme_copy_range_f1 {
        __le16                  nlb;
        __u8                    rsvd18[8];
        __u8                    elbt[10];
-       __le16                  elbatm;
        __le16                  elbat;
+       __le16                  elbatm;
 };
```

which is okay I think to merge.

PR #591, opened 2023-03-14
fabrics: Add nvmf_hostkey_from_file()

martin-belanger:
The [nvme-cli documentation](https://github.com/linux-nvme/nvme-cli/blob/master/Documentation/nvme-config.txt#L131) mentions that a default Host Key may be defined in /etc/nvme/hostkey. However, there was no function to read that default key. This patch adds the missing function.

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -40,7 +40,29 @@
 		PyDict_SetItemString(p, key, val); /* Does NOT steal reference to val .. */
 		Py_XDECREF(val);                   /* .. therefore decrement ref. count. */
 	}
-%}
+	PyObject *hostnqn_from_file() {
+		char * val = nvmf_hostnqn_from_file();
+		PyObject * obj = PyUnicode_FromString(val);
+		free(val);
+		return obj;
+	}
+	PyObject *hostid_from_file() {
+		char * val = nvmf_hostid_from_file();
+		PyObject * obj = PyUnicode_FromString(val);
+		free(val);
+		return obj;
+	}
+	PyObject *hostkey_from_file() {
+		char * val = nvmf_hostkey_from_file();
+		PyObject * obj = PyUnicode_FromString(val);
+		free(val);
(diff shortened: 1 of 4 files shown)


igaw:
The whole concept of storing a default key in a file is not going to scale well. Also there is sentiment against storing keys in the filesystem.

Also the next thing feature request is already clear: different keys for subsystems.

I just discussed with @hreinecke about the key store topic in the context of TLS. And it turns out it makes sense to use the kernel keyring instead storing the keys in a config file. This would mean maintaining/updating the keys is all done via the kernel keyring and we just reference these keys from the nvme-cli configuration side. 

Basically, we would just provide a key name via

martin-belanger:
Thanks for the update @igaw 

> ... Also there is sentiment against storing keys in the filesystem.

Yep, I believe I mentioned that before... :wink:

When/if we use the kernel keyring, we'll need to document how it's done. I mean, who initializes the keyring? How do keys survive a reboot? etc...

BTW, there are newly released NVMe specs related to authentication (e.g. TP8017, TP8019).  For example, TP8017 talks about authentication for Discovery Controllers (DC). I'm looking for a "default host key" in `nvme-stas` for the case where the DCs are discovered through mDNS (per TP8009). For these DCs, `nvme-stas` only knows their IP address and NQN. It doesn't know which particular Host NQN/ID/Key to use. Therefore, it uses the "default host definitions" located in `/etc/nvme/hostnqn` and `/etc/nvme/hostid`. And that's why I was looking for a default Host key in `/etc/nvme/hostkey`.

martin-belanger:
Closing the PR. I will submit a separate pull request just for the Python stuff. We can revisit using `/etc/nvme/hostkey` as a default Host Key at a later time (i.e. once the whole kernel keyring thing has been defined).

igaw:
Make sense. @hreinecke is working on getting the kernel keyring sorted out, so let's see how this goes first.

PR #598, opened 2023-03-20
types: add opcode field to error log page (TP4113)

minwooim:
Add support for opcode field in error information log page based on TP
4113. It is easy to identify which command is associated with the error.


igaw:
Thanks!

PR #599, opened 2023-03-22
nvme: add 'nvme_insert_tls_key()' function

hreinecke:
Add a function to derive a TLS key and insert it into the given keyring.

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -21,6 +21,11 @@
 #include <openssl/engine.h>
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
+#include <openssl/kdf.h>
+
+#ifdef CONFIG_KEYUTILS
+#include <keyutils.h>
+#endif
 
 #ifdef CONFIG_OPENSSL_3
 #include <openssl/core_names.h>
@@ -470,9 +475,20 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
 	memcpy(key, secret, key_len);
 	return 0;
 }
+
+long nvme_insert_tls_key(const char *keyring, const char *hostnqn,
+			 const char *subsysnqn, int hmac,
+			 unsigned char *configured_key, int key_len)
+{
+	nvme_msg(NULL, LOG_ERR, "TLS key operations not supported; "\
+		 "recompile with OpenSSL support.\n");
+	return -EOPNOTSUPP;
+}
+
 #endif /* !CONFIG_OPENSSL */
 
 #ifdef CONFIG_OPENSSL_1
+#ifdef CONFIG_KEYUTILS
 int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
 			unsigned int key_len, unsigned char *secret,
 			unsigned char *key)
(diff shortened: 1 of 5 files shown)


igaw:
Any chance also to add the OpenSSL v3 version? v1 seems to reach EOL soon

- Version 1.1.1 will be supported until 2023-09-11

https://www.openssl.org/policies/releasestrat.html

src/nvme/linux.c
```diff
@@ -543,6 +559,187 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+		goto out_free_retained_key;
+	if (EVP_PKEY_CTX_add1_hkdf_info(ctx, "tls13 ", 6) <= 0)
+		goto out_free_retained_key;
+	if (EVP_PKEY_CTX_add1_hkdf_info(ctx, "HostNQN", 7) <= 0)
+		goto out_free_retained_key;
+	if (EVP_PKEY_CTX_add1_hkdf_info(ctx, hostnqn, strlen(hostnqn)) <= 0)
+		goto out_free_retained_key;
+
+	if (EVP_PKEY_derive(ctx, retained_key, &retained_len) <= 0) {
+		fprintf(stderr, "EVP_KDF derive failed\n");
```
igaw, line 596: Printing to `stderr` for a library is not ideal. We have `nvme_msg` but without a context it would also print to stderr.  That's why the other OpenSSL code doesn't do any printing at all.

src/nvme/linux.c
```diff
@@ -543,6 +559,187 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+	EVP_PKEY_CTX_free(ctx);
+out_free_psk:
+	free(psk);
+out_free_identity:
+	free(psk_identity);
+
+	return key;
+}
+
+long nvme_insert_tls_key(const char *keyring, const char *hostnqn,
```
igaw, line 689: `int` return type?

src/nvme/linux.c
```diff
@@ -543,6 +559,187 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+	case 1:
+		md = EVP_sha256();
+		break;
+	case 2:
+		md = EVP_sha384();
+		break;
+	default:
+		fprintf(stderr, "Invalid key hash %df\n", hmac);
+		errno = EINVAL;
+		return -1;
```
igaw, line 707: Return 1 on error.

src/nvme/linux.c
```diff
@@ -543,6 +559,187 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+	if (!retained_key) {
+		fprintf(stderr, "Failed to derive retained key\n");
+		errno = EINVAL;
+		return -1;
+	}
+	key = derive_tls_key(md, keyring_id, hostnqn,
+			     subsysnqn, retained_key, key_len);
+	if (key < 0) {
+		fprintf(stderr, "Failed to derive TLS key\n");
+		errno = (int)key;
```
igaw, line 727: Not sure about returning anything else than a E code in errno is really good idea in IMO.

src/nvme/linux.c
```diff
@@ -470,9 +475,20 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
 	return 0;
 }
+
+long nvme_insert_tls_key(const char *keyring, const char *hostnqn,
+			 const char *subsysnqn, int hmac,
+			 unsigned char *configured_key, int key_len)
+{
+	nvme_msg(NULL, LOG_ERR, "TLS key operations not supported; "\
+		 "recompile with OpenSSL support.\n");
+	return -EOPNOTSUPP;
```
igaw, line 485:
```c
errno=-EOPNOTSUPP;
return 1;
```

igaw: Just noticed that we are using -1 as error code in this part of the API and the rest of the library 1. We really should ditch the `errno/return 1` stupidity in future.

src/nvme/linux.c
hreinecke: Will be replacing it with nvme_msg() or removing it, as appropriate.

src/nvme/linux.c
hreinecke: Weird keyutils semantic. What I'm actually want is to return a key serial (which is defined as 'long'), but returning a type 'key_serial_t' would give us a hit on keyutils.h, and it won't be optional anymore.
But I'll check if I can make the call to the function optional, then it should be possible to return a key_serial_t here.

src/nvme/linux.c
hreinecke: That's on purpose; I want to return the key serial from nvme_insert_tls_key() such that the caller knows which key has been generated (if he cares to look). So I'd rather change the convention to return '0' if we failed, and the actual key serial number on success.

src/nvme/linux.c
hreinecke: I've now changed the function to return the key serial (of type 'long') or '0' on failure.

Issue #448, opened 2022-07-26
Add endpoint data structure for nvme API


igaw:
The MI API uses the endpoint data struct to abstracts from implementation details. Introduce an similiar API for the nvme interface. The goal here is to decouple from the low level API and be able to mock certain APIs for testing.

jk-ozlabs:
It would be great if we can use a common struct between the direct and MI interfaces - this means we could avoid duplicating the API for each transport.

I'm happy to propose something for this, if you like.

igaw:
Sorry, for the late response. Public holiday :)

Let me have a look at it now. IIRC this work was depending on other work? Anyway, I'l review it now

jk-ozlabs:
Yes, this would be a fairly big change - it would alter the arguments for almost every function. Definitely something for a ABI-breaking milestone.

I'll put together some sketches for this.

igaw:
I've started to work on libnvme 2. Here are the first bits an early RFC so please have a look and let me know if this is going into the right direct. @jk-ozlabs 

https://github.com/nvme-experiments/libnvme/commits/passthru/

igaw:
> Looks good! I would see the next step as unifying the `nvme_passthru_dev`:
> 
> struct nvme_passthru_dev {
> 	int fd;
> };
> 
> into a structure that contains both transport types (ie, ioctl and MI); that would mean we could remove all of the specialisation between transports, and just have one set of functions for all¹ commands. Is that what you're intending here?

Yes, that was my idea. First, I wanted to pass in `nvme_root_t` around but then I figured this wont work, because we have a device for each `ctrl` and `ns` in `tree.c`. So simplest thing is to introduce an abstraction just for the file descriptor. My idea with the `nvme_root_t` was to be able to have a logging context but we can easily also extend the `struct nvme_passthru_dev` to be able to carry  around the logging context. This is for later though.

> Also, any reason it's named 'passthru' ? isn't this primarily the structure that represents our communication facility with the device?

No particular reason other than I wanted to name it clearly what it supposed to do. Though `struct nvme_dev` is nicer after all :)

> 1: We'd still need spearate MI functions for the MI-specific commands, of course

Yes, next step. 

Adding @keithbusch @hreinecke so they are aware what's going on.

jk-ozlabs:
> I've updated libnvme-mi to use nvme_dev_t instead of nvme_mi_ctrl_t. There is already a nvme_ctrl_t so we can't use it for the transport abstraction.

[...]

PR #600, opened 2023-03-23
RFC: xnvme backend support

vikash-k5:
Extend the api-types with the following device-handle types:

* struct dev_handle
* enum nvme_dev_type

The 'dev_handle' provides a, possibly opaque, device handle instead of a
"fixed" file-descriptor. This allows for non-OS managed device-types
such as user-space NVMe-drivers. Additionally, the types allows for
library-side dispatch.

This is done in preparation for xNVMe, and thereby, support for
user-space NVMe driver, io_uring_cmd etc.

This provides the means to use nvme-cli/libnvme via user-space NVMe
drivers, io_uring command on Linux. As well as using nvme-cli natively
on FreeBSD. This is done by dispatching commands to xNVMe.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -61,6 +61,9 @@ else
 endif
 conf.set('CONFIG_JSONC', json_c_dep.found(), description: 'Is json-c required?')
 
+# Check for libxnvme availability
+libxnvme_dep = dependency('xnvme', version: '>=0.2.0', required: true)
+
 # Check for OpenSSL availability
 openssl_dep = dependency('openssl',
                          version: '>=1.1.0',
--- a/src/meson.build
+++ b/src/meson.build
@@ -32,6 +32,7 @@ endif
 deps = [
     json_c_dep,
     openssl_dep,
+    libxnvme_dep,
 ]
 
 mi_deps = [
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -83,13 +83,13 @@ int nvme_scan_topology(struct nvme_root *r, nvme_scan_filter_t f, void *f_args)
 
 	if (!r)
 		return 0;
-
 	num_ctrls = nvme_scan_ctrls(&ctrls);
 	if (num_ctrls < 0) {
 		nvme_msg(r, LOG_DEBUG, "failed to scan ctrls: %s\n",
 			 strerror(errno));
 		return num_ctrls;
 	}
+
 
 	for (i = 0; i < num_ctrls; i++) {
 		nvme_ctrl_t c = nvme_scan_ctrl(r, ctrls[i]->d_name);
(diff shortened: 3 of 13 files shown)


src/nvme/tree.h
```diff
@@ -469,7 +469,7 @@ nvme_ns_t nvme_subsystem_next_ns(nvme_subsystem_t s, nvme_ns_t n);
  *
  * Return: File descriptor associated with @n or -1
  */
-int nvme_ns_get_fd(nvme_ns_t n);
+void* nvme_ns_get_fd(nvme_ns_t n);
```
igaw, line 472: The function name should also be updated alongside. Also I'd like to see a proper type returned as handle.

test/test.c
```diff
@@ -38,7 +38,8 @@ static int test_ctrl(nvme_ctrl_t c)
 	static __u8 buf[0x1000];
 
 	enum nvme_get_features_sel sel = NVME_GET_FEATURES_SEL_CURRENT;
-	int ret, temp, fd = nvme_ctrl_get_fd(c);
+	int ret, temp; 
+	struct dev_handle *hnd = nvme_ctrl_get_fd(c);
```
igaw, line 42: I prefer `hdl` as short hand for handle, but I don't have a strong opinion on this.

src/meson.build
```diff
@@ -32,6 +32,7 @@ endif
 deps = [
     json_c_dep,
     openssl_dep,
+    libxnvme_dep,
```
igaw, line 35: Without introducing the dependency object `libxnvme_dep ` this will not work.

src/nvme/tree.c
```diff
@@ -971,9 +972,7 @@ static void __nvme_free_ctrl(nvme_ctrl_t c)
 {
 	struct nvme_path *p, *_p;
 	struct nvme_ns *n, *_n;
-
```
igaw, line 974: Please avoid changing white space changes here (btw, this new line here is okay)

igaw (changes requested):
Yes this make sense for libnvme v2

igaw:
Also please update the MI interfaces accordingly, see https://github.com/linux-nvme/libnvme/issues/448

meson.build
```diff
@@ -61,6 +61,9 @@ else
 endif
 conf.set('CONFIG_JSONC', json_c_dep.found(), description: 'Is json-c required?')
 
+# Check for libxnvme availability
+libxnvme_dep = dependency('xnvme', version: '>=0.2.0', required: true)
```
tbzatek, line 65: Please make this dependency optional, this might pull in whole set of secondary dependencies that might not be suitable for some systems. Think of initramfs usage, different init systems (linux is not only about systemd) and constrained embedded environments.

src/nvme/tree.c
```diff
@@ -353,9 +353,11 @@ nvme_path_t nvme_namespace_next_path(nvme_ns_t ns, nvme_path_t p)
 static void __nvme_free_ns(struct nvme_ns *n)
 {
 	list_del_init(&n->entry);
-	close(n->fd);
+	struct dev_handle *hnd = n->hnd;
```
tbzatek, line 356: Declarations mixed inside the code.

vikash-k5:
Created a new PR addressing review comments.

PR #601, opened 2023-03-23
Patches for TLS support

hreinecke:
Here are patches for the upcoming TLS support for NVMe/TCP.
The first patch is function to store a TLS PSK in the specified keyring, and the other patches add support for the fabrics options 'keyring' and 'tls_key'.
Idea of these options are that the 'keyring' option references a keyring which holds all valid keys for this connection; the kernel code will then look up keys from this keyring and start the TLS connection with these keys.
The 'tls_key' option specifies a single key with which the kernel should start the TLS connection.

### DIFF
--- a/doc/config-schema.json
+++ b/doc/config-schema.json
@@ -94,6 +94,10 @@
 		    "description": "Controller DH-HMAC-CHAP key",
 		    "type": "string"
 		},
+		"keyring": {
+		    "description": "Keyring to store and lookup keys",
+		    "type": "string",
+		},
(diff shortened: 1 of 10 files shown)


hreinecke:
Superseded by PR #602

igaw:
You can also update your existing PR with 'git push -f'. This avoid opening new PRs all the time. Though it also kills the previous version and the UI gets easily confused with forced updates. Anyway, just saying if you want to avoid opening new PRs all the time, there is a way  :)

PR #602, opened 2023-03-23
Patches for TLS support (v2)

hreinecke:
Here's an updated version of my patches for the upcoming TLS support for NVMe/TCP.
The first patch is function to store a TLS PSK in the specified keyring, and the other patches add support for the fabrics options 'keyring' and 'tls_key'.
Idea of these options are that the 'keyring' option references a keyring which holds all valid keys for this connection; the kernel code will then look up keys from this keyring and start the TLS connection with these keys.
The 'tls_key' option specifies a single key with which the kernel should start the TLS connection.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -98,6 +98,15 @@ endif
 conf.set('CONFIG_OPENSSL', openssl_dep.found(),
          description: 'Is OpenSSL/LibreSSL available?')
 
+if get_option('keyutils').disabled()
+    keyutils_dep = dependency('', required: false)
+else
+    keyutils_dep = dependency('libkeyutils',
+                              required : get_option('keyutils'))
+endif
+conf.set('CONFIG_KEYUTILS', keyutils_dep.found(),
+         description: 'Is libkeyutils available?')
+
 if get_option('libdbus').disabled()
     libdbus_dep = dependency('', required: false)
 else
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+LIBNVME_1_4 {
+	global:
+		nvme_lookup_keyring;
+};
+
 LIBNVME_1_3 {
 	global:
 		nvme_ctrl_is_unique_discovery_ctrl;
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -3,6 +3,9 @@
 LIBNVME_1_4 {
 	global:
 		nvme_lookup_keyring;
+		nvme_describe_key_serial;
+		nvme_lookup_key;
+		nvme_set_keyring;
 		nvme_insert_tls_key;
 };
 
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -3,6 +3,7 @@
 LIBNVME_1_4 {
 	global:
 		nvme_lookup_keyring;
+		nvme_insert_tls_key;
 };
 
 LIBNVME_1_3 {
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -21,6 +21,7 @@
 #include <openssl/engine.h>
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
+#include <openssl/kdf.h>
 
 #ifdef CONFIG_OPENSSL_3
 #include <openssl/core_names.h>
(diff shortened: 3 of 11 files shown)


meson.build
```diff
@@ -93,6 +93,10 @@ if openssl_dep.found()
            description: 'OpenSSL/LibreSSL API version @0@'.format(api_version))
 endif
 
+keyutils_dep = dependency('libkeyutils',
+                          required : get_option('keyutils'))
```
igaw, line 97: We should the same approach as we have done with libdbus. Introduce a meson config and make the dependency really optional. Unfortunately, the required argument of dependency doesn't allow to drop the dependency completely.

src/nvme/linux.c
```diff
@@ -474,6 +475,143 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+	if (EVP_PKEY_CTX_add1_hkdf_info(ctx,
+					(const unsigned char *)hostnqn, strlen(hostnqn)) <= 0)
+		goto out_free_retained_key;
+
+	if (EVP_PKEY_derive(ctx, retained_key, &retained_len) > 0)
+		err = 0;
+
+out_free_retained_key:
+	if (err) {
+		nvme_msg(NULL, LOG_DEBUG,
```
igaw, line 527: Couldn't just make the function to return the error code and the key is part of the argument list?

Because printing to stderr in the library should be avoided unless very very important.

I understand these errors here could happen during normal operation, so we really should return proper error codes which can be presented to the user.

src/nvme/linux.h
```diff
@@ -205,4 +205,24 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+ * @configured_key:	Configured key data to derive the key from
+ * @key_len:	Length of @configured_key
+ *
+ * Derives a 'retained' TLS key as specified in NVMe TCP 1.0a and
+ * stores it as type @key_type in the keyring specified by @keyring.
+ *
+ * Return: The key serial number if the key could be inserted into
+ * the keyring or 0 otherwise.
+ */
+long nvme_insert_tls_key(const char *keyring, const char *key_type,
```
igaw, line 260: As said previously, we should probably return error codes so that application can print/display some meaningful error message and not just 'operation failed'.

src/nvme/linux.c
```diff
@@ -656,11 +794,69 @@ long nvme_lookup_keyring(const char *keyring)
...
 	return 0;
 }
+
+long nvme_insert_tls_key(const char *keyring, const char *key_type,
+			 const char *hostnqn, const char *subsysnqn, int hmac,
+			 unsigned char *configured_key, int key_len)
+{
+	if (derive_nvme_keys(NULL, NULL, 0, NULL, 0))
+		return 0;
+	return 0;
```
igaw, line 912: Twice `return 0`

src/nvme/linux.c
```diff
@@ -795,6 +795,39 @@ long nvme_lookup_keyring(const char *keyring)
...
+	return desc;
+}
+
+long nvme_lookup_key(const char *type, const char *identity)
+{
+	key_serial_t key;
+
+	key = keyctl_search(KEY_SPEC_SESSION_KEYRING, type, identity, 0);
+	if (key < 0) {
+		nvme_msg(NULL, LOG_ERR,
```
igaw, line 813: `keyctl_search` seems to set `errno`, I would just document this and don't print anything.

src/nvme/linux.h
```diff
@@ -205,6 +205,42 @@ int nvme_gen_dhchap_key(char *hostnqn, enum nvme_hmac_alg hmac,
...
+ * nvme_describe_key_serial() - Return key description
+ * @key_id:    Key serial number
+ *
+ * Fetches the description of the key or keyring identified
+ * by the serial number @key_id.
+ *
+ * Return: The description of @key_id or NULL on failure.
+ * The returned string needs to be freed by the caller.
+ */
+char *nvme_describe_key_serial(long key_id);
```
igaw, line 218: It seems that the `keyctl` library sets `errno` on error so we should just forward it and document it here (and the following functions).

igaw (changes requested):
I think the only real issue I have is the error handling. We should stick to setting `errno` to make it a consistent and also make sure to forward all error code. This avoids the need to print anything to stderr.

The last two patches could properly merged.

Overall, I think we almost there.

BTW, I've scheduled a release Friday next week. I think we can get this sorted out till then and get this new feature to all devs quickly. We could mark it as experimental if we are not really sure with the API yet.

meson.build
hreinecke: I don't mind. Being as I am no expert on meson I'd go with whatever the recommendation is.

src/libnvme.map
```diff
@@ -10,6 +10,7 @@ LIBNVME_1_3 {
 		nvme_io_mgmt_send;
 		nvme_host_is_pdc_enabled;
 		nvme_host_set_pdc_enabled;
+		nvme_lookup_keyring;
```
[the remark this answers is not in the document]
hreinecke: Sure. Just put it there as I didn't know how you'd like to handle it.

src/nvme/linux.c
hreinecke: Nasty dummy functions. Probably time for the __unused__ attribute.

src/nvme/linux.c
hreinecke: Sure.

src/nvme/linux.h
hreinecke: Sure.

meson.build
igaw: I got a beating for introducing the libdbus dependency. So I am a bit more careful what is necessary :)

src/nvme/linux.c
igaw: Indeed error handling is ugly in C and it leads to compromises in the API design.

src/nvme/linux.h
hreinecke: Done with the update.

src/nvme/linux.c
hreinecke: Done with the update.

src/nvme/linux.c
hreinecke: Removed with the update.

src/nvme/linux.h
hreinecke: Done with the update.

src/nvme/linux.c
hreinecke: Reworked functions to return 0/-1 and setting errno.

src/libnvme.map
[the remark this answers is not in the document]
hreinecke: Moved to new 1.4 section.

meson.build
hreinecke: Updated the dependencies; do check whether it's okay now.

meson.build
```diff
@@ -93,6 +93,14 @@ if openssl_dep.found()
            description: 'OpenSSL/LibreSSL API version @0@'.format(api_version))
 endif
 
+if get_option('keyutils').disabled()
+    keyutils_dep = dependency('', required: false)
+else
+    keyutils_dep = dependency('libkeyutils',
+                              required : true)
```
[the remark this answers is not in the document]
hreinecke: Indeed.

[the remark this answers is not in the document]
martin-belanger: There's another option (but I'm not sure if Daniel will agree with this approach for your use-case :wink:). 

Check what I did for `python`. In `meson_options.txt` I define the `python` option as follows with `auto` being the default:
```
option('python', type : 'combo', choices : ['auto', 'true', 'false'], description : 'Generate libnvme python bindings')
```

- If `python==auto`, then we check whether the dependencies (swig, etc.) are present and if they are, then we build the python wrappers. If the deps are not found, then we simply skip building python.
- If `python==false`, then we don't build the python wrappers (even if dependencies are present).
- If `python==true`, then we must build the python wrappers and the dependencies are `required`.

Here's what it looks like in [libnvme/meson.build](https://github.com/linux-nvme/libnvme/blob/master/libnvme/meson.build#L8):
```
want_python = get_option('python')
if want_python != 'false'
    python3 = import('python').find_installation('python3')
    py3_dep = python3.dependency(required: want_python == 'true')
    swig = find_program('swig', required: want_python == 'true')
    header_found = cc.has_header('Python.h', dependencies: py3_dep, required: want_python == 'true')
    have_python_support = py3_dep.found() and swig.found() and header_found
else
    have_python_support = false
endif
```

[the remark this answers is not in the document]
martin-belanger: Sounds good. Let me experiment with that. 

The problem with Python is that it requires multiple dependencies:
- `python3`: The Python interpreter
- `swig`: An executable/program
- `Python.h`: A header file indicating that the C-Python framework is installed

We need all 3 to be present in order to build the Python bindings.

igaw:
- Only a couple of missing errno setting added.
- Update the documentation that errno is set
- Fixed one return erro check (function return 0 on error not -1 in `nvme_insert_tls_key()`

igaw (approved):

PR #603, opened 2023-03-24
tree: fix generic device open failure

minwooim:
If scan_namespace is called with a generic device (e.g., /dev/ng0n1) given, it fails to scan a namespace based on the generic device because there's no entry point in /sys/block/ for the generic device.

This patch provides two helpers to change the given generic device name to a block device name based on the instances:
	ng0n1 -> nvme0n1

This patch fixes command failure:
	root@vm:~/work/nvme-cli.git# nvme show-regs /dev/ng0n1
	Unable to find ng0n1
	get-property: Invalid argument

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -1864,27 +1864,62 @@ free_ns:
 	return NULL;
 }
 
+static inline bool nvme_ns_is_generic(const char *name)
+{
+	int instance, head_instance;
+
+	if (sscanf(name, "ng%dn%d", &instance, &head_instance) != 2)
+		return false;
+	return true;
+}
+
+static char *nvme_ns_generic_to_blkdev(const char *generic)
+{
+
+	int instance, head_instance;
+	char blkdev[PATH_MAX];
+
+	if (!nvme_ns_is_generic(generic))
+		return strdup(generic);
+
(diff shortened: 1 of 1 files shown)


src/nvme/tree.c
```diff
@@ -1864,19 +1864,45 @@ static nvme_ns_t nvme_ns_open(const char *name)
...
+
 static struct nvme_ns *__nvme_scan_namespace(const char *sysfs_dir, const char *name)
 {
 	struct nvme_ns *n;
 	char *path;
 	int ret;
+	char *blkdev;
 
-	ret = asprintf(&path, "%s/%s", sysfs_dir, name);
+	blkdev = (char *)nvme_ns_generic_to_blkdev(name);
```
igaw, line 1898: `blkdev` is leaked here if it's it was a generic device name. I would just update the function signature to return 'char *' and do a `strdup` on generic as well and also  free `blkdev` after the `asprintf` call.

igaw (changes requested):

src/nvme/tree.c
minwooim: Sure, let me update. Thanks!

minwooim: Updated.

igaw (approved):

PR #605, opened 2023-03-24
python: Change python option from combo to feature

martin-belanger:
@igaw - Per our earlier [discussion](https://github.com/linux-nvme/libnvme/pull/602#discussion_r1147668694), changing `python` option from a `combo` to a `feature`. 

Thanks for pointing me to `feature`. It is a bit cleaner than using a `combo` and it aligns with the rest of the options.


igaw:
Thanks. That reminds me to update the build how-to documentation...

PR #610, opened 2023-03-30
Validate DMI UUID

### LINKED ISSUE #568: nvme gen-hostnqn generates invalid UUID
@maurizio-lombardi  reports: " This is due to the fact that the dmi entry used to generate the hostnqn contains only zeros and
libnvme doesn't check for this case."

See also:
http://bsdbackstore.eu/misc/libnvme_zero_uuid.patch

igaw:
fixes #568

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1018,6 +1018,25 @@ static int uuid_from_device_tree(char *system_uuid)
  */
 #define DMI_SYSTEM_INFORMATION	1
 
+static bool is_dmi_uuid_valid(const char *buf, size_t len)
+{
+	int i;
+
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -1018,6 +1018,25 @@ static int uuid_from_device_tree(char *system_uuid)
...
+static bool is_dmi_uuid_valid(const char *buf, size_t len)
+{
+	int i;
+
+	/* UUID bytes are from byte 8 to 23 */
+	if (len <= 23)
+		return false;
+
+	/* Test it's a invalid UUID with all zeros */
+	for (i = 8; i < 23; i++) {
```
igaw, line 1030: the upper limit is 24 not 23.

PR #622, opened 2023-04-13
Python: Add NBFT support

martin-belanger:
This PR adds Python wrapper code to allow reading NBFT data. I also added a unit test for it.


tbzatek:
I think we can afford having both test suites in place. Thinking ahead - `nvme nbft show --output-format=json --nbft-path=<PATH>` can be used as a third test over the same collection of tables.

igaw:
Thanks!

PR #623, opened 2023-04-13
nbft: Move added symbols to LIBNVME_1_5

tbzatek:
Now that the v1.4 release is out I assume the NBFT stuff is going to v1.5.


igaw:
Thanks!

PR #624, opened 2023-04-13
nbft: Add simple unit test

tbzatek:
A simple table dump utility, a set of real ACPI NBFT table files and corresponding set of reference dumps, compared against each other as part of the meson test run.

Please check the README file for details.


test/nbft/nbft-dump.c
```diff
@@ -0,0 +1,121 @@
...
+	unsigned int i, j;
+	struct nbft_info_hfi **hfi;
+	struct nbft_info_security **sec;
+	struct nbft_info_discovery **disc;
+	struct nbft_info_subsystem_ns **ssns;
+
+	printf("raw_nbft_size=%zd\n", table->raw_nbft_size);
+
+	printf("host.id=");
+	print_hex(table->host.id, 16);
```
martin-belanger, line 33: Just nitpicking... Instead of `16` you could use `NVME_UUID_LEN` (but since this is just test code it's not really necessary).

PR #631, opened 2023-04-20
Add 'application' strings to nvme_root and nvme_subsystem

hreinecke:
This patchset adds an 'application' string to both, the nvme root and the nvme subsystem.
With that an application can restrict lookup to subsystems which are managed  by that application.
For any application not setting the string during initialisation (ie does not call `nvme_root_set_application()`)
lookup will ignore subsystems which have an application string set.
Application strings are automatically set when creating new subsystems.

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -2,9 +2,11 @@
 
 LIBNVME_1_5 {
 	global:
-		nvme_nbft_read;
-		nvme_nbft_free;
 		nvme_ipaddrs_eq;
+		nvme_nbft_free;
+		nvme_nbft_read;
+		nvme_subsystem_get_application;
+		nvme_subsystem_set_application;
 };
 
 LIBNVME_1_4 {
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -5,6 +5,8 @@ LIBNVME_1_5 {
 		nvme_ipaddrs_eq;
 		nvme_nbft_free;
 		nvme_nbft_read;
+		nvme_root_get_application;
+		nvme_root_set_application;
 		nvme_subsystem_get_application;
 		nvme_subsystem_set_application;
 };
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -819,6 +819,7 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
 		  const struct nvme_fabrics_config *cfg)
 {
 	nvme_subsystem_t s;
+	const char *root_app, *app;
 	char *argstr;
 	int ret;
 
@@ -853,6 +854,23 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
 				nvme_ctrl_set_dhchap_key(c, key);
 		}
 
+	}
+
+	root_app = nvme_root_get_application(h->r);
+	app = nvme_subsystem_get_application(s);
+	if (root_app) {
+		/*
+		 * configuration is managed by an application,
+		 * refuse to act on subsystems which either have
+		 * no application set or which habe a different
+		 * application string.
+		 */
+		if (!app || strcmp(app, root_app)) {
+			nvme_msg(h->r, LOG_INFO, "skip %s, not managed by %s\n",
+				 nvme_subsystem_get_nqn(s), root_app);
+			errno = ENVME_CONNECT_INVAL;
+			return -1;
+		}
 	}
 
 	nvme_ctrl_set_discovered(c, true);
--- a/doc/config-schema.json.in
+++ b/doc/config-schema.json.in
@@ -59,6 +59,10 @@
 		    "type": "array",
 		    "items": { "$ref": "#/$defs/port" }
 		},
+		"application": {
+		    "description": "Program managing this subsystem",
+		    "type": "string"
+		},
 		"required": [ "nqn" ]
 	    }
 	},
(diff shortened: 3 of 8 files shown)


hreinecke:
This patchset is intended to clean up interaction between nvme-cli / udev autoconnect and nvme-stas.
Intention is that 'nvme-stas' calls `nvme_root_set_application(r, "nvme-stas")` during initialisation, and any subsystems which should be managed by nvme-stas should set the application string to 'nvme-stas', too.
With that nvme-cli should ignore all subsystems for which are manged by nvme-stas, and nvme-stas will only see the subsystems managed by itself.

martin-belanger:
Hi @hreinecke - Will the `application` string get saved somewhere in the file system? In other words, after connections to controllers have been established by an application, how will other applications know who created that connection?

hreinecke:
> Hi @hreinecke - Will the `application` string get saved somewhere in the file system? In other words, after connections to controllers have been established by an application, how will other applications know who created that connection?

It's all saved in the json configuration file. And as nvme-stas is using libnvme (and hence reads in the configuration file as well) we should be able to synchronize easily.

hreinecke:
That is, the application string for the subsystem is saved in the configuration file.
The application string for the nvme_root object needs to be set by the application on startup.

martin-belanger:
A lot of connections made by nvme-stas are not created from the json file. Nvme-stas discovers Discovery Controllers (DC) using mDNS and connect to them (those DCs are not in the json file). We then connect to the I/O Controllers (IOC) listed in the discovery log pages (those IOCs are not in the json file either).

hreinecke:
Hmm. Would it be possible to write out the json configuration file (ie call `nvme_save_config()`) after the DC has been read from the mDNS records? Then we would have the correct 'application' tag for the discovery controller, and we could expand the matching algorithms to apply the application tag to all subsystems listed in that discovery log page ...
Hmm.

martin-belanger:
> Hmm. Would it be possible to write out the json configuration file (ie call `nvme_save_config()`) after the DC has been read from the mDNS records? Then we would have the correct 'application' tag for the discovery controller, and we could expand the matching algorithms to apply the application tag to all subsystems listed in that discovery log page ... Hmm.

Yep. That's what I've been thinking too. I'm also thinking that in addition to the persistent json file under `/etc/nvme/`, maybe we should have a volatile file (e.g. `/run/nvme/config.json`). The reasoning being that the configuration collected from mDNS and the CDC should not be persisted through reboots. In other words, on a hard reboot we want to start fresh and relearn everything from the network and not use some (potentially) stale config.

Another thing that has been on my mind regarding the json configuration, is that we can't currently define a default Host NQN and ID. I know that we define defaults in `/etc/nvme/[hostnqn|hostid]`. I was just wondering if we should migrate everything to json. Default Host parameters are needed for those DCs discovered through mDNS. Can nvme-stas continue using `/etc/nvme/[hostnqn|hostid]` or should we define the defaults in the json file so that everything can be found in one place?

hreinecke:
> Another thing that has been on my mind regarding the json configuration, is that we can't currently define a default Host NQN and ID. I know that we define defaults in `/etc/nvme/[hostnqn|hostid]`. I was just wondering if we should migrate everything to json. Default Host parameters are needed for those DCs discovered through mDNS. Can nvme-stas continue using `/etc/nvme/[hostnqn|hostid]` or should we define the defaults in the json file so that everything can be found in one place?

Hmm? We certainly can; config.json has several 'host' sections, each with a different hostnqn/hostid.
Which to select, though, to create new controllers is up to the application.
Maybe we should have application-specific NQN ...

src/libnvme.map
```diff
@@ -1,5 +1,11 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+LIBNVME_1_6 {
+	global:
+		nvme_subsystem_get_application;
+		nvme_subsystem_set_application;
+};
```
igaw, line 9: v1.5 is not yet released :)

src/nvme/fabrics.c
```diff
@@ -811,6 +812,23 @@ int nvmf_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
...
+	root_app = nvme_root_get_application(h->r);
+	app = nvme_subsystem_get_application(s);
+	if (root_app) {
+		/*
+		 * configuration is managed by an application,
+		 * refuse to act on subsystems which either have
+		 * no application set or which habe a different
+		 * application string.
+		 */
+		if (!app || strcmp(app, root_app)) {
```
igaw, line 824: For blktests the calling application is nvme-cli itself via udev. So should we start trying to figure in which context nvme-cli is started or should just have a way to say ignore subsys with a given name should be ignored?

igaw: Basically the problem is how to figure out the calling context for nvme-cli to get this working here.

src/libnvme.map
igaw: FWIW, I've fixed this up.

igaw:
> I'm also thinking that in addition to the persistent json file under `/etc/nvme/`, maybe we should have a volatile file (e.g. `/run/nvme/config.json`). The reasoning being that the configuration collected from mDNS and the CDC should not be persisted through reboots

Is the idea that `/run/nvme/config.json` file is under control of an external program? Just wondering how to serialize write access to this file. 

Note: I like the idea :)

igaw:
Just rebased.

Let's add the simple filter feature. I look into adding the volatile file (e.g. /run/nvme/config.json) feature. This something I could use with blktests as well.

PR #635, opened 2023-04-24
examples: fix incorrect controller status in MI info output

wlbe4:
In the mi-mctp example, we're incorrectly reporting the percent drive life used as the controller status. Fix the controller status output to use the correct (ccs) field.


birkelund:
LGTM.

jk-ozlabs:
LGTM, thanks.

igaw:
Thanks!

Issue #634, opened 2023-04-24
swig warning during libnvme build


martin-gpy:
Seeing the below warning during libnvme build on a SLES15 SP4 MU host:

```
# ninja -C .build
ninja: Entering directory `.build'
[46/59] Generating nvme.py with a custom command
notype name zns
../src/nvme/types.h:7727: Warning 503: Can't wrap class $unnamed7$ unless renamed to a valid identifier.
[59/59] Linking target libnvme/_nvme.cpython-36m-x86_64-linux-gnu.so
```
```
# rpm -qa | grep gcc
gcc-7-3.9.1.x86_64
gcc-c++-7-3.9.1.x86_64
libgcc_s1-11.2.1+git610-150000.1.6.6.x86_64
libstdc++6-pp-gcc11-11.2.1+git610-150000.1.6.6.x86_64
libstdc++6-devel-gcc7-7.5.0+r278197-4.30.1.x86_64
gcc7-7.5.0+r278197-4.30.1.x86_64
gcc7-c++-7.5.0+r278197-4.30.1.x86_64
```

martin-belanger:
Yep. Saw that too. Seems to be a bug with swig. We've configured swig to ignore structs, classes, unions, etc., but it still complains about anonymous structs, classes, unions, etc. This is reference to an anonymous union here: https://github.com/linux-nvme/libnvme/blob/master/src/nvme/types.h#L7725

@igaw - would it be OK if we added a name to this union? We may have to add a name to the struct inside as well.

martin-belanger:
Submitted https://github.com/linux-nvme/libnvme/pull/636

martin-gpy:
@martin-belanger, the swig 503 warning no longer shows up now with your change, but I still see 'notype name zns' displayed during the build:

```
# ninja -C .build
ninja: Entering directory `.build'
[47/59] Generating nvme.py with a custom command
notype name zns
[59/59] Linking target libnvme/_nvme.cpython-36m-x86_64-linux-gnu.so
```

tbzatek:
I guess it really depends on swig version: `swig-4.1.1-4.fc38.x86_64`

```
[57/853] Generating libnvme/nvme.py with a custom command
Deprecated command line option: -py3. Ignored, this option is no longer supported.
notype name zns
../src/nvme/types.h:7727: Warning 503: Can't wrap class $unnamed7$ unless renamed to a valid identifier.
```

igaw:
[...]

PR #638, opened 2023-04-27
python/swig: Wrap swig-sensitive struct inside #ifwdef SWIG

martin-belanger:
This is to suppress warning on anonymous struct/union

Ref: https://github.com/linux-nvme/libnvme/issues/634


martin-belanger:
> Maybe mention that the swig binding doesn't need the structs anyway in the commit message.

Done.

PR #639, opened 2023-04-27
python/swig: Do not use deprecated -py3 switch

tbzatek:
```
[56/854] Generating libnvme/nvme.py with a custom command
Deprecated command line option: -py3. Ignored, this option is no longer supported.
```

### DIFF
--- a/libnvme/meson.build
+++ b/libnvme/meson.build
@@ -22,7 +22,7 @@ if build_python_bindings
         'nvme.py',
         input:   ['nvme.i'],
         output:  ['nvme.py', 'nvme_wrap.c'],
-        command: [swig, '-python', '-py3', '-o', '@OUTPUT1@', '@INPUT0@'],
+        command: [swig, '-python', '-o', '@OUTPUT1@', '@INPUT0@'],
         install: true,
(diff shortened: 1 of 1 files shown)


martin-belanger:
Hi @tbzatek - I wanted to do exactly this, but I was wondering if that could cause problems for people using an older version of swig that requires `-py3`. What do you think? Can this change have negative effects for older swig versions?

tbzatek:
> Hi @tbzatek - I wanted to do exactly this, but I was wondering if that could cause problems for people using an older version of swig that requires `-py3`. What do you think? Can this change have negative effects for older swig versions?

I have no idea :-) The warning was produced with `swig-4.1.1`. Just tested with `swig-3.0.12` and got no warning either.

igaw:
https://repology.org/project/swig/versions

igaw:
Forget my idea. Indeed check for the version and only add the -py3 for older version sounds like the best plan forward.

martin-belanger:
Submitted https://github.com/linux-nvme/libnvme/pull/640 that will check the swig version and only use `-py3` if version is less than 4.1.0.

PR #640, opened 2023-04-27
python/swig: Check swig version to determine whether -py3 is needed

martin-belanger:
The swig option `-py3` has been deprecated as of swig-4.1.0. A deprecation warning is printed when used with swig-4.1.0 or later.


tbzatek (approved):
Confirming it works as expected on both `swig-3.0.12` and `swig-4.1.1`.

igaw:
Thanks a lot!

PR #643, opened 2023-05-12
fabrics: handle /dev/nvme-fabrics read failure

calebsander:
The ability to read from `/dev/nvme-fabrics` to find supported options is a newer Linux kernel feature added in `f18ee3d988157` (5.17-rc1). On earlier kernels, this read returns `EINVAL`,
preventing the controller from being added:
```
$ nvme discover --transport tcp --traddr 192.168.1.62
Failed to read from /dev/nvme-fabrics: Invalid argument
failed to add controller, error Invalid argument
```

So don't treat `EINVAL` as a fatal error, and instead fall back to a default set of supported options.
With this change, controllers can be created successfully:
```
$ nvme discover --transport tcp --traddr 192.168.1.62

Discovery Log Number of Records 4, Generation counter 125
...
```

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -196,6 +196,37 @@ const char *nvmf_cms_str(__u8 cm)
 	return arg_str(cms, ARRAY_SIZE(cms), cm);
 }
 
+/*
+ * Derived from Linux's supported options (the opt_tokens table)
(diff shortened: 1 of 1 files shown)


igaw:
Nice fix! FWIW, I've updated the comment style. Rest looks good!

PR #645, opened 2023-05-12
fabrics: check genctr after getting discovery entries

calebsander:
From the NVMe base spec (version 2.0c, section 5.16.1.23):
> If the host reads the Discovery Log Page using multiple Get Log Page commands the host should ensure that there has not been a change in the contents of the data. The host should read the Discovery Log Page contents in order (i.e., with increasing Log Page Offset values) and then re-read the Generation Counter after the entire log page is transferred. If the Generation Counter does not match the original value read, the host should discard the log page read as the entries may be inconsistent.

`nvme_get_log_page()` will issue multiple Get Log Page commands to fetch the discovery log page if it exceeds 4 KB. Since GENCTR is at the start of the log page, this ordering is possible:
- GENCTR is read by a Get Log Page command for the first 4 KB
- The log page is modified, changing GENCTR
- Other Get Log Page commands read the remainder of the log page

So the check that GENCTR hasn't changed will incorrectly pass, despite the log page having been modified. This can lead to inconsistent, missing, or duplicate log page entries.

Ensure a GENCTR update is not missed by fetching log page header again after all entries.

Also use `NVME_LOG_PAGE_PDU_SIZE` to match other `nvme_get_log_page()` calls instead of hard-coding the 4 KB max transfer length. And ensure LPO is correctly reset if the log page is read again.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -992,9 +992,10 @@ static struct nvmf_discovery_log *nvme_discovery_log(nvme_ctrl_t c,
 		nvme_msg(r, LOG_DEBUG, "%s: get header (try %d/%d)\n",
 			name, retries, max_retries);
 		args->rae = true;
+		args->lpo = 0;
 		args->len = size;
 		args->log = log;
-		ret = nvme_get_log_page(fd, 4096, args);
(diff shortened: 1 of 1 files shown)


igaw:
Thanks! FWIW, just update the comment style to `/* */`

PR #646, opened 2023-05-13
ioctl: Add debugging feature to show command outputs

ikegami-t:


src/libnvme.map
```diff
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+LIBNVME_1_6 {
```
tbzatek, line 3: The current git master is still v1.5 (unreleased)

ikegami-t: Thansk for your comment. Just fixed the patch to use the map file v1.5 definition.

src/nvme/tree.h
```diff
@@ -801,7 +801,7 @@ const char *nvme_ctrl_get_firmware(nvme_ctrl_t c);
 const char *nvme_ctrl_get_model(nvme_ctrl_t c);
 
 /**
- * nvme_ctrl_get_state() - Running state of an controller
+ * nvme_ctrl_get_state() - Running state of a controller
```
igaw, line 804: This seems to be an unrelated change.

igaw (changes requested):

src/nvme/tree.h
ikegami-t: Splited the change.

src/nvme/ioctl.c
```diff
@@ -86,10 +88,57 @@ static int nvme_submit_passthru64(int fd, unsigned long ioctl_cmd,
...
+	struct timeval end;
+	int err;
+
+	gettimeofday(&start, NULL);
+
+	err = ioctl(fd, ioctl_cmd, cmd);
+
+	gettimeofday(&end, NULL);
+
+	if (nvme_debug(false))
```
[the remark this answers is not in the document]
ikegami-t: Fixed as mentioned.

src/nvme/ioctl.c
```diff
@@ -86,10 +88,57 @@ static int nvme_submit_passthru64(int fd, unsigned long ioctl_cmd,
...
+
 static int nvme_submit_passthru(int fd, unsigned long ioctl_cmd,
 				struct nvme_passthru_cmd *cmd, __u32 *result)
 {
-	int err = ioctl(fd, ioctl_cmd, cmd);
+	struct timeval start;
+	struct timeval end;
+	int err;
+
+	gettimeofday(&start, NULL);
```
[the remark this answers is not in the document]
ikegami-t: This also fixed as suggested. Thank you.

igaw:
I was pondering if this could also be done with ftrace, but I think it is not so easy as I thought.

ikegami-t:
Just rebased.

src/libnvme.map
```diff
@@ -20,6 +20,8 @@ LIBNVME_1_5 {
 		nvme_root_set_application;
 		nvme_subsystem_get_application;
 		nvme_subsystem_set_application;
+		nvme_set_debug;
+		nvme_get_debug;
```
igaw, line 24: These two entries should be under the 1.6 section as we already released 1.5.

src/nvme/ioctl.h
```diff
@@ -3855,4 +3855,16 @@ int nvme_zns_append(struct nvme_zns_append_args *args);
  */
 int nvme_dim_send(struct nvme_dim_args *args);
 
+/**
+ * nvme_debug - Set NVMe command debugging output
```
igaw, line 3859: kdoc is not happy because `nvme_debug` should be `nvme_set_debug`. The same problem with the kdoc section below

src/libnvme.map
ikegami-t: Fixed the version.

src/nvme/ioctl.h
ikegami-t: Fixed the kdoc also. Thank you.

igaw:
Thanks

PR #647, opened 2023-05-16
tree: Fix string comparison

martin-belanger:
When comparing host/subsystem/controller objects for equality, we may conclude erroneously that two objects are equal if, for example, one object contains a NULL pointer when the other object contains a non-NULL pointer.

For example, the following does not work as expected:

```
if (subsysnqn && s->subsysnqn &&
    strcmp(s->subsysnqn, subsysnqn))
	continue;
```

In the above code, if subsysnqn==NULL but s->subsysnqn!=NULL, we would not hit the "continue" statement as we ought to.

The correct way to compare two string pointers that may be NULL is as follows:

```
if ((subsysnqn != s->subsysnqn)
    && (!s->subsysnqn || !subsysnqn
	|| strcmp(s->subsysnqn, subsysnqn)))
	continue;
```

With this, the only time that the two strings will be considered equal is if both pointers are equal (including if they're both NULL) or if the strcmp() returns 0.

Also, one must use case-insensitive string comparison when comparing IPv6 addresses.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -1081,11 +1081,13 @@ nvme_ctrl_t __nvme_lookup_ctrl(nvme_subsystem_t s, const char *transport,
 		if (traddr && c->traddr &&
 		    strcasecmp(c->traddr, traddr))
 			continue;
-		if (host_traddr && c->cfg.host_traddr &&
-		    strcmp(c->cfg.host_traddr, host_traddr))
+		if (c->cfg.host_traddr != host_traddr &&
+		    (!c->cfg.host_traddr || !host_traddr ||
+		     strcasecmp(c->cfg.host_traddr, host_traddr)))
 			continue;
-		if (host_iface && c->cfg.host_iface &&
-		    strcmp(c->cfg.host_iface, host_iface))
+		if (c->cfg.host_iface != host_iface &&
+		    (!c->cfg.host_iface || !host_iface||
+		     strcmp(c->cfg.host_iface, host_iface)))
 			continue;
 		if (trsvcid && c->trsvcid &&
 		    strcmp(c->trsvcid, trsvcid))


src/nvme/tree.c
```diff
@@ -430,11 +430,13 @@ struct nvme_subsystem *nvme_lookup_subsystem(struct nvme_host *h,
 	struct nvme_subsystem *s;
 
 	nvme_for_each_subsystem(h, s) {
-		if (subsysnqn && s->subsysnqn &&
-		    strcmp(s->subsysnqn, subsysnqn))
+		if ((subsysnqn != s->subsysnqn)
+		    && (!s->subsysnqn || !subsysnqn
+			|| strcmp(s->subsysnqn, subsysnqn)))
```
igaw, line 435: The formatting should be

```
	if (subsysnqn != s->subsysnqn &&
	    (!s->subsysnqn || !subsysnqn ||
	     strcmp(s->subsysnqn, subsysnqn))
```

src/nvme/tree.c
```diff
@@ -1076,19 +1081,25 @@ nvme_ctrl_t __nvme_lookup_ctrl(nvme_subsystem_t s, const char *transport,
 
 	c = p ? nvme_subsystem_next_ctrl(s, p) : nvme_subsystem_first_ctrl(s);
 	for (; c != NULL; c = nvme_subsystem_next_ctrl(s, c)) {
-		if (strcmp(c->transport, transport))
```
igaw, line 1079: subsysname, subsysnqn, transport, traddr and host_traddr can never be null. Thus this change is not neccessary.

src/nvme/tree.c
martin-belanger: Ok. I will fix it.

src/nvme/tree.c
martin-belanger: I think that `host_traddr` can be null. Can't it? Or is it set to an empty string when not used?

igaw:
I tried to convince myself that there is something br0ken (see #650) but don't see what's not working as expected.

martin-belanger:
Yep. It's a bit tricky to reproduce. And it's only for the case where `host_traddr` and/or `host_iface` are being set to non-null values.

For example, if an interface has more than one IP address assigned to it and the user wants to specify a different IP address than the default with `host_traddr`. Now, let's say that a second user wants to connect to the same controller, but this time they want to let the system pick the default source address (i.e. they don't specify `host_traddr). In this case, we will be comparing one controller that has host_traddr!=null with one that has host_traddr==null and libnvme will erroneously declare that the existing connection is a match, when in fact it's not. 

I know. This is a corner case that is not likely to happen in real life (although RH did report the issue).

We can hit a similar issue with `host_iface` as well.

src/nvme/tree.c
igaw: I mean the entries in the tree have these values always initialized to a valid string.

igaw:
Let me commit my unit test and then I would suggest you add this scenario. I think it will help in the discussion (hint @hreinecke was also not really convince we need it).

Anyway, I really want to get this get right and avoid discussions that there is something wrong.

igaw:
Thanks for the detailed report. I've finally understood what the issue is.

> However, because the following code fails to recognize that
> host_traddr==NULL is not the same as c->cfg.host_traddr==192.168.56.104,
> libnvme erroneously thinks that it can reuse the existing connection 
> instead of creating a new one.

Well, I think the core of this a policy question. The current implementation is not 'wrong' is just deciding that in absence of a host_traddr/host_iface it looks for the best match.  I really like to hear from @hreinecke  and @martin-gpy their thoughts on this.

martin-belanger:
> Well, I think the core of this a policy question. The current implementation is not 'wrong' is just deciding that in absence of a host_traddr/host_iface it looks for the best match. I really like to hear from @hreinecke and @martin-gpy their thoughts on this.

Me too. 

I would like to know what the policy should be. For discovery controllers it may be OK to reuse connections that do not match the `host-traddr` and/or `host-iface`. For I/O controllers, however, I would argue that if someone wants the connection on a specific interface and display a specific source address, using a connection that is not a perfect match may not be desirable.

igaw:
Maybe we should bring this question on to the linux-nvme mailing list. I think it would make sense to get the attention of everyone. I don't expect everyone is really following what's happening here (and I don't expect them too).

martin-belanger:
FYI: Moved the discussion to linux-nvme mailing list. 
http://lists.infradead.org/pipermail/linux-nvme/2023-June/040295.html

martin-belanger:
Closing for now. Will resubmit incremental changes based on community recommendations.

igaw:
Thanks a lot!

Issue #648, opened 2023-05-16
IP address comparison may fail even when two addresses are equivalent


martin-belanger:
This is somehow related to PR: https://github.com/linux-nvme/libnvme/pull/647

IPv6 has a special syntax where consecutive sections of zeros can be replaced with two colons (`::`). In other words, the following two IPv6 addresses are equivalent:

```
2001:0db8:0000:0000:0000:ff00:0042:8329 == 2001:0db8::ff00:0042:8329
```

Also, IPv6 supports "IPv4-mapped IPv6 addresses" where an IPv4 address can be expressed as an IPv6 address. Again, this can lead to comparison failures when in fact two addresses match. Here's an example:
```
::ffff:192.0.2.128  == 192.0.2.128
```

Although unlikely, IPv4 addresses can include leading 0s.
```
192.0.2.128 == 192.000.002.128
```

Currently, `libnvme` uses case-insensitive string comparison (`strcasecmp`) to determine if two IP addresses are equal. This would fail for the above cases.

I don't have a solution for this. Python provides a module (`ipaddress`) that makes comparing IP addresses easy and covers all cases above. I don't know if there is an equivalent C library that we could use. Otherwise, we may have to write an IP address comparison function.

martin-belanger:
Just to give you an idea of what this is trying to solve.
```
##########################################################################
Step 1) Connect using compressed IPv6 (fe80::800:27ff:fe00:0)
and specify interface using scope syntax (%enp0s8).
##########################################################################
$ sudo nvme discover -p -t tcp -s 8009 -a fe80::800:27ff:fe00:0%enp0s8
$ nvme list -v
Subsystem        Subsystem-NQN                                                                                    Controllers
---------------- ------------------------------------------------------------------------------------------------ ----------------
nvme-subsys0     nqn.2014-08.org.nvmexpress.discovery                                                             nvme0

Device   SN                   MN                                       FR       TxPort Address        Subsystem    Namespaces
-------- -------------------- ---------------------------------------- -------- ------ -------------- ------------ ----------------
nvme0    139315012e3c3b70fe9c Linux                                    5.19.0-4 tcp    traddr=fe80::800:27ff:fe00:0%enp0s8,trsvcid=8009,src_addr=fe80::c71b:9e1:c456:f855 nvme-subsys0

Device       Generic      NSID       Usage                      Format           Controllers
------------ ------------ ---------- -------------------------- ---------------- ----------------


##########################################################################
[...]

PR #649, opened 2023-05-16
util: Add ipaddrs_eq() to check whether two IP addresses are equal

martin-belanger:
@igaw - This is prep work for issue: https://github.com/linux-nvme/libnvme/issues/648

The new function `ipaddrs_eq()` can be used to compare IP addresses. Once https://github.com/linux-nvme/libnvme/pull/647 gets merged (assuming it does get merged), I will submit another PR to use `ipaddrs_eq()` everywhere that we compare IP addresses.

### DIFF
--- a/src/nvme/util.c
+++ b/src/nvme/util.c
@@ -8,6 +8,7 @@
  */
 
 #include <stdio.h>
+#include <stdbool.h>
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


martin-belanger:
Doh! I completely messed up the mrge/rebase. Let me close this and resubmit.

PR #653, opened 2023-05-23
examples: Fix wrong indentation in discover-loop.py

bdrung:
Running `examples/discover-loop.py` fails:

```
  File "examples/discover-loop.py", line 59
    c.disconnect()
    ^
IndentationError: expected an indented block after 'try' statement on line 58
```


igaw:
Thanks!

PR #654, opened 2023-06-02
util: rename ipaddrs_eq() to nvme_ipaddrs_eq() and make public.

martin-belanger:
As per https://github.com/linux-nvme/libnvme/issues/648, make `nvme_ipaddrs_eq()`  public.


igaw:
Thanks!

PR #656, opened 2023-06-12
tree: Use nvme_ipaddrs_eq() to compare IP addresses

martin-belanger:
Use `nvme_ipaddrs_eq()` to compare IP addresses. This is especially important for IPv6 addresses that can be expressed in upper or lower case, or in compressed or expanded form. For example the following two IP addresses are the same, but a simple string comparison would declare them to be different.

```
2001:0DB8:0000:0000:0000:FF00:0042:8329
```
is the same as
```
2001:0db8::ff00:0042:8329
```


igaw:
Thanks!

PR #659, opened 2023-06-14
Extend execution context filtering

igaw:
The newly introduced feature to allow coexisting nvme-cli and nvme-stas needs some more tuning to get it working.

For example, we should also fitler out the well known discovery controllers.


igaw:
Fully tested with https://github.com/osandov/blktests/pull/121 and https://github.com/linux-nvme/nvme-cli/pull/1978

Issue #206, opened 2022-01-31
API cleanup: nvme_$NAME_get() vs nvme_get_$NAME()


igaw:
The current API isn't really consistent in the naming pattern. We should use one an stick to it.

ikegami-t:
The PR #660 is closed as correct implementation so seems the issue can be closed now.

igaw:
I've opened this issue for the various variants of our getters/setters for the non MI API. When we do the 2.0 we should go over the existing function names an make them consistent.

ikegami-t:
Seems there is no other remaining function name to fix as below.
```sh
tokunori@tokunori-desktop:~/nvme-cli/subprojects/libnvme$ git grep "_get("
examples/mi-conf.c:     bus = dbus_bus_get(DBUS_BUS_SYSTEM, &berr);
examples/mi-mctp.c:int do_config_get(nvme_mi_ep_t ep, int argc, char **argv)
examples/mi-mctp.c:             rc = do_config_get(ep, argc, argv);
libnvme/nvme.i: const char *nvme_host_dhchap_key_get(struct nvme_host *h) {
libnvme/nvme.i: const char *nvme_subsystem_name_get(struct nvme_subsystem *s) {
libnvme/nvme.i: struct nvme_host *nvme_subsystem_host_get(struct nvme_subsystem *s) {
libnvme/nvme.i: const char *nvme_ctrl_name_get(struct nvme_ctrl *c) {
libnvme/nvme.i: struct nvme_subsystem *nvme_ctrl_subsystem_get(struct nvme_ctrl *c) {
libnvme/nvme.i: const char *nvme_ctrl_state_get(struct nvme_ctrl *c) {
libnvme/nvme.i: const char *nvme_ctrl_dhchap_key_get(struct nvme_ctrl *c) {
libnvme/nvme.i: const char *nvme_ctrl_dhchap_host_key_get(struct nvme_ctrl *c) {
libnvme/nvme.i: const char *nvme_ns_name_get(struct nvme_ns *n) {
libnvme/nvme.i: PyObject *nbft_get(const char * filename)
libnvme/nvme.i:PyObject *nbft_get(const char * filename);
libnvme/tests/test-nbft.py:        actual_nbft = nvme.nbft_get(args.filename)
src/nvme/json.c:        attr_obj = json_object_object_get(port_obj, "transport");
src/nvme/json.c:        attr_obj = json_object_object_get(port_obj, "traddr");
[...]

PR #660, opened 2023-06-14
mi: Cleanup nvme_mi_mi_config_get() to nvme_mi_get_mi_config()

ikegami-t:


ikegami-t:
Changes for the issue #206.

igaw:
@jk-ozlabs

igaw:
My take on this, we will do this type of work for a major version update, e.g. libnvme 2.0.

jk-ozlabs:
Nack on this one. The `nvme_mi_mi_` prefix is intentional: we use this for all of the `nvme_mi_` functions that send MI-type messages. So we have:

 * `nvme_mi_mi_` for the MI commands sent over MI; and
 * `nvme_mi_admin` for the Admin commands sent over MI

This is documented at: https://github.com/linux-nvme/libnvme/blob/42ac4535963556610a5bba8d1277c5a398a6ed5a/src/nvme/mi.h#L35-L44

igaw:
True, I forgot that we already discussed this. Let's keep it as it is.

jk-ozlabs:
(and the `config_get` is a direct reference to the MI Configuration Get command; not sure if we want to apply the verb-noun ordering to this one...)

ikegami-t:
Thanks for your comments. I could understand them. By the way seems the PCIe commands over MI not implemented yet. (Also seems the mi resp status NVME_MI_RESP_PCIE_INACCESSIBLE is for the unimplemented command.)

jk-ozlabs:
Yes, that's correct - I haven't done any work on the PCIe command-set. It's optional in the spec, and I don't know whether any of the device I have actually support it!

Is this something you're looking for, to be implemented?

ikegami-t:
Thanks for the confirmation. No I thought to confirm if just my understanding for the PICe command-set support implementaton is correct or not. (I am also not sure if my device supported it.)

PR #661, opened 2023-06-15
util: Use HAVE_NETDB instead of HAVE_LIBNSS

ikegami-t:


martin-belanger:
Just curious... 

I think that this whole logic around HAVE_LIBNSS/HAVE_NETDB is to make sure that `getaddrinfo()` will be able to resolve hostnames. However, isn't `getaddrinfo()` part of `glibc` (not `libnss` or `netdb`). And if so, wouldn't `getaddrinfo()` handle a missing libnss/netdb gracefully? 

I'm not sure why any of this is needed.

igaw:
`getaddrinfo()` and friends are only available when linking dynamically against glibc. Thus when building statically these function are not available and the linking step will fail.

PR #662, opened 2023-06-17
types: Define NS_FLBAS MASK and SHIFT to use NVME_GET

ikegami-t:


ikegami-t:
Changes for the issue #148 .

ikegami-t:
Rebased and fixed the NVME_NS_FLBAS_META_EXT macro definition.

src/nvme/types.h
```diff
@@ -1854,27 +1854,34 @@ enum nvme_id_nsfeat {
...
+	NVME_NS_FLBAS_LOWER_SHIFT	= 0,
+	NVME_NS_FLBAS_META_EXT_SHIFT	= 4,
+	NVME_NS_FLBAS_HIGHER_SHIFT	= 5,
+	NVME_NS_FLBAS_LOWER_MASK	= 0xf,
+	NVME_NS_FLBAS_META_EXT_MASK	= 0x1,
+	NVME_NS_FLBAS_HIGHER_MASK	= 0x3,
 };
 
+#define NVME_NS_FLBAS_LOWER(flbas)	NVME_GET(flbas, NS_FLBAS_LOWER)
+#define NVME_NS_FLBAS_META_EXT(flbas)	NVME_GET(flbas, NS_FLBAS_META_EXT)
```
igaw, line 1882: The change of `NVME_NS_FLBAS_META_EXT` from a simple define to a macro will break existing users. While for nvme-cli that might be okay as we have to update the minimum dependency to libnvme anyway, it's not so easy to tell about other users. I am aware of that the udisk and nvme-stats project is using libnvme but not this define. 

So what we can do is merge it but if someone complains we are reverting it. What do you think?

igaw:
BTW, the build fails due to kdoc test failing:
```
src/nvme/types.h:1880: warning: Enum value 'NVME_NS_FLBAS_LOWER_SHIFT' not described in enum 'nvme_id_ns_flbas'
src/nvme/types.h:1880: warning: Enum value 'NVME_NS_FLBAS_META_EXT_SHIFT' not described in enum 'nvme_id_ns_flbas'
src/nvme/types.h:1880: warning: Enum value 'NVME_NS_FLBAS_HIGHER_SHIFT' not described in enum 'nvme_id_ns_flbas'
```

ikegami-t:
Thanks for your review comments. Changed the patch to remain existing macro definitions and add also additional macro definitions for a simple macro define.

igaw:
It looks like kdoc is not able to figure out that it should ignore the right handside of the enum definition. I couldn't figure out how fix address this.

ikegami-t:
Sorry for the kernel-doc warnings. Fixed as remaining the original definitions.

igaw:
No problem. It seems that this version also avoids breakage. Nice.

Issue #477, opened 2022-09-22
Rename nvme_cmd_get_log_telemetry_host_lsp to nvme_log_telemetry_host_lsp


igaw:
Agreed. I will drop the `get`. 
FWIW, I was trying to stay consistent with another "get log" LSP enum (i.e. `nvme_cmd_get_log_telemetry_host_lsp`)

_Originally posted by @martin-belanger in https://github.com/linux-nvme/libnvme/pull/476#discussion_r977777108_

ikegami-t:
As mentioned by https://github.com/linux-nvme/libnvme/pull/663#issuecomment-1596610812 should be closed.

PR #663, opened 2023-06-17
types: Rename nvme_cmd_get_log_telemetry_host_lsp to nvme_log_telemet…

ikegami-t:
…ry_host_lsp


igaw:
Thanks for these changes. Note, the #477 issue is one of the 2.0 milestones issues. Most the 2.0 milestone issue are breaking the existing API. 

Given that we are going to generate the API from the upcoming official NVME JSON file, the #477 issue doesn't really make sense at all. I think we should just close it.

ikegami-t:
Thanks for your review comments. Okay will close the PR.

Issue #613, opened 2023-04-04
libnvme using deprecated json-c APIs


cleech:
This popped up in some testing with a libnvme built against json-c 0.14 that was then installed on a system with json-c 0.15.  The printbuf APIs were moved from the JSONC_0.14 tag to JSONC_PRIVATE, breaking the json-c ABI.  They continue to work if recompiling against 0.15+, but are deprecated for external use and may be removed in a future version.

It looks like the printbuf is being used as a re-allocating heap buffer during JSON parsing only, and it's use should be removed.

cleech:
I think we can pass the read buffers to json_tokener_parse_ex incrementally, and not need a flexible buffer to hold a large file at all.  Or maybe just mmap the file.

I can work up merge request for this.

ikegami-t:
The PR #664 merged into master so resolved.

igaw:
Thanks!

PR #664, opened 2023-06-17
json: Use memory block allocated by realloc() instead printbuf

ikegami-t:


ikegami-t:
Changes for the issue #613. If possible could you please try the test if the changes work correctly?

src/nvme/json.c
```diff
@@ -193,16 +193,20 @@ static struct json_object *parse_json(nvme_root_t r, int fd)
 {
 	char buf[JSON_FILE_BUF_SIZE];
 	struct json_object *obj = NULL;
-	struct printbuf *pb;
+	char *str;
```
igaw, line 196: The compiler seems not be able to figure out that str is always defined when calling `json_tokener_parse_ex()`. So you might just want to explicitly initialize it here to NULL.

ikegami-t: Thanks for your review comment. Just updated the patch to add allocated data length checking instead of NULL pointer initialization.

ikegami-t: Sorry refixed the patch to initialize the pointer to NULL as mentioned.

igaw: No worries, that's why we have some CI builds :)

PR #665, opened 2023-06-20
fabrics: use SECTYPE to determine whether to use TLS

calebsander:
The NVMe specfications are clear that the discovery log page entry SECTYPE field indicates whether TLS is supported.
Currently the TREQ field is used, which results in enabling TLS even when SECTYPE = "No Security" and TREQ = "Not required". Only enable TLS if SECTYPE indicates a TLS version is enabled.

From the NVMe/TCP transport specification, version 1.0c:
> Security Type (SECTYPE): Specifies the type of security used by the NVMe/TCP port. If SECTYPE is a value of 0h (No Security), then the host shall set up a normal TCP connection.

From TP8018:
> The SECTYPE value of 0h (No Security) specifies that TLS is not supported.

From TP8025 (the "TLS permitted" host case):
> If the SECTYPE field in the TSAS field in the discovery log entry for the remote entity is cleared to zero and the TSC field is not set to 01b (i.e., Required), then initiate TCP connections without TLS.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1033,8 +1033,7 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 		c->cfg.disable_sqflow = true;
 
 	if (e->trtype == NVMF_TRTYPE_TCP &&
-	    (e->treq & NVMF_TREQ_REQUIRED ||
-	     e->treq & NVMF_TREQ_NOT_REQUIRED))
+	    e->tsas.tcp.sectype != NVMF_TCP_SECTYPE_NONE)
 		c->cfg.tls = true;
 
 	ret = nvmf_add_ctrl(h, c, cfg);


igaw:
Looks reasonable to me, but I am not really into this topic. @hreinecke any feedback?

hreinecke:
We can only merge these changes (actually, _talk_ about these changes) once the TPARs are published. To my knowledge this isn't the case. So we have to wait until then.

calebsander:
> We can only merge these changes (actually, talk about these changes) once the TPARs are published.

I am sympathetic to that argument, but does it not also apply to the existing code that's already in `libnvme`? `nvme connect-all` currently breaks if it connects to a NVMe/TCP controller that doesn't support TLS but sets TREQ = "Not required" (which appears reasonable from the spec's limited explanation of the TREQ values). It looks from your NVMe/TLS kernel patch series that setting the `tls` option in the `/dev/nvme-fabrics` connect string is also not how TLS is going to be enabled; looks like userspace needs to use `keyring` or `tls_key` to specify a specific key. So how about just removing this code path for setting the `tls` option that the kernel won't understand?
Also, the meaning of SECTYPE = 0 is already defined by the (ratified) NVMe/TCP spec. I reference the TPs only because I think they reinforce the sentiment. But the spec itself seems quite clear that connections should not be made with TLS if the controller advertises SECTYPE = 0.

calebsander:
TP 8025 has been ratified. Assuming the Linux host is acting in the "TLS Permitted" case, it specifies the following behavior:
> If the SECTYPE field in the TSAS field in the Discovery Log Page Entry for the
remote entity is not cleared to 0h, then initiate TCP connections with TLS,
irrespective of the value of the TSC field in that Discovery Log Page Entry. If
establishing any TCP connection with TLS fails and the TSC field in that Discovery
Log Page Entry is not set to 01b (i.e., Required), the host may fall back to initiate
TCP connections without TLS.
If the SECTYPE field in the TSAS field in the Discovery Log Page Entry for the
remote entity is cleared to 0h and the TSC field is not set to 01b (i.e., Required),
then initiate TCP connections without TLS. If the SECTYPE field in the TSAS field
in the Discovery Log Page Entry for the remote entity is cleared to 0h and the TSC
field is set to 01b (i.e., Required), then that Discovery Log Page Entry is inconsistent
and TCP connections without TLS may or may not be initiated.

So it seems clear that `SECTYPE` is what the host should be checking to determine whether to use TLS, not `TSC` (the subfield of `TREQ`).

igaw:
Okay looks reasonable to me, after reading several times the section in the TP... Anyway, someone will surely complain if we get this wrong :)

igaw:
Thanks!

Issue #619, opened 2023-04-12
Refactor use of nvme_msg in nbft.c


johnmeneghini:
```
Functions calling this function: nvme_msg

  File   Function          Line
0 nbft.c verify             84 nvme_msg(NULL, LOG_DEBUG, "file %s: " message "\n", \
1 nbft.c __get_heap_obj     99 nvme_msg(NULL, LOG_DEBUG,
2 nbft.c __get_heap_obj    110 nvme_msg(NULL, LOG_DEBUG,
3 nbft.c __get_heap_obj    117 nvme_msg(NULL, LOG_DEBUG,
4 nbft.c read_ssns         231 nvme_msg(NULL, LOG_DEBUG,
5 nbft.c read_ssns         264 nvme_msg(NULL, LOG_DEBUG,
6 nbft.c read_ssns         282 nvme_msg(NULL, LOG_DEBUG,
7 nbft.c read_ssns         291 nvme_msg(NULL, LOG_DEBUG,
8 nbft.c read_hfi_info_tcp 335 nvme_msg(NULL, LOG_DEBUG,
9 nbft.c read_hfi          412 nvme_msg(NULL, LOG_DEBUG,
a nbft.c read_discovery    454 nvme_msg(NULL, LOG_DEBUG,
b nbft.c read_discovery    460 nvme_msg(NULL, LOG_DEBUG,
c nbft.c nvme_nbft_read    695 nvme_msg(NULL, LOG_ERR, "Failed to open %s: %s\n",
d nbft.c nvme_nbft_read    703 nvme_msg(NULL, LOG_ERR, "Failed to read from %s: %s\n",
e nbft.c nvme_nbft_read    715 nvme_msg(NULL, LOG_ERR, "Failed to allocate memory for NBFT table");
f nbft.c nvme_nbft_read    723 nvme_msg(NULL, LOG_ERR, "Failed to read from %s: %s\n",
g nbft.c nvme_nbft_read    737 nvme_msg(NULL, LOG_ERR, "Could not allocate memory for NBFT\n");
h nbft.c nvme_nbft_read    748 nvme_msg(NULL, LOG_ERR, "Failed to parse %s\n", filename);
```

johnmeneghini:
See the discussion in: https://github.com/linux-nvme/libnvme/pull/572#discussion_r1158343257

igaw:
[...]

PR #669, opened 2023-06-24
nbft: Refactor use of nvme_msg in nbft.c

ikegami-t:


ikegami-t:
Changes for the issue #619.

src/nvme/nbft.h
```diff
@@ -1227,7 +1229,7 @@ struct nbft_info {
  *
  * Return: 0 on success, errno otherwise.
  */
-int nvme_nbft_read(struct nbft_info **nbft, const char *filename);
+int nvme_nbft_read(nvme_root_t r, struct nbft_info **nbft, const char *filename);
```
igaw, line 1232: Just passing in a `nvme_root_t` root object for printing the message is not the right thing to do. During review of this code we figured that this is not the right thing to do. That's why this argument is missing. 

We have the same problem for all non fabric functions that we don't have a `nvme_root_t` object and we might also want to print errors etc in those functions. 

The only way I see a way around this is to introduce a global variable containing the file descriptor to print to. Obviously, not totally happy with it, but what else can we do?

ikegami-t:
Thanks for your review comments. Just fixed the patch as mentioned.

ikegami-t:
Sorry missed to set NULL the variable after the nvme root deleted so I will update the changes later.

igaw:
export new function, update docs, reorder free sequence

PR #677, opened 2023-07-05
tree: missing closedir() causes fd leak for "/sys/bus/pci/slots"

martin-belanger:
In nvme_ctrl_lookup_phy_slot(), we are missing a closedir(), which causes file descriptors to leak. Also, there was a missing free() when the function returns with ENOMEM.


igaw:
Thanks!

Issue #672, opened 2023-06-29
Possible file descriptor leak (i.e. many namespace devices are left opened)


martin-belanger:
In nvme-stas, I use the Python bindings to create connections to controllers. A daemon creates a "root" and a "host" objects. Then a controller object is created and the "connect" method is invoked for that object (this invokes the C function `nvmf_add_ctrl`). These 3 objects (root, host, ctrl) remain in existence for as long as the daemon is running (and as long as nobody manually tells the daemon to remove the connection to the controller).

Using `lsof -p [daeemon-pid]` (see output below), I observe that the daemon has a lot of namespace devices opened. I'm just wondering why. This doesn't scale. As the daemon connects to more controllers the number of opened namespace devices keeps increasing until we hit the daemon's limit of opened files (default 1024).

I know that libnvme scans the sysfs to build a picture of what nvme devices are already in the system. Is it possible that we're leaking file descriptors (i.e. we forget to call `close()`) while scanning the sysfs. I looked at the code, but can't find where this is happening.

```
# lsof -p 42683
COMMAND   PID USER   FD      TYPE             DEVICE SIZE/OFF   NODE NAME
python3 42683 root  cwd       DIR               0,49      272    256 /
python3 42683 root  rtd       DIR               0,49      272    256 /
python3 42683 root  txt       REG               0,49    10560 195219 /usr/bin/python3.6
python3 42683 root  mem       REG               0,49    15376 134253 /usr/lib64/python3.6/site-packages/systemd/id128.cpython-36m-x86_64-linux-gnu.so
python3 42683 root  mem       REG               0,49    41640 134251 /usr/lib64/python3.6/site-packages/systemd/_reader.cpython-36m-x86_64-linux-gnu.so
python3 42683 root  mem       REG               0,49    10896 134250 /usr/lib64/python3.6/site-packages/systemd/_journal.cpython-36m-x86_64-linux-gnu.so
python3 42683 root  mem       REG               0,49    97112 194704 /usr/lib64/python3.6/lib-dynload/_pickle.cpython-36m-x86_64-linux-gnu.so
python3 42683 root  mem       REG               0,49   219592 183508 /usr/lib64/libudev.so.1.7.2
python3 42683 root  mem       REG               0,49    84120 194692 /usr/lib64/python3.6/lib-dynload/_datetime.cpython-36m-x86_64-linux-gnu.so
[...]

PR #678, opened 2023-07-05
tree: Don't open nvme devices until it's absolutely required

martin-belanger:
Don't open nvme devices while scanning the tree. Only open devices when we actually need to write commands to them.

This patch also provides functions to close fds when a user no longer needs them to be open.

Ref: https://github.com/linux-nvme/libnvme/issues/672

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -341,6 +341,14 @@ void nvme_free_tree(nvme_root_t r)
 	free(r);
 }
 
+void nvme_root_release_fds(nvme_root_t r)
+{
+	struct nvme_host *h, *_h;
+
+	nvme_for_each_host_safe(r, h, _h)
+		nvme_host_release_fds(h);
+}
+
 const char *nvme_subsystem_get_nqn(nvme_subsystem_t s)
 {
 	return s->subsysnqn;
@@ -412,7 +420,7 @@ nvme_path_t nvme_namespace_next_path(nvme_ns_t ns, nvme_path_t p)
 static void __nvme_free_ns(struct nvme_ns *n)
 {
 	list_del_init(&n->entry);
-	close(n->fd);
+	nvme_ns_release_fd(n);
 	free(n->generic_name);
 	free(n->name);
 	free(n->sysfs_dir);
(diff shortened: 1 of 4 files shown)


src/nvme/tree.c
```diff
@@ -1622,9 +1656,29 @@ static int nvme_bytes_to_lba(nvme_ns_t n, off_t offset, size_t count,
 
 int nvme_ns_get_fd(nvme_ns_t n)
 {
+	if (n->fd < 0) {
+		n->fd = nvme_open(n->name);
+		if (n->fd < 0) {
+			nvme_root_t r = n->s && n->s->h ? n->s->h->r :
+					n->c && n->c->s && n->c->s->h ? n->c->s->h->r :
+					NULL;
```
igaw, line 1664: Maybe we should introduce a helper for this?

src/nvme/tree.c
```diff
@@ -787,18 +815,27 @@ static int nvme_ctrl_scan_path(nvme_root_t r, struct nvme_ctrl *c, char *name)
...
 				 "Failed to open ctrl %s, errno %d\n",
 				 c->name, errno);
+		}
 	}
 	return c->fd;
 }
 
+void nvme_ctrl_release_fd(nvme_ctrl_t c)
+{
+	if (c->fd > -1) {
```
igaw, line 833:
```
if (c->fd < 0)
    return;
````

File: src/nvme/tree.c [hunk 3]
```diff
@@ -1622,9 +1656,29 @@ static int nvme_bytes_to_lba(nvme_ns_t n, off_t offset, size_t count,
...
+				 n->name, errno);
+		}
+	}
+
 	return n->fd;
 }
 
+void nvme_ns_release_fd(nvme_ns_t n)
+{
+	if (n->fd > -1) {
```
igaw, line 1676: Same here, just do a early return.

src/nvme/tree.c
```diff
@@ -1959,34 +2013,37 @@ static void nvme_ns_set_generic_name(struct nvme_ns *n, const char *name)
...
+		goto free_ns;
 
 	if (nvme_ns_init(n) != 0)
-		goto close_fd;
+		goto free_ns;
 
 	list_head_init(&n->paths);
 	list_node_init(&n->entry);
 
+	nvme_ns_release_fd(n); /* Do not leak fds */
```
igaw, line 2042: Alright, so the aim is to get rid of the fd eventually so we don't need to open it at all.

igaw:
Looks good, just a few nitpicks.

src/nvme/tree.c
hreinecke: That was actually not what I had in mind.
`# cat /sys/block/nvme0n1/nsid`
provides you with the same result; we should rather read _all_ information from sysfs and _never_ open a filedescriptor for the initial scanning. Doing so will block if the namespace/controller is undergoing a reset, and hence 'nvme disconnect' will hang for systems where the controller is at fault.

igaw: Yes, this where we aiming towards. This is towards this goal.

martin-belanger: Personally, I would prefer to never cache the fds and always open/close before/after every read/write/ioctl operations. But that would mean a big change in the way libnvme works and would require users to always remember to close the fd after use. 

Like Daniel said, this PR is a step towards better handling of file descriptors.

@igaw - I will make the changes you suggested this morning.

PR #683, opened 2023-07-25
fabrics: Do not pass disable_sqflow if not supported

igaw:
Only retry a connect attempt with disable_sqflow if the kernel
actually supports this option.

Reported-by: Sagi Grimberg <sagi@grimberg.me>

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1029,7 +1029,8 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 		return NULL;
 	}
 
-	if (e->treq & NVMF_TREQ_DISABLE_SQFLOW)
+	if (e->treq & NVMF_TREQ_DISABLE_SQFLOW &&
+	    nvmf_check_option(h->r, disable_sqflow))
 		c->cfg.disable_sqflow = true;
 
 	if (e->trtype == NVMF_TRTYPE_TCP &&
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -357,10 +357,16 @@ static int __add_argument(char **argstr, const char *tok, const char *arg)
 	return 0;
 }
 
+static int __nvmf_supported_options(nvme_root_t r);
+#define nvmf_check_option(r, tok)					\
+({									\
+	!__nvmf_supported_options(r) && (r)->options->tok;		\
+})
+
 #define add_bool_argument(o, argstr, tok, arg)				\
 ({									\
 	int ret;							\
-	if (r->options->tok) {						\
+	if (nvmf_check_option(r, tok)) {				\
 		ret = __add_bool_argument(argstr,			\
 					  stringify(tok),		\
 					  arg);				\
@@ -376,7 +382,7 @@ static int __add_argument(char **argstr, const char *tok, const char *arg)
 #define add_int_argument(o, argstr, tok, arg, allow_zero) \
 ({									\
 	int ret;							\
-	if (r->options->tok) {						\
+	if (nvmf_check_option(r, tok)) {				\
 		ret = __add_int_argument(argstr,			\
 					stringify(tok),			\
 					arg,				\
@@ -393,7 +399,7 @@ static int __add_argument(char **argstr, const char *tok, const char *arg)
 #define add_int_or_minus_one_argument(o, argstr, tok, arg)		\
 ({									\
 	int ret;							\
-	if (r->options->tok) {						\
+	if (nvmf_check_option(r, tok)) {				\
 		ret = __add_int_or_minus_one_argument(argstr,		\
 						     stringify(tok),	\
 						     arg);		\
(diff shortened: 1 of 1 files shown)


src/nvme/fabrics.c
```diff
@@ -652,6 +678,62 @@ static int build_options(nvme_host_t h, nvme_ctrl_t c, char **argstr)
...
+	int i;
+
+	for (i = 0; i < ARRAY_SIZE(default_options_strings); i++) {
+		if (!strcmp(opt, default_options_strings[i]))
+			return true;
+	}
+	return false;
+}
+
+static bool nvmf_check_option(const char *opt)
```
calebsander, line 692: This seems to almost completely duplicate `nvmf_read_options()` and `default_options`. Why can't it be implemented like this?
```c
#define nvmf_check_option(r, tok) (!nvmf_read_options(r) && r->options->tok)
```
The call site would have to change to
```c
nvmf_check_option(r, disable_sqflow)
```
This also has the benefit of caching the supported options.

calebsander (changes requested):

src/nvme/fabrics.c
igaw: Yes, I got the same feedback on the mailing list. Let me update this accordingly.

src/nvme/fabrics.c
```diff
@@ -357,10 +357,18 @@ static int __add_argument(char **argstr, const char *tok, const char *arg)
 	return 0;
 }
 
+static int __nvmf_supported_options(nvme_root_t r);
+#define nvmf_check_option(r, tok)					\
+({									\
+	if (!(r)->options)						\
+		__nvmf_supported_options(r);				\
+	(r)->options->tok;						\
```
calebsander, line 365: If `__nvmf_supported_options()` fails to allocate memory, it can return with `(r)->options` still NULL, which would lead to a segfault. Should we check the return value?

src/nvme/fabrics.c
```diff
@@ -357,10 +357,18 @@ static int __add_argument(char **argstr, const char *tok, const char *arg)
 	return 0;
 }
 
+static int __nvmf_supported_options(nvme_root_t r);
+#define nvmf_check_option(r, tok)					\
+({									\
+	if (!(r)->options)						\
```
calebsander, line 363: This makes the `if (r->options) return 0;` check in `__nvmf_supported_options()` redundant now?

src/nvme/fabrics.c
```diff
@@ -1026,8 +1031,11 @@ nvme_ctrl_t nvmf_connect_disc_entry(nvme_host_t h,
 		return NULL;
 	}
 
-	if (e->treq & NVMF_TREQ_DISABLE_SQFLOW)
+	if (e->treq & NVMF_TREQ_DISABLE_SQFLOW &&
+	    nvmf_check_option(h->r, disable_sqflow))
 		c->cfg.disable_sqflow = true;
+	else
+		c->cfg.disable_sqflow = false;
```
calebsander, line 1038: Explicitly setting it to `false` is probably not necessary because `c->cfg` is 0-initialized by `nvmf_default_config()` (called from `nvme_create_ctrl()` above)

martin-belanger: Could we simply write:
```
c->cfg.disable_sqflow = (e->treq & NVMF_TREQ_DISABLE_SQFLOW) && nvmf_check_option(h->r, disable_sqflow);
```

igaw: `cfg` is passed into the fucntion, so we don't have any control on the settings. We need to explicitly set it.

I find the explict if/else thing simpler too read, but I don't care too much. So if you insist I update it.

src/nvme/fabrics.c
igaw: Yeah makes sense, though IIRC such allocations will never fail, instead the program gets terminated by the OOM killer.

src/nvme/fabrics.c
```diff
@@ -357,10 +357,19 @@ static int __add_argument(char **argstr, const char *tok, const char *arg)
 	return 0;
 }
 
+static int __nvmf_supported_options(nvme_root_t r);
+#define nvmf_check_option(r, tok)					\
+({									\
+	if (!(r)->options)						\
+		if (__nvmf_supported_options(r))			\
+			return false;					\
```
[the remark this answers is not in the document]
igaw: Yes obviously wrong. Indeed, we have the `r->options` check in `nvmf_read_options`. I'll update it to your first suggestion.

src/nvme/fabrics.c
calebsander: But `c->cfg` is not derived from `cfg` as far as I could tell?
I would agree that not duplicating `c->cfg.disable_sqflow` in each of the cases makes it a bit easier to follow the purpose of this code, but I don't feel strongly.

igaw: You are right. I confused `c->cfg` with `cfg`. In this case we don't need to set it explicitly to false.

igaw: Should be okay now :)

src/nvme/fabrics.c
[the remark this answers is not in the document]
igaw: Thanks for pointing this out. I am trying to work on too many things at the same time leads obviously to crapy code.

PR #687, opened 2023-07-30
doc: fix minor mistake in README.md about dependencies

cvubrugier:
OpenSSL is used for TLS over TCP whereas Keyutils is used for authentication.

Also fix spelling mistake: dependend -> dependent.


igaw:
Thanks. 

Just as info OpenSSL is also used for key derivation, so it is also part of the authentication step, but I don't think this detail is really important here.

PR #688, opened 2023-07-31
Test discovery log page functions by mocking NVMe passthru ioctls

calebsander:
Functions issuing admin/IO passthru ioctls sorely lack unit tests.
It would be great for unit tests not to need a real NVMe controller.
It's also useful to be able to test responses to commands
that might be impossible to trigger with real controllers.

To that end, implement infrastructure for mocking `ioctl()`,
allowing tests to set expectations for the NVMe passthru ioctls
that will be issued and control the corresponding responses.

The mock library can be used with `LD_PRELOAD` so that `libnvme`'s `ioctl()`
calls are redirected from `libc`. No changes to `libnvme` are needed.

Add unit tests for `nvmf_get_discovery_log()`.
They provide coverage of the logic in `nvme_discovery_log()`,
`nvme_get_log_page()`, and `nvme_get_log()` too.
The tests use the mock `ioctl()` infra to validate the Get Log Page
commands issued and inject responses triggering different code paths.

### DIFF
--- a/test/ioctl/discovery.c
+++ b/test/ioctl/discovery.c
@@ -1,0 +1,418 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include <libnvme.h>
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ccan/array_size/array_size.h>
+#include <ccan/endian/endian.h>
+
+#include "../../src/nvme/private.h"
+#include "mock.h"
+#include "util.h"
+
+#define TEST_FD 0xFD
+
+static void test_no_entries(nvme_ctrl_t c)
+{
+	struct nvmf_discovery_log header = {};
+	/* No entries to fetch after fetching the header */
+	struct mock_cmd mock_admin_cmds[] = {
+		{
+			.opcode = nvme_admin_get_log_page,
+			.data_len = sizeof(header),
+			.cdw10 = (sizeof(header) / 4 - 1) << 16 /* NUMDL */
+			       | 1 << 15 /* RAE */
+			       | NVME_LOG_LID_DISCOVER, /* LID */
+			.out_data = &header,
+		},
+	};
+	struct nvmf_discovery_log *log = NULL;
+
+	set_mock_admin_cmds(mock_admin_cmds, ARRAY_SIZE(mock_admin_cmds));
+	check(nvmf_get_discovery_log(c, &log, 1) == 0, "discovery failed: %m");
+	end_mock_cmds();
+	cmp(log, &header, sizeof(header), "incorrect header");
+	free(log);
+}
+
+static void test_four_entries(nvme_ctrl_t c)
+{
+	struct nvmf_disc_log_entry entries[4];
+	struct nvmf_discovery_log header = {
+		.numrec = cpu_to_le64(ARRAY_SIZE(entries)),
+	};
+	/*
+	 * All 4 entries should be fetched at once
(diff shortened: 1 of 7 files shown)


test/ioctl/discovery.c
```diff
@@ -0,0 +1,417 @@
...
+
+	struct nvmf_discovery_log *log = NULL;
+	check(!nvmf_get_discovery_log(c, &log, 1), "discovery failed: %m");
+	compare(log, &header, sizeof(header), "incorrect header");
+	free(log);
+}
+
+static void test_four_entries(nvme_ctrl_t c)
+{
+	struct nvmf_disc_log_entry entries[4];
```
igaw, line 39: ` entries[4] = { 0, };` should pre initialize the struct so you don't need the memset below as far I can tell.

calebsander: Yes, I wanted each entry to have distinct contents so that we can validate the right entries are returned. This was the most succinct way I could come up with. Do you have a different suggestion?

calebsander: > unhappy that the definition and real code are intermixed.

Sorry, perhaps I am missing your point. What definition and code intermixing are you referring to? Would you like the test expectations to be defined separately from the action performed by the test case? Or the variable declarations to be moved to the top of each scope, à la C89?

igaw: Sorry for the delay was on holiday last week and only bored once :)

Yeah, I would prefer to keep the coding style consistent and follow the Linux Coding Style. I don't have a strong opinion what is better but it really helps if the code base is looking the same.

igaw: E.g. `checkpatch` is pretty unhappy with the formatting. Sure, no need to blindly follow all the 'recommendations' from `checkpatch` but still there are plenty of complains I agree on.

```shell
$ ~/work/linux/scripts/checkpatch.pl -g HEAD~
[...]
WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
#49: FILE: test/ioctl/mock.c:1:
+#include "mock.h"

ERROR: code indent should use tabs where possible
#125: FILE: test/ioctl/mock.c:77:
+^I        mock_cmd->metadata, \$

ERROR: code indent should use tabs where possible
#126: FILE: test/ioctl/mock.c:78:
+^I        cmd->metadata_len, \$

ERROR: code indent should use tabs where possible
#127: FILE: test/ioctl/mock.c:79:
+^I        "incorrect metadata"); \$

WARNING: Missing a blank line after declarations
#168: FILE: test/ioctl/mock.c:120:
+       bool result64;
+       switch (request) {

WARNING: Prefer using '"%s...", __func__' to using 'ioctl', this function's name, in a string
#186: FILE: test/ioctl/mock.c:138:
+               fail("unexpected ioctl %lu", request);

WARNING: Missing a blank line after declarations
#191: FILE: test/ioctl/mock.c:143:
+       const struct mock_cmd *mock_cmd = mock_cmds->cmds++;
+       mock_cmds->remaining_cmds--;
[...]
```

igaw: Don't get me wrong, I really do like this submission!

calebsander: I think I've made `checkpatch.pl` shut up, except for warnings about updating the `MAINTAINERS` file (which AFAICT is a Linux-specific concept). I'm all in support of consistent style.

igaw:
Thanks a lot for the style changes. I know it's painful but I think it makes the long time maintenance simpler. Really appreciate your contribution!

PR #689, opened 2023-08-04
nvme-tree: avoid segfault in 'list-subsys /dev/nvmeXnY'

martin-gpy:
While iterating through nvme_for_each_subsystem() in nvme_scan_subsystem() and after a subsystem match is found, one occasionally encounters a strcmp segfault further in the loop due to junk _s->name values showing up in certain error scenarios on a scaled up config. For e.g. with libnvme debug logging enabled:

nvme list-subsys /dev/nvme99n1
...
scan subsystem nvme-subsys0
filter out subsystem nvme-subsys0
NQN mismatch for subsystem 'nvme-subsys0'
failed to scan subsystem nvme-subsys0: Invalid argument scan subsystem nvme-subsys1
filter out subsystem nvme-subsys1
...

Avoid this segfault by breaking from the nvme_for_each_subsystem() loop once the subsystem match is found.

Fixes: fbd45f1 ("tree: Scan all subsystems")

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -693,6 +693,7 @@ static int nvme_scan_subsystem(struct nvme_root *r, const char *name,
 				continue;
 			s = _s;
 			__nvme_scan_subsystem(r, s, f, f_args);
+			break;
 		}
 	}
 	if (!s) {


igaw:
Unfortunately this will break multi hostnqn setups. We need to figure out what's going on here, as I understood that we should only operate on properly initialized data.

hreinecke:
Are we sure it's _s->name ?
From the logs I suspect it's more this code:

```
static void __nvme_scan_subsystem(struct nvme_root *r, nvme_subsystem_t s,
				  nvme_scan_filter_t f, void *f_args)
{
	if (f && !f(s, NULL, NULL, f_args)) {
		nvme_msg(r, LOG_DEBUG, "filter out subsystem %s\n", s->name);
		__nvme_free_subsystem(s);
		return;
	}
	nvme_subsystem_scan_namespaces(r, s, f, f_args);
}

```
which really is the culprit; when the filter triggers the subsystem 's' is freed, with no indication to the caller.

PR #692, opened 2023-08-08
Change handles from file-descriptors to 'struct dev_handle'

vikash-k5:
api: refactor for transport agnostic library backend

Extend the api-types with the following device-handle types:

* struct dev_handle
* enum nvme_dev_type

The 'dev_handle' provides a, possibly opaque, device handle instead of a "fixed" file-descriptor. This allows for non-OS managed device-types such as user-space NVMe-drivers. Additionally, the types allows for library-side dispatch.

This is done in preparation for xNVMe, and thereby, support for
user-space NVMe driver, io_uring_cmd etc.

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -57,6 +57,8 @@ LIBNVME_MI_1_1 {
 		nvme_mi_open_mctp;
 		nvme_mi_scan_mctp;
 		nvme_mi_scan_ep;
+		get_dev;
+		dev_close;
 	local:
 		*;
 };
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -260,7 +260,7 @@ LIBNVME_1_0 {
 		nvme_ns_get_csi;
 		nvme_ns_get_ctrl;
 		nvme_ns_get_eui64;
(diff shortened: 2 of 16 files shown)


calebsander:
This changes the public interface of a huge portion of the libnvme API. I'm not familiar with the "xNVMe" use case you mention, but it seems like it would be good to avoid having to make such widespread changes to users of libnvme. If you need to generalize the notion of a file descriptor, could you add new versions of the functions taking in the new type while leaving the original ones intact? The old functions could call the new functions under the hood, but I think it would be good to keep the existing interface intact as much as possible.

igaw:
@calebsander These changes are RFC status only and planed for the next major version update of the library. So this will not land in the 1.x branch. We had some offline discussion on these ideas and decided to have some early PR to discuss the API. So don't fear, we wont break 1.x :)

For example this PR does only a minimal change necessary to introduce the abstract handle. I think we should go bigger, e.g. see https://github.com/linux-nvme/libnvme/issues/448

This looks like an excellent topic for the next ALPSS conference :)

igaw:
Let's close this one. It is libnvme v2 material anyway.

PR #693, opened 2023-08-08
tree: fix segfault in nvme_scan_subsystem()

martin-gpy:
The wrong nvme_subsystem struct was being passed to __nvme_subsystem_scan() which caused it to segfault. Fix it.

Fixes: d08fd10 ("make __nvme_scan_subsystem() returning bool")


igaw:
Thanks!

PR #695, opened 2023-08-14
tree: Improve ctrl lookup algorithm for TCP

martin-belanger:
As previously reported [1], there are cases where libnvme may identify an existing controller as matching a candidate controller when that is not really the case. In other cases, libnvme may fail to identify an existing controller as matching a candidate controller. This patch fixes the issue.

The problem only manifests itself when the parameters host-traddr and/or host-iface are used. There is no problem when these parameters are NULL for both the existing controller and the candidate controller. Several users have reported issues when setting host-traddr and/or host-iface to non-NULL values.

The success of the new matching algorithm depends on whether the kernel exposes the src_addr through the sysfs (kernel 6.1 or later). When the src_addr is exposed by the kernel, the new algorithm has a 100% success rate. If the src_addr is not available, a few improvements have been made to improve the success rate, but it's not 100%.

Note that to simplify the code and to make it more readable, the function used for matching controllers has been broken into several smaller functions. There is one function per transport type. It's only for TCP transport that additional tests are required. The remaining transport types have little to no changes.

[1] http://lists.infradead.org/pipermail/linux-nvme/2023-June/040295.html


src/nvme/tree.c
```diff
@@ -1194,40 +1240,264 @@ struct nvme_ctrl *nvme_create_ctrl(nvme_root_t r,
...
+	return true;
+}
+
+static bool no_match(struct nvme_ctrl *c, struct candidate_args *candidate)
+{
+	return false;
+}
+
+typedef bool (*ctrl_match_t)(struct nvme_ctrl *, struct candidate_args *);
+static ctrl_match_t ctrl_matcher(const char *transport)
```
igaw, line 1444: `get_ctrl_matcher` ?

igaw (changes requested):
Please split the change into two changes. A feature patch (you could even split this into a refactoring part (e.g. _nvme_ipaddrs_eq) and a test patch. 

About the test patch: I would prefer if the existing test written in C is actually extended instead introducing another test written in Python. I don't mind having additional tests in Python but I would like to have the C code also testing this new feature. Reason being, the cross builds do not build/test the Python code at this point.

test/tree.c
```diff
@@ -235,19 +235,19 @@ static void test_lookup_2(void)
 
 	c2 = nvme_lookup_ctrl(s, "tcp", "192.168.1.1", "192.168.1.20",
 			      "eth0", NULL, NULL);
-	assert(c1 == c2);
+	assert(c1 != c2);
```
igaw, line 238: The changes need some more documentation why this behavior change here is correct. 

As far I understood @hreinecke and also @martin-gpy  the above test case is what expected to happen until now.  But with the new lookup implementation, this isn't the case anymore.

igaw (changes requested):

src/nvme/tree.c
```diff
@@ -1194,40 +1240,264 @@ struct nvme_ctrl *nvme_create_ctrl(nvme_root_t r,
...
+	/* traddr and trsvcid must match exactly */
+	if (!streq0(c->trsvcid, candidate->trsvcid) ||
+	    !nvme_ipaddrs_eq(c->traddr, candidate->traddr))
+		return false;
+
+	/* Does the candidate care about host_traddr / host_iface?
+	 * If not, we can skip the check.
+	 */
+	if (candidate->host_traddr || candidate->host_iface) {
+		char *src_addr, buffer[100]; /* big enough for IPv6 max length */
```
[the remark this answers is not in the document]
martin-belanger: Will do. Thanks.

test/tree.c
martin-belanger: Yep! That is exactly what I'm trying to solve. When a candidate wants a specific host-traddr and/or host-iface we need to make sure we match perfectly and not just return any controller that is "close enough".

martin-belanger:
Closing this PR. I will resubmit separate PRs that will address all the review comments.

PR #697, opened 2023-08-16
tree, util: Preliminary work for improved TCP ctrl lookup 

martin-belanger:
As discussed in https://github.com/linux-nvme/libnvme/pull/695, this is a small PR to prepare for a larger PR that will introduce an improved TCP controller lookup algorithm.

### DIFF
--- a/test/tree.c
+++ b/test/tree.c
@@ -6,10 +6,13 @@
 
 #include <assert.h>
 #include <string.h>
+#include <stdbool.h>
+#include <stdlib.h>
 
 #include <ccan/array_size/array_size.h>
 
 #include <libnvme.h>
+#include <nvme/private.h>
 
 struct test_data {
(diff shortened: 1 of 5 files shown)


martin-belanger:
The failure to build seems to be a GitHub failure and not something that I changed. I can't find a way to restart the workflow (maybe only maintainers can do that).

test/tree.c
```diff
@@ -254,10 +258,96 @@ static void test_lookup_2(void)
 	nvme_free_tree(r);
 }
 
+bool test_src_addr()
```
igaw, line 261: Thanks for the test. Could you split this patch into the function adding and the test? It makes the back porting simpler (in cases where we still don't have the test infrastructure backported :))

igaw:
Thanks a lot!

PR #698, opened 2023-08-18
tree: Improve tcp controller matching algorithm

martin-belanger:
This PR addresses the review comments in this closed PR https://github.com/linux-nvme/libnvme/pull/695

### DIFF
--- a/test/tree.c
+++ b/test/tree.c
@@ -8,6 +8,7 @@
 #include <string.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <arpa/inet.h>
 
 #include <ccan/array_size/array_size.h>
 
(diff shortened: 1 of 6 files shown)


test/tree.c
```diff
@@ -240,19 +241,19 @@ static void test_lookup_2(void)
 
 	c2 = nvme_lookup_ctrl(s, "tcp", "192.168.1.1", "192.168.1.20",
 			      "eth0", NULL, NULL);
-	assert(c1 == c2);
+	assert(c1 != c2);
```
igaw, line 244: After a bit of pondering, I think this is just failing because `nvme_lookup_ctrl` is not able to lookup the IP configuration of the host via `getifaddrs`. And the new test case `test_ctrl_match_tcp` doesn't work on my system either. I suspect due to the same reason.

And another thing is that looking at `test_lookup_1` and `test_lookup_2` are doing not testing the right thing. Instead testing the host_traddr/host_iface variation they test the host_iface/trsvcid combination. As you explained in the excellent commit message we trsvcid is not optional for tcp.

Thus, I think we should 
- update/fix `test_lookup_1` and `test_lookup_2` so it tests the host_* combination for tcp and maybe have an individual test function for the different transports where these two arguments don't play a role.
- introduce a `getifaddrs` mock like we have for the `ioctl` so we do not depend on the host configuration at all.

What do you think?

martin-belanger: > * update/fix `test_lookup_1` and `test_lookup_2` so it tests the host_* combination for tcp and maybe have an individual test function for the different transports where these two arguments don't play a role.
> * introduce a `getifaddrs` mock like we have for the `ioctl` so we do not depend on the host configuration at all.
> 
> What do you think?

Good idea! I'll work on that.

martin-belanger: I removed `test_lookup_2()` and replaced by transport-specific tests: `test_ctrl_match_fc()`, `test_ctrl_match_rdma()`, `test_ctrl_match_tcp()`.

test/tree.c
```diff
@@ -405,14 +359,636 @@ bool test_src_addr()
...
+	/* Reference ID 2 */
+	set_ctrl_args(&reference, "fc", "21:00:00:e0:8b:05:05:01", "4420", NULL, NULL, NULL);
+
+	set_ctrl_args(&candidate, "fc", "21:00:00:e0:8b:05:05:01", "4420", "21:00:00:e0:8b:05:05:20", NULL, NULL);
+	pass &= ctrl_match("FC", 2, 0, &reference, &candidate, true);
+
+	set_ctrl_args(&candidate, "fc", "21:00:00:e0:8b:05:05:01", "4420", NULL, NULL, NULL);
+	pass &= ctrl_match("FC", 2, 1, &reference, &candidate, true);
+
+	set_ctrl_args(&candidate, "fc", "21:00:00:e0:8b:05:05:01", "4420", NULL, "eth0", NULL);
```
igaw, line 504: Do these FC tests make sense? I mean the values for `trsvcid` and `host_iface`. Obviously, the libnvme doesn't care about these values but the kernel will. This is just a minor issue. I'll go ahead and merge it. I've updated my old test case and played a bit with it, to get a better understanding what's happening.

martin-belanger: Thanks for reviewing my PR. I provided these FC tests to make sure things don't blow up when passing illegal values and/or parameters. 

BTW, this is one of the reasons I provided transport-specific code in my earlier PR (ref: https://github.com/linux-nvme/libnvme/pull/695/files#diff-f131da46870af4738f860f74d79b2b4b8f8afd9900d5ea8e7956a37550c56c32R1421). The intention was to make it clear which parameters apply to each transport type and only test those parameters.

igaw: Ah okay. I was just a bit confused when reading through this section.

PR #704, opened 2023-08-30
Account for discovery log page entry space stripping and simplify space stripping code

calebsander:
`strchomp()` is only needed to strip trailing spaces from a string.
There is no need to replace `NUL` characters with `NUL` characters.
So simplify the check for "NUL or space" to just "space".
Update the function description as well,
as it doesn't strip all whitespace, just spaces.

The NVMe base specification defines fields TRSVCID and TRADDR
as "ASCII strings", meaning they are space-padded.
Therefore, whether spaces need to be stripped from them
doesn't depend on the transport or address family.
Some combinations already appear to be missing,
such as RDMA using IB addresses.
And if new transports are added in the future,
`sanitize_discovery_log_entry()` would need to be updated.
So strip spaces from these fields regardless of the entry's
transport type or address family.

The discovery log page tests incorrectly assumed the log page
`nvmf_get_discovery_log()` returns exactly matches what was fetched.
This failed to account for the trailing space trimming
of the TRSVCID and TRADDR fields.
If the arbitrary entries happened to contain trailing spaces
for these fields, the test cases would fail.
Generate arbitrary entries with space-padded strings for the log page
and corresponding expected entries with `NUL`-padded strings.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1166,30 +1166,10 @@ out_free_log:
 	return NULL;
 }
 
-static void sanitize_discovery_log_entry(struct nvmf_disc_log_entry  *e)
+static void sanitize_discovery_log_entry(struct nvmf_disc_log_entry *e)
 {
-	switch (e->trtype) {
-	case NVMF_TRTYPE_RDMA:
-	case NVMF_TRTYPE_TCP:
-		switch (e->adrfam) {
-		case NVMF_ADDR_FAMILY_IP4:
-		case NVMF_ADDR_FAMILY_IP6:
-			strchomp(e->traddr, NVMF_TRADDR_SIZE);
-			strchomp(e->trsvcid, NVMF_TRSVCID_SIZE);
-			break;
-		}
-		break;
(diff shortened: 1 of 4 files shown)


src/nvme/fabrics.c
```diff
@@ -47,19 +47,16 @@
...
+ * strchomp() - Strip trailing spaces
  * @str: String to strip
  * @max: Maximum length of string
  */
 static void strchomp(char *str, int max)
 {
 	int i;
 
-	for (i = max - 1; i >= 0; i--) {
-		if (str[i] != '\0' && str[i] != ' ')
```
igaw, line 59: I digged into the history and found the source of the NULL check:

e48ba33d19ae ("fabrics: Handle space-padded TRSVCID and TRADDR fields")
https://github.com/linux-nvme/nvme-cli/commit/e48ba33d19ae

I suppose  @rolandd is fine with this change ;)

igaw:
Sorry that it took so long. Busy weeks and I had to fix the CI build first...

PR #712, opened 2023-09-05
util: remove unused nvme_status_to_errno function

sc108-lee:
this function is not used anymore
related nvme-cli modification below
https://github.com/linux-nvme/nvme-cli/commit/9643090317e8eef5b1f9b2c9abda18e06b3e5902 https://github.com/linux-nvme/nvme-cli/pull/2045


igaw:
Unfortunately, we can't remove this function just because nvme-cli doesn't use it anymore. There might be other users of this function. We can remove it when we do a 2.x major version update of the library but for the 1.x branch we have to leave it in.

Issue #684, opened 2023-07-26
stack smashing detected with 'nvme sanitize-log /dev/nvme0'


igaw:
There are several reports that the sanitize log command is able to smash the stack somehow. 

https://github.com/storaged-project/udisks/issues/1152

For tracking and awareness reasons, I'll open here an issue here too.

igaw:
FWIW, I am not able to reproduce it and I really wonder how it is actually possible to achieve this. It looks all sane when poking with gdb at the addresses and alignments.

So next I'd like to check if there is a way when the firmware misbehaves, it could result in overwriting the stack. I assume this should be really really hard to do but who knows. @keithbusch in case you have some input on this, it would highly appreciated.

igaw:
With linux-nvme/nvme-cli#2051 merged, this problem should be solved.

gyakovlev:
Hi, are you planning tagging release anytime soon?
this is a bit awkward to backport to 1.5 and it affects libblockdev and udisks a lot.

ReillyBrogan:
> Hi, are you planning tagging release anytime soon? this is a bit awkward to backport to 1.5 and it affects libblockdev and udisks a lot.

On Solus we encountered several issues after updating libblockdev/udisks and we have all of them resolved with [this libnvme patch](https://github.com/linux-nvme/libnvme/commit/c56910f807795528fff7ba6b81f8efcdb4babe98) and the latest release versions of libblockdev and udisks. We currently have no reported udisks crash issues, so we're not going to bother backporting the linked PR and will instead wait for it to be in a release.

bulletmark:
@ReillyBrogan I experience this repeatable bug on Arch Linux. I built a new libnvme package with that patch applied and find it does not fix this issue.

Note: Arch updates just prior to me seeing this bug were:
```
2023-09-21 07:23:34 libnvme            1.5-1 installed
2023-09-21 07:23:34 libblockdev        2.28-4 -> 3.0.3-4
2023-09-21 07:23:34 udisks2            2.9.4-4 -> 2.10.1-1
```
Since then:
```
$ journalctl -qg 'stack smashing detected'
Sep 21 07:27:30 lt udisksd[1244]: *** stack smashing detected ***: terminated
Sep 22 07:27:24 lt udisksd[1250]: *** stack smashing detected ***: terminated
Sep 22 07:28:50 lt udisksd[859]: *** stack smashing detected ***: terminated
[...]

Issue #681, opened 2023-07-20
Do not issue commands when scaning sysfs


igaw:
As @hreinecke  pointed out there is no need to issues any commands when scanning sysfs. This discussion came up during https://github.com/linux-nvme/libnvme/issues/672. So the aim is to use all the infos only from sysfs.

PR #713, opened 2023-09-06
Add quirks framework to libnvme

igaw:
We had a bunch of bug reports that all were related to misbehaving devices. The world is not perfect so we have to deal with this somehow. I think introducing some sort of quirk database is a reasonable way forward.

@keithbusch pointed out it would be good if the database would be loaded from a file. I decided to use the already existing json infrastructure for this.

The loading happens with the elf library constructor hook because I wanted to make sure the data is also available for the ioctl calls but I don't know if it's worth trying to these problems solved on this level. Maybe we should just use it with the tree API where we have a `nvme_root_t` object. This would also make some error logging possible. 

So this is just the first rough version to gather some feedback.

https://github.com/linux-nvme/nvme-cli/issues/2029
https://github.com/storaged-project/libblockdev/pull/951
https://github.com/linux-nvme/nvme-cli/issues/2029

Fixes: https://github.com/linux-nvme/libnvme/issues/684

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -34,6 +34,7 @@
 #include "fabrics.h"
 #include "log.h"
 #include "private.h"
+#include "quirks.h"
 
 /**
  * struct candidate_args - Used to look for a controller matching these parameters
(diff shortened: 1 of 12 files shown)


src/nvme/tree.c
```diff
@@ -2309,7 +2330,8 @@ static int nvme_ns_init(struct nvme_ns *n)
 	n->lba_util = le64_to_cpu(ns.nuse);
 	n->meta_size = le16_to_cpu(ns.lbaf[flbas].ms);
 
-	if (!nvme_ns_identify_descs(n, descs))
+	if (!(n->c->quirk_flags & NVME_QUIRKS_NS_ID_DESC_LIST) &&
```
igaw, line 2333: This doesn't work because `n->c` is not valid yet. The controller is to the namespace after `nvme_ns_init` has been called when getting called from `nvme_scan_topology`. But there are other callers such as `nvme_scan_namespace` which only have an string as argument. 

Hmm, maybe we should just try to get rid of all ioctl calls in the `nvme_scan_topology` call chain. See also https://github.com/linux-nvme/libnvme/issues/681

igaw: We discussed this during ALPSS and yep, let's ditch all commands from the scan_topology call and completely rely on sysfs.

igaw:
We decided to stop issuing any commands during the topology scan and rely completely on sysfs. There are a couple of attributes missing right now but I am working on it. So no point to keep this open.

PR #714, opened 2023-09-07
linux: Added functions to enable faster telemetry data retrieval.

lgdacunh:
Moved telemetry data area support detection into separate function. Added possibility to modify data transfer chunk size. Enable telemetry extraction up to specified data area.

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -122,17 +122,44 @@ int nvme_fw_download_seq(int fd, __u32 size, __u32 xfer, __u32 offset,
 	return err;
 }
 
-static int nvme_get_telemetry_log(int fd, bool create, bool ctrl, bool rae,
-				  struct nvme_telemetry_log **buf, enum nvme_telemetry_da da,
-				  size_t *size)
+int nvme_get_telemetry_max(int fd, enum nvme_telemetry_da *da, size_t *data_tx)
+{
+	struct nvme_id_ctrl id_ctrl;
+	int err = nvme_identify_ctrl(fd, &id_ctrl);
+
+	if (err)
+		return err;
+
+	if (data_tx) {
+		*data_tx = id_ctrl.mdts;
+		if (id_ctrl.mdts) {
+			/* assuming CAP.MPSMIN is zero minimum Memory Page Size is at least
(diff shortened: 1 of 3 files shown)


src/nvme/linux.c
```diff
@@ -122,17 +122,47 @@ int nvme_fw_download_seq(int fd, __u32 size, __u32 xfer, __u32 offset,
...
-static int nvme_get_telemetry_log(int fd, bool create, bool ctrl, bool rae,
-				  struct nvme_telemetry_log **buf, enum nvme_telemetry_da da,
-				  size_t *size)
+enum nvme_telemetry_da nvme_get_telemetry_max_da(int fd, size_t *max_data_tx)
+{
+	struct nvme_id_ctrl id_ctrl;
+	int err = nvme_identify_ctrl(fd, &id_ctrl);
+
+	if (err) {
+		perror("identify-ctrl");
```
igaw, line 131: Unfortunately, we can't use printf & Co in the library.

src/nvme/linux.c
```diff
@@ -122,17 +122,47 @@ int nvme_fw_download_seq(int fd, __u32 size, __u32 xfer, __u32 offset,
...
+		errno = EINVAL;
+		return 0;
```
igaw, line 133: I think we should stay with the current pattern and just return the error status and have the a function argument for the result, as we have with the `max_data_tx` argument.

src/nvme/linux.c
lgdacunh: Removed!

src/nvme/linux.c
lgdacunh: Done as suggested!

src/nvme/linux.c
```diff
@@ -207,6 +227,14 @@ static int nvme_get_telemetry_log(int fd, bool create, bool ctrl, bool rae,
 		goto free;
 	}
 
+	if (dalb == 0) {
+		fprintf(stderr, "ERROR: No telemetry data block");
```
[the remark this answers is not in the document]
lgdacunh: Done as suggested!

igaw:
Thanks!

PR #716, opened 2023-09-12
tree: Add 2 new public functions to lookup existing controllers

martin-belanger:
There is duplicate code between **libnvme** and **nvme-cli**. When trying to instantiate a controller, both libnvme and nvme-cli scan the sysfs looking for an existing controller that can be reused.

This patch adds `nvme_ctrl_find()` and `nvme_ctrl_config_match()`. These functions can be used by nvme-cli to lookup existing controllers. This will allow removing the duplicate code in nvme-cli (separate commit to follow on the nvme-cli GitHub repo).

Note that some of the existing code had to be reorganized to make it possible for the new functions to reuse that existing code.

### DIFF
--- a/src/nvme/private.h
+++ b/src/nvme/private.h
@@ -180,7 +180,7 @@ int json_dump_tree(nvme_root_t r);
 nvme_ctrl_t __nvme_lookup_ctrl(nvme_subsystem_t s, const char *transport,
 			       const char *traddr, const char *host_traddr,
 			       const char *host_iface, const char *trsvcid,
-			       nvme_ctrl_t p);
+			       const char *subsysnqn, nvme_ctrl_t p);
 
 #if (LOG_FUNCNAME == 1)
 #define __nvme_log_func __func__
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 LIBNVME_1_6 {
 	global:
+		nvme_ctrl_config_match;
+		nvme_ctrl_find;
 		nvme_ctrl_get_src_addr;
 		nvme_ctrl_release_fd;
 		nvme_host_release_fds;
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -834,6 +834,7 @@ static const char *lookup_context(nvme_root_t r, nvme_ctrl_t c)
 					       NULL,
 					       NULL,
 					       nvme_ctrl_get_trsvcid(c),
+					       NULL,
 					       NULL))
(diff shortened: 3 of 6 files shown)


src/nvme/private.h
```diff
@@ -180,7 +180,8 @@ int json_dump_tree(nvme_root_t r);
 nvme_ctrl_t __nvme_lookup_ctrl(nvme_subsystem_t s, const char *transport,
 			       const char *traddr, const char *host_traddr,
 			       const char *host_iface, const char *trsvcid,
-			       nvme_ctrl_t p);
+			       nvme_ctrl_t p, const char *subsysnqn,
+			       bool dc_only);
```
igaw, line 184: `dc_only` could be an enum type which makes it a bit simpler to read on the caller side IMO.

martin-belanger: That's an interesting idea. I didn't really like this parameter in the first place, but we're a bit stuck. The issue is that for the longest time DCs only had the well-known NQN (nqn.2014-08.org.nvmexpress.discovery). This meant that by looking at the subsysnqn we could easily determine whether we were dealing with an I/O or Discovery controller.

With TP8013, however, DCs can now have a unique NQN. On top of that, one can create a connection to a DC using the well-known NQN, but during the identify phase, the NQN reported through the sysfs can change to a unique NQN. So the config subsysnqn may not match the NQN reported by the sysfs. There are other attributes that can be read from the sysfs to make differentiate a discovery from an I/O controller.

Long story short, we could indeed use an enum here. I can make the change.

martin-belanger: Actually, this can be simplified. We don't need an additional parameter to identify the controller type. I went down this path because of how it was done in nvme-cli, but the reality is that we only need to use special handling when dealing with the well-known NQN (`"nqn.2014-08.org.nvmexpress.discovery"`). For all other cases, the unique `subsysnqn` is all we need to worry about. 

Let me clean this up and resubmit.

src/nvme/private.h
```diff
@@ -180,7 +180,7 @@ int json_dump_tree(nvme_root_t r);
 nvme_ctrl_t __nvme_lookup_ctrl(nvme_subsystem_t s, const char *transport,
 			       const char *traddr, const char *host_traddr,
 			       const char *host_iface, const char *trsvcid,
-			       nvme_ctrl_t p);
+			       nvme_ctrl_t p, const char *subsysnqn);
```
igaw, line 183: Isn't the `p` argument used as starting point of iterator? In this case I would swap the order of `p` and `subsysnqn` so that `p` stays at the last position.

martin-belanger: Done

igaw:
Looks good. Thanks!

PR #717, opened 2023-09-15
Fixes and unit tests for Get/Set Features functions

calebsander:
The functions for sending Set Features and Get Features commands are pretty buggy.
That's probably because `nvme-cli` only calls `nvme_set_features()` and `nvme_get_features()` directly.
I have no idea if anyone is actually using these functions, but may as well fix them if they exist and add tests.


igaw:
The compiler did complain for `test/test.c`. The CI builds report it but they don't fail. Have to figure out why.

```shell
 ../test/test.c:201:9: error: ‘nvme_get_features_err_recovery’ is deprecated [-Werror=deprecated-declarations]
  201 |         ret = nvme_get_features_err_recovery(fd, sel, &result);
      |         ^~~
In file included from ../src/nvme/linux.h:14,
                 from ../src/libnvme.h:18,
                 from ../test/test.c:23:
../src/nvme/ioctl.h:2762:5: note: declared here
 2762 | int nvme_get_features_err_recovery(int fd, enum nvme_get_features_sel sel,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../test/test.c:261:9: error: ‘nvme_get_features_resv_mask’ is deprecated [-Werror=deprecated-declarations]
  261 |         ret = nvme_get_features_resv_mask(fd, sel, &result);
      |         ^~~
../src/nvme/ioctl.h:3068:5: note: declared here
 3068 | int nvme_get_features_resv_mask(int fd, enum nvme_get_features_sel sel,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
../test/test.c:266:9: error: ‘nvme_get_features_resv_persist’ is deprecated [-Werror=deprecated-declarations]
  266 |         ret = nvme_get_features_resv_persist(fd, sel, &result);
      |         ^~~
../src/nvme/ioctl.h:3097:5: note: declared here
 3097 | int nvme_get_features_resv_persist(int fd, enum nvme_get_features_sel sel,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Anyway, I fixed up the test myself. The rest looks good. Thanks!

calebsander:
Hmm, specifying a NSID of 0 doesn't really make sense, probably those usages should be updated to use an actual NSID. But thanks for taking a look at the code!

igaw:
I see. I could not test it, no hardware which supports these features. Don't know if anyone is actually running this test anymore. It really looks like it was there to get some testing done. Your unit tests are way better :). Maybe we could just ditch it. We can't run in a CI context anyway, it as it needs root rights and real hardware.

calebsander:
Well there is probably some value in testing against real hardware. My unit tests are only as bug-free as my reading/understanding of the spec. And it's always possible devices have non-spec-compliant implementations.
That being said, those tests probably aren't too valuable without any validatation of the output.

igaw:
I still hope to find some time to work on getting a test setup where we run some tests, e.g. blktests against real hardware. blktests is using exercising some parts of the nvme-cli/libnvme code base. I suppose if I get this working we could run this test also. On a second thought let's keep it and rather figure out how to get CI tests against real hardware.

PR #721, opened 2023-09-23
mi-mctp: Fix free() in error path of mi_open_mctp

jk-ozlabs:
If the malloc of our struct nvme_mi_transport_mctp fails, we'll attempt to free ->rsp_buf of this (now zero) pointer.

Instead, structure the error path to progressively undo the initialisation operations. This means we'll need to save the errno at the site of each possible failure.

In doing this, add a comment to the call to nvme_mi_close(), just to clarify behaviour with regards to the cleanups through that path.

Reported-by: Barnabás Pőcze <pobrn@protonmail.com>


igaw:
Thanks!

PR #724, opened 2023-09-30
test: handle POSIX ioctl prototype

thesamesam:
glibc has the following prototype for ioctl: int ioctl(int fd, unsigned long request, ...)
POSIX (inc. musl) has the following for ioctl: int ioctl(int fd, int request, ...)

Check which prototype is used in <sys/ioctl.h> to avoid a conflict and conditionally
define the right one for the system.

Bug: https://bugs.gentoo.org/914921
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -273,7 +273,9 @@ subdir('internal')
 subdir('ccan')
 subdir('src')
 subdir('libnvme')
-subdir('test')
+if get_option('tests')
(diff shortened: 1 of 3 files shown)


igaw:
Thanks!

thesamesam:
Thank you!

PR #726, opened 2023-10-10
ioctl: MSB variable-size storage/reference tags

bpaupore-wdc:
The spec defines these values on a bitwise basis within the shared 80-bit region, stored from MSB to LSB for each tag. Mechanically this can be achieved by setting the values as big-endian using the same logic currently in place.

Reviewed-by: Jeffrey Lien <jeff.lien@wdc.com>

Here's a link to discussion around fixing a similar issue with the copy command: https://github.com/linux-nvme/libnvme/pull/587


igaw:
Thanks!

PR #727, opened 2023-10-10
tree: Allocate aligned payloads for namespace scan

tbzatek:
Related to https://github.com/linux-nvme/nvme-cli/pull/2051, there are a couple more ioctls made during the tree scan inside of libnvme. Last bits to fix the whole story, hopefully.

Noticed by https://bugs.archlinux.org/task/79760#comment222718


igaw:
Thanks!

igaw:
IIRC, we have a few more places where we should use the newly introduced `__nvme_alloc` helper.

PR #730, opened 2023-10-12
make: add checkpatch to Makefile

martin-belanger:
@igaw - Hi Daniel. I noticed that you are asking people to run the `checkpatch.pl` script to make sure the code follows the same coding style as the kernel. To make it easier for developers to check their code, I added a new command to the `Makefile` (i.e. `make checkpatch`). Hopefully you will find this useful. Otherwise, please ignore.

This automatically retrieves the `checkpatch.pl` script from the Linux kernel repo and runs it to check the modified or untracked source files in the current repo.

The command can be invoked with:
```
$ make checkpatch
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,3 +48,19 @@ rpm: ${BUILD-DIR}
 	tar rf libnvme.tar ${BUILD-DIR}/libnvme.spec
 	gzip -f -9 libnvme.tar
 	rpmbuild -ta libnvme.tar.gz -v
+
+# Retrieve and Invoke Linux kernel's checkpatch script on modified source files
+ifeq (checkpatch,$(strip $(filter $(MAKECMDGOALS),checkpatch)))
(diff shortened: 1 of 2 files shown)


igaw:
I don't mind something like this to the Makefile, though it might be simpler just to add it to scripts/build.sh (or scripts/check.sh and call this from the Makefile). 

Though I think it would also be nice if it could support my main use case, which is checking all new commits. That could also be useful as CI step. Just saying :)

Do you think this is feasible?

martin-belanger:
> Do you think this is feasible?
Ok. I will look into it. But if I'm not mistaking, the code is not 100% "clean". Adding this to the CI will require cleaning all of the code.

martin-belanger:
@igaw - I found a GitHub action that performs the checkpatch. 

https://github.com/webispy/checkpatch-action

I will try that instead of the changes I made to the Makefile. 

I'm closing this pull request.

PR #731, opened 2023-10-12
Aligned payloads allocation (continued)

tbzatek:
Couple more obvious alloc changes, mostly about `struct nvme_id_ctrl` and `struct nvme_id_ns`.

Received only minimal testing.


igaw:
Thanks. 

I saw that we have a few of those in mi.c but I think these might not be changed. @jk-ozlabs do you know of any constrains on the payload alignment?

jk-ozlabs:
The mi paths have no requirements on alignment of buffers, we're interacting with a socket rather than passing buffers to `ioctl`. We'll generally end up adding headers anyway too.

PR #732, opened 2023-10-14
Make ASAN work with tests mocking libc functions and fix several errors caught by UBSAN

calebsander:
Testing out ASAN revealed that the new tests using `LD_PRELOAD` with a shared library containing mocks for libc functions cause ASAN to abort at startup. Suppress this check because ASAN seems to have no problems catching memory issues in `libnvme`, the tests, or the mocks.
Also fix several issues caught by enabling UBSAN in the unit tests:
- Passing NULL pointers to `memcmp()` and `memset()`
- Dereferencing unaligned pointers
- Overflow from bit shifts on `int` values that should be `u32`s

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,3 +178,5 @@ Note that when using the sanitize feature, the library `libasan.so` must be avai
 ```
 meson setup .build -Db_sanitize=address && LD_PRELOAD=/lib64/libasan.so.6 ninja -C .build test
 ```
+
+It's also possible to enable the undefined behavior sanitizer with `-Db_sanitize=undefined`. To enable both, use `-Db_sanitize=address,undefined`.
--- a/README.md
+++ b/README.md
@@ -171,7 +171,9 @@ To enable address sanitizer (advanced debugging of memory issues):
 meson setup .build -Db_sanitize=address
 ```
 
-This option adds `-fsanitize=address` to the gcc options. Note that when using the sanitize feature, the library `libasan.so` must be available and must be the very first library loaded when running an executable. Ensuring that `libasan.so` gets loaded first can be achieved with the `LD_PRELOAD` environment variable as follows: 
+This option adds `-fsanitize=address` to the gcc options. The tests can then be run normally (`meson test -C .build`).
+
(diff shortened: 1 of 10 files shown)


src/nvme/ioctl.c
```diff
@@ -525,7 +525,7 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
 int nvme_set_features(struct nvme_set_features_args *args)
 {
 	__u32 cdw10 = NVME_SET(args->fid, FEATURES_CDW10_FID) |
-			NVME_SET(!!args->save, SET_FEATURES_CDW10_SAVE);
+			NVME_SET((__u32)!!args->save, SET_FEATURES_CDW10_SAVE);
```
igaw, line 528: Wouldn't it make sense to fix the `NVME_SET` macro to handle this correctly? I really do not like sprinling around  casts in the code base.

src/nvme/ioctl.c
```diff
@@ -575,7 +575,7 @@ int nvme_set_features_arbitration(int fd, __u8 ab, __u8 lpw, __u8 mpw,
 	__u32 value = NVME_SET(ab, FEAT_ARBITRATION_BURST) |
 			NVME_SET(lpw, FEAT_ARBITRATION_LPW) |
 			NVME_SET(mpw, FEAT_ARBITRATION_MPW) |
-			NVME_SET(hpw, FEAT_ARBITRATION_HPW);
+			NVME_SET((__u32)hpw, FEAT_ARBITRATION_HPW);
```
igaw, line 578: e.g. here we have two `NVME_SET` without a cast and the last one with one.

igaw (commented):
Nitpick: the subject prefix should be the file name I think.

src/nvme/ioctl.c
calebsander: That's a good idea. is `NVME_SET()` assumed to always operate on `u32` values?

igaw: Not sure. Maybe we can peek into the Linux kernel see how the handle this. I am sure we are not the first one running into this issue :)

igaw:
Just splitted the last two patches, because I think it groups the changes a bit better and thus easier to spot the what's going on.

igaw:
Thanks!

PR #733, opened 2023-10-20
nvme-copy: add cross-namespace copy formats, status codes, ONCS bits

jgu-pure:
Add support for NVMe TP4130 ("Cross-Namespace Copy"):
- Add Copy Descriptor Formats 2h and 3h
- Add new status codes for Copy: Incompatible Namespace or Format, Fast Copy Not Possible, Overlapping I/O Range, and Insufficient Resources
- Add two new ONCS bits NVMCSA and NVMAFC
- Add Copy Descriptor Formats Enable (CDFE) to Host Behavior Support
  Data Structure

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -1506,6 +1506,14 @@ enum nvme_id_ctrl_cqes {
  *					the Verify command.
  * @NVME_CTRL_ONCS_COPY:		If set, then the controller supports
  *					the copy command.
+ * @NVME_CTRL_ONCS_COPY_SINGLE_ATOMICITY: If set, then the write portion of a
+ *					Copy command is performed as a single
+ *					write command to which the same
+ *					atomicity requirements that apply to
+ *					a write command apply.
+ * @NVME_CTRL_ONCS_ALL_FAST_COPY:	If set, then all copy operations for
+ *					the Copy command are fast copy
+ *					operations.
  */
 enum nvme_id_ctrl_oncs {
 	NVME_CTRL_ONCS_COMPARE			= 1 << 0,
@@ -1517,6 +1525,8 @@ enum nvme_id_ctrl_oncs {
 	NVME_CTRL_ONCS_TIMESTAMP		= 1 << 6,
 	NVME_CTRL_ONCS_VERIFY			= 1 << 7,
 	NVME_CTRL_ONCS_COPY			= 1 << 8,
+	NVME_CTRL_ONCS_COPY_SINGLE_ATOMICITY	= 1 << 9,
+	NVME_CTRL_ONCS_ALL_FAST_COPY		= 1 << 10,
 };
 
 /**
@@ -4734,10 +4744,14 @@ struct nvme_plm_config {
  * struct nvme_feat_host_behavior - Host Behavior Support - Data Structure
  * @acre:	Advanced Command Retry Enable
  * @rsvd1:	Reserved
+ * @cdfe:       Copy Descriptor Formats Enable
+ * @rsvd6:	Reserved
  */
 struct nvme_feat_host_behavior {
 	__u8 acre;
-	__u8 rsvd1[511];
+        __u8 rsvd1[3];
+        __u16 cdfe;
+	__u8 rsvd6[506];
 };
 
 /**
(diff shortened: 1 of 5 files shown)


src/nvme/types.h
```diff
@@ -1485,27 +1485,35 @@ enum nvme_id_ctrl_cqes {
...
+ *					  value in the Get Features command.
+ * @NVME_CTRL_ONCS_RESERVATIONS:	  If set, then the controller supports
+ *					  reservations.
+ * @NVME_CTRL_ONCS_TIMESTAMP:		  If set, then the controller supports
+ *					  the Timestamp feature.
+ * @NVME_CTRL_ONCS_VERIFY:		  If set, then the controller supports
+ *					  the Verify command.
+ * @NVME_CTRL_ONCS_COPY:		  If set, then the controller supports
+ *					  the copy command.
+ * @NVME_CTRL_ONCS_COPY_SINGLE_ATOMICITY: If set, then the write portion of a
```
johnmeneghini, line 1509: Please submit the format change of the above lines in a separate commit.  You are only adding two new arguments as a part of this change and it would be best to eliminate all unneeded white space changes from your MR.

src/nvme/util.c
```diff
@@ -416,7 +420,39 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
 			copy[i].elbt[9 - j] = (eilbrts[i] >> (8 * j)) & 0xff;
 		copy[i].elbt[1] = 0;
 		copy[i].elbt[0] = 0;
-	}
```
johnmeneghini, line 419: Another unneeded white space change?

johnmeneghini (changes requested):
Looks good.  Just a few small tweaks.

src/nvme/util.c
```diff
@@ -419,6 +423,38 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
...
 }
 
+void nvme_init_copy_range_f2(struct nvme_copy_range_f2 *copy, __u32 *snsids,
+			  __u16 *nlbs, __u64 *slbas, __u16 *sopts,
+			  __u32 *eilbrts, __u32 *elbatms, __u32 *elbats,
+			  __u16 nr)
+{
+	int i;
+
+	nvme_init_copy_range((struct nvme_copy_range *)copy, nlbs, slbas,
```
calebsander, line 433: It is undefined behavior to access storage through a pointer type that doesn't match the type the storage was declared with. (This is because the two types may have different alignment requirements, and because C compilers are allowed to assume pointers of different types don't alias.)

There are a few ways to avoid this. One is to just repeat the field assignments in each of the `nvme_init_copy_range_f*()` functions. Another is to define a `struct nvme_copy_range` on the stack, pass a pointer to it into `nvme_init_copy_range()`, and then `memcpy()` it to the `nvme_copy_range_f2`. Or a macro to deduplicate the common code in these functions. And there are plenty of other possible ways to do this...

src/nvme/util.c
```diff
@@ -419,6 +423,38 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
...
+		eilbrts, elbatms, elbats, nr);
+
+	for (i = 0; i < nr; i++) {
+		copy[i].snsid = cpu_to_le32(snsids[i]);
+		copy[i].sopt = cpu_to_le16(sopts[i]);
+	}
```
calebsander, line 439: It would probably be more performant to set the ranges in a single pass instead of looping over them twice

src/nvme/util.h
```diff
@@ -150,6 +150,40 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
...
+ * @slbas:	Starting LBA
+ * @sopts:	Source options
+ * @eilbrts:	Expected initial logical block reference tag
+ * @elbatms:	Expected logical block application tag mask
+ * @elbats:	Expected logical block application tag
+ * @nr:		Number of descriptors to construct
+ */
+void nvme_init_copy_range_f2(struct nvme_copy_range_f2 *copy, __u32 *snsids,
+			     __u16 *nlbs, __u64 *slbas, __u16 *sopts,
+			     __u32 *eilbrts, __u32 *elbatms, __u32 *elbats,
```
calebsander, line 167: All pointers besides `struct nvme_copy_range_f2 *copy` could be marked `const`

src/nvme/types.h
```diff
@@ -4802,6 +4812,58 @@ struct nvme_copy_range_f1 {
...
+ * @elbatm:	Expected Logical Block Application Tag Mask
+ * @elbat:	Expected Logical Block Application Tag
+ */
+struct nvme_copy_range_f3 {
+	__le32			snsid;
+	__u8			rsvd4[4];
+	__le64			slba;
+	__le16			nlb;
+	__u8			rsvd18[4];
+	__le16			sopt;
```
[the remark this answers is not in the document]
jgu-pure: Huh, I see 23:22, with 21:20 reserved. (From the version in here: [NVM Express 2.0 Ratified TPs 10162023](https://nvmexpress.org/wp-content/uploads/NVM-Express-2.0-Ratified-TPs-10162023.zip))

src/nvme/util.h
jgu-pure: Another change? The existing nvme_init_copy_range and nvme_init_copy_range_f1 also have this issue.

calebsander: The existing functions could be modified too; I think that's a backwards-compatible change. But if you'd rather not do that, I'm fine with dropping the comment.

src/nvme/types.h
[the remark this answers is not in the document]
calebsander: I see that for formats 0/2 but format 1/3 says 23:20 for Source Range 0. The other source ranges do suggest only the latter 2 bytes (63:62 for source range 1, 4063:4062 for source range 101). Perhaps we should ask for a fix to the spec.

src/nvme/util.c
```diff
@@ -419,6 +423,45 @@ void nvme_init_copy_range_f1(struct nvme_copy_range_f1 *copy, __u16 *nlbs,
...
+		copy[i].snsid = cpu_to_le32(snsids[i]);
+		copy[i].nlb = cpu_to_le16(nlbs[i]);
+		copy[i].slba = cpu_to_le64(slbas[i]);
+		copy[i].sopt = cpu_to_le16(sopts[i]);
+		copy[i].elbatm = cpu_to_le16(elbatms[i]);
+		copy[i].elbat = cpu_to_le16(elbats[i]);
+		for (j = 0; j < 8; j++)
+			copy[i].elbt[9 - j] = (eilbrts[i] >> (8 * j)) & 0xff;
+		copy[i].elbt[1] = 0;
+		copy[i].elbt[0] = 0;
```
calebsander, line 461: Might be worth extracting this to a helper that can be shared with format 1

src/nvme/types.h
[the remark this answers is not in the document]
jgu-pure: Randy Jennings filed a bug for this here: https://bugzilla.nvmexpress.org/show_bug.cgi?id=228

src/nvme/util.h
jgu-pure: This requires a corresponding change in nvme-cli's `libnvme-wrap.c`. If we expect nvme-cli and libnvme to always be "in sync" I think it's ok to make this change, but I'm not sure if that's the case.

src/nvme/util.c
```diff
@@ -386,6 +390,15 @@ const char *nvme_status_to_string(int status, bool fabrics)
 	return s;
 }
 
+static inline void nvme_init_copy_range_elbt(__u8 *elbt, __u64 eilbrt)
+{
+	int j;
+	for (j = 0; j < 8; j++)
```
calebsander, line 396: Could use `i` now

src/nvme/util.h
calebsander: Yes, they are distributed together, but can be built separately. That's a fair point, so let's not worry about making things `const`.

src/nvme/util.c
```diff
@@ -386,6 +390,15 @@ const char *nvme_status_to_string(int status, bool fabrics)
...
+		elbt[9 - j] = (eilbrt >> (8 * j)) & 0xff;
```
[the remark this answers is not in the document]
jgu-pure: I think I will leave it alone. I think even though in practice the higher order bits are discarded, it's clearer if we show that that's what we intend.

calebsander (approved):

igaw:
I've applied the patch directly after some minor whitespace cleanups. Thanks.

PR #736, opened 2023-10-25
types: add Host Behavior Support field definitions

calebsander:
Bytes 1 and 2 are now defined as fields ETDAS and LBAFEE, so update the `struct nvme_feat_host_behavior` definition.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -4743,13 +4743,17 @@ struct nvme_plm_config {
 /**
  * struct nvme_feat_host_behavior - Host Behavior Support - Data Structure
  * @acre:	Advanced Command Retry Enable
- * @rsvd1:	Reserved
+ * @etdas:	Extended Telemetry Data Area 4 Supported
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -4733,11 +4733,15 @@ struct nvme_plm_config {
 /**
  * struct nvme_feat_host_behavior - Host Behavior Support - Data Structure
  * @acre:	Advanced Command Retry Enable
+ * @etdas:	Extended Telemetry Data Area 4 Supported
+ * @lbafee:	LBA Format Extension Enable
  * @rsvd1:	Reserved
```
jgu-pure, line 4738: rsvd3?

calebsander: Ah good call

jgu-pure (approved):

igaw:
Rebase and resolved merge conflict due to TP4130.

igaw:
Thanks!

Issue #735, opened 2023-10-24
Length field not included in HKDF-Expand-Label computation for retained and TLS PSK


prashanth-nayak:
RFC 8446 defines HKDF-Expand-Label as:

HKDF-Expand-Label(Secret, Label, Context, Length) =
            HKDF-Expand(Secret, HkdfLabel, Length)

Where HkdfLabel is specified as:

struct {
           uint16 length = Length;
           opaque label<7..255> = "tls13 " + Label;
           opaque context<0..255> = Context;
} HkdfLabel;

Currently in linux.c:derive_retained_key() and linux.c:derive_tls_key() the 2 byte key_len is not included in HKDF-Expand-Label operation. https://github.com/linux-nvme/libnvme/blob/da8c28e5e220be4742442114252d136097056928/src/nvme/linux.c#L576

An additional EVP_PKEY_CTX_add1_hkdf_info() call with the key_len  as info   needs to be added.

prashanth-nayak:
Fix submitted.

PR #737, opened 2023-10-30
nvme: Add length field to Hkdf-Expand-Label computation

prashanth-nayak:
Fix to add the 2 byte length field to the HKDF-Expand-Label computation for retained and TLS PSK.
#735

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -556,6 +556,7 @@ static int derive_retained_key(const EVP_MD *md, const char *hostnqn,
 			       size_t key_len)
 {
 	EVP_PKEY_CTX *ctx;
+	uint16_t length = key_len & 0xFFFF;
 	int ret;
 
 	ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_HKDF, NULL);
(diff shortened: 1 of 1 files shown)


hreinecke:
Can you give a pointer to the HKDF specification where this is mandated?
And does it have an implication to the kernel implementation if we change it?

prashanth-nayak:
> Can you give a pointer to the HKDF specification where this is mandated? 
RFC 8446 section 7.1 defines HKDF-Expand-Label function which is used to derive the retained-PSK and TLS-PSK: https://datatracker.ietf.org/doc/html/rfc8446#section-7.1

As per RFC8446:
HKDF-Expand-Label(Secret, Label, Context, Length) =
            HKDF-Expand(Secret, HkdfLabel, Length)

       Where HkdfLabel is specified as:

       struct {
           uint16 length = Length;
           opaque label<7..255> = "tls13 " + Label;
           opaque context<0..255> = Context;
       } HkdfLabel;


>And does it have an implication to the kernel implementation if we change it?
I don't see any implication to the current NVMe/TCP-TLS kernel implementation.

igaw:
Looks reasonable to me. I wonder if we couldn't add some unit test for this.

prashanth-nayak:
> Looks reasonable to me. I wonder if we couldn't add some unit test for this.

Do you want me to add unit tests for this workflow ?
I believe the computation for Retained-PSK and TLS-PSK will change again when support for TP8018 is added. 
Should we wait for that before adding the unit tests?

igaw:
If it is feasible to add a unit test I'd rather have one now than later. When support for TP8018 is added, we just update the unit test.

hreinecke:
Indeed, that change looks good.

PR #740, opened 2023-10-31
CI, Build: Fix GitHub "build" workflow

martin-belanger:
GitHub Workflows are now being run with Python 3.12. The Python package "`distutils`" has been deprecated in Python 3.12. This means that older versions of `meson`, which rely on `distutils`, will fail when run with Python 3.12. This patch fixes the GitHub workflow by updating to the latest version of `meson`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,12 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.x'
+      - name: "install/update dependencies"
+        run: |
+          apt update
+          apt-get install --yes --quiet python3-pip
+          python3 -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


igaw:
Thanks for the fix! Though, I switched over to pre-compiled containers as there were way too many build fails due to some network issue. 

I still host the this project on in my own project space as I am still trying to figure out the details. Eventually we should move it the linux-nvme space though.

https://github.com/igaw/linux-nvme-build-img

I've added the python3-disktutils to the Dockerfile, which should fix this

https://github.com/igaw/linux-nvme-build-img/commit/35d79ed77d25e6311fb5aa6391e896d945cef8f2

igaw:
Found the issue. The container was fine but I had forgot to remove the setup-python action which mapped the github Python toolchain into the build container. Fixed now.

PR #742, opened 2023-11-01
json-schema: add keyring and tls_key details

martin-gpy:
Update the JSON schema with the keyring and tls_key details.


igaw:
Thanks!

PR #743, opened 2023-11-02
build, CI: Add a GitHub Action to run the checkpatch.pl script

martin-belanger:
Adding new workflow to run the checkpatch.pl script from the Linux kernel.


igaw:
Thanks. Let's see how this goes.

igaw:
Ah, thanks for the pointer. I think we should make it a bit shorter. Even though we have wide monitors these days, I have several terminals aligned and this limits the max line length again. And I assume I am not the only one which prefers not to have too long lines either.

PR #744, opened 2023-11-08
Fix a few memory leaks in libnvme

maurizio-lombardi:
Static checkers complain about memory leaks in the NBFT code.


igaw:
Thanks!

PR #745, opened 2023-11-12
build: Add -std=c99 and delete libcurl, libarchive and libpkgconf for…

ikegami-t:
… muon

The changes required as same with the nvme-cli build for muon.


igaw:
I've merged this via my PR.

PR #750, opened 2023-11-23
tree: Fix clearing application strings

tbzatek:
Freeing strings without clearing to NULL may potentially lead to double-free later when freeing the tree structs.


igaw:
Thanks!

PR #751, opened 2023-11-29
Use cleanup functions more

calebsander:


src/nvme/tree.c
```diff
@@ -725,8 +725,8 @@ static int nvme_scan_subsystem(struct nvme_root *r, const char *name,
 			if (strcmp(_s->name, name))
 				continue;
 			if (!__nvme_scan_subsystem(r, _s, f, f_args)) {
-				errno = -EINVAL;
-				goto out_free;
+				errno = EINVAL;
+				return -1;
```
igaw, line 729: The `errno` change could properly be it's own patch as it fixes an existing bug. I suppose the `errno` should have been `ret`. But then it's just very minor thing and I am fine as it is. It's something I stumbled upon while reading through the change.

src/nvme/tree.c
```diff
@@ -1589,13 +1579,15 @@ static void _candidate_free(struct candidate_args *candidate)
...
+#define _cleanup_candidate_ __cleanup__(_candidate_free)
+
 nvme_ctrl_t __nvme_lookup_ctrl(nvme_subsystem_t s, const char *transport,
 			       const char *traddr, const char *host_traddr,
 			       const char *host_iface, const char *trsvcid,
 			       const char *subsysnqn, nvme_ctrl_t p)
 {
 	struct nvme_ctrl *c, *matching_c = NULL;
-	struct candidate_args candidate;
+	_cleanup_candidate_ struct candidate_args candidate;
```
igaw, line 1590: ` = {}` ?

src/nvme/tree.c
```diff
@@ -1620,19 +1610,14 @@ bool nvme_ctrl_config_match(struct nvme_ctrl *c, const char *transport,
 			    const char *subsysnqn, const char *host_traddr,
 			    const char *host_iface)
 {
-	bool match;
 	ctrl_match_t ctrl_match;
-	struct candidate_args candidate;
+	_cleanup_candidate_ struct candidate_args candidate;
```
igaw, line 1614: `= {}` ?

igaw (commented):
Looks good. Just a couple of question.

src/nvme/tree.c
calebsander: Sure, I can break that out into a separate patch. Looks like the code meant to set `ret = -EINVAL` instead of `errno = -EINVAL`.

src/nvme/tree.c
calebsander: There's no early `return`, so it's getting cleaned up at the end of the function just as in the existing code. Both the existing code and the new code are depending on `_candidate_init()` to initialize it, which is perhaps bad practice. So I don't think there's any behavior change, but I'm happy to add an explicit initializer if you'd prefer.

src/nvme/tree.c
calebsander: Same here

src/nvme/tree.c
igaw: Ah okay, so it's the same as in the other places where you left the initializer away. I don't have any preferences. Adding explicit initializer will obviously make the code a bit robuster when someone is factoring stuff around. Let's see how things are going. Adding them is pretty easy :)

igaw:
Nice! Thanks a lot!

Issue #550, opened 2022-12-04
nvme-list produces error log entry each time (Samsung 950 Pro)


dswarbrick:
nvme-cli 2.2.x (and possibly earlier 2.x versions) results in an error log entry being created whenever the `nvme list` command is executed, in a system with a Samsung 950 Pro. Version 1.16 did **not** exhibit this.

Drive firmware is 2B0QBXX7 (latest available), and according to smartctl, the supported NVMe version is < 1.2.

The produced error log entries are always the same:
```
# nvme error-log -e 2 /dev/nvme0
Error Log Entries for device:nvme0 entries:2
.................
 Entry[ 0]   
.................
error_count     : 200
sqid            : 0
cmdid           : 0xa00c
status_field    : 0x2002(INVALID_FIELD: A reserved coded value or an unsupported value in a defined field)
phase_tag       : 0
parm_err_loc    : 0
lba             : 0
nsid            : 0x1
vs              : 0
trtype          : The transport type is not indicated or the error is not transport related.
cs              : 0
trtype_spec_info: 0
.................
 Entry[ 1]   
.................
error_count     : 199
sqid            : 0
cmdid           : 0x200a
status_field    : 0x2002(INVALID_FIELD: A reserved coded value or an unsupported value in a defined field)
phase_tag       : 0
parm_err_loc    : 0
lba             : 0
nsid            : 0x1
vs              : 0
trtype          : The transport type is not indicated or the error is not transport related.
cs              : 0
trtype_spec_info: 0
.................
```
Since this system runs a Prometheus node_exporter textfile collector (https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/blob/master/nvme_metrics.sh), which is executed at regular intervals, the `nvme list` command within that script will continually produce error log entries.

Additionally, the kernel log records errors such as these when the nvme-list command is performed:
```
[185931.468688] nvme0: Admin Cmd(0x6), I/O Error (sct 0x0 / sc 0x2) MORE
```
Kernel is 6.0.10.

Bacto:
FYI, I get the same problem with a SAMSUNG PM983 (MZQLB960HAJR-00007) and firmware EDA5702Q.

igaw:
With libnvme v1.7 and the kernel release 6.8 this is fixed.

PR #754, opened 2023-12-07
tree: read all attributes from  sysfs when available

igaw:
The kernel already exposes parts or all attributes we are looking up
with the ns id command. By reading these from the sysfs we can remove
the last command we are issuing during the topology scan.

Depends on: https://lore.kernel.org/linux-nvme/20231218165954.29652-1-dwagner@suse.de/

Fixes: #681 #550

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -2501,7 +2501,6 @@ static void nvme_ns_set_generic_name(struct nvme_ns *n, const char *name)
 static nvme_ns_t nvme_ns_open(const char *sys_path, const char *name)
 {
 	struct nvme_ns *n;
-	int fd;
 
 	n = calloc(1, sizeof(*n));
 	if (!n) {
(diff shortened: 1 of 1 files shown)


src/nvme/tree.c
```diff
@@ -2322,60 +2322,142 @@ int nvme_ns_flush(nvme_ns_t n)
...
 {
-	void *d = descs;
-	int i, len;
+	char *endptr;
+	__u64 v;
 
-	for (i = 0; i < NVME_IDENTIFY_DATA_SIZE; i += len) {
-		struct nvme_ns_id_desc *desc = d + i;
+	errno = 0;
+	v = strtoul(str, &endptr, 0);
```
martin-belanger, line 2331: should this be `strtoull()` for an unsigned 64-bit (`unsigned long long`)?

src/nvme/tree.c
```diff
@@ -2322,60 +2322,164 @@ int nvme_ns_flush(nvme_ns_t n)
...
+static int nvme_strtouuid(const char *str, void *res)
+{
+	memcpy(res, str, NVME_UUID_LEN);
+	return 0;
+}
+
+struct sysfs_attr_table {
+	void *var;
+	int (*parse)(const char *str, void *res);
+	bool mandetory;
```
tbzatek, line 2400: typo ;o)

PR #755, opened 2023-12-07
Get LBA Status does not specify data_len; returning no data

### LINKED ISSUE #753: Get LBA Status returns no data
The current `nvme_get_lba_status` method never programs the `data_len` field in the passthru command, resulting in no data being returned in the LBAS buffer.

This is an easy enough fix - but there are a couple of solutions and I was hoping one of the maintainers could comment on the preferred one.

1. Add a dedicated `data_len` field to the args, and simply forward that to the passthru field. Allows for the buffer to be sized independently of the mndw field, anything larger than mndw is OK but anything smaller would be an overflow bug. Requires a nvme-cli change.
2. Infer the `data_len` value from the `mndw` value `.data_len = (mndw + 1) << 2`. This is easy and would _not_ require a corresponding nvme-cli change, but if the buffer is any smaller would be an overflow bug.
3. Replace current `mndw` field with `data_len` in the args, and infer mndw from data_len using `__u32 mndw = (args->len >> 2) -1;`. This is the technique used for many other commands: nvme_get_log, nvme_fw_download, nvme_directive_send|recv, and the reservation commands nvme_resv_* . Would require a nvme-cli change but be consistent within the library.

NateThornton:
Current `nvme_get_lba_status()` function does specify any `data_len` in the admin command, resulting in zero data being returned. Of the possible implementations described in #753 , this is the least intrusive change as it requires no update to nvme-cli.

If an alternative approach is preferred, please comment.

Closes #753

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -1568,6 +1568,7 @@ int nvme_get_lba_status(struct nvme_get_lba_status_args *args)
 		.opcode =  nvme_admin_get_lba_status,
 		.nsid = args->nsid,
 		.addr = (__u64)(uintptr_t)args->lbas,
+		.data_len = (args->mndw + 1) << 2,
 		.cdw10 = cdw10,
 		.cdw11 = cdw11,
 		.cdw12 = cdw12,


igaw:
Updated commit message.

igaw:
Could you add your SoB? Thanks!

igaw:
Thanks!

PR #756, opened 2023-12-16
util: Added function to find specific UUID in UUID list.

lgdacunh:


src/nvme/util.c
```diff
@@ -951,6 +951,22 @@ int nvme_uuid_random(unsigned char uuid[NVME_UUID_LEN])
...
+	const unsigned char uuid_end[NVME_UUID_LEN] = {0};
+
+	if ((!uuid_list) || (!uuid))
+		return 0;
+
+	for (int i = 0; i < NVME_ID_UUID_LIST_MAX; i++) {
+		if (memcmp(uuid_end, &uuid_list->entry[i].uuid, NVME_UUID_LEN) == 0)
+			return 0;
+		if (memcmp(uuid, &uuid_list->entry[i].uuid, NVME_UUID_LEN) == 0)
+			return i + 1;
```
igaw, line 965: Why not just use a `int` as return type and then return `-1` in with errno set as we do in all other commands? This allows to return the correct index here.

lgdacunh: @igaw , the initial idea here was to produce output already on the type used by command structures like nvme_get_log_args.uuidx, nvme_get_features_args.uuidx, nvme_set_features_args.uuidx, and nvme_lockdown_args.uuidx.
I guess it is ok to let the compiler automatically cast when assigning the value to the uuidx property, and use the extra bits to signal certain conditions.
But I don't see were we would get an errno here. Maybe we could get a segfault if we receive bad uuid_list/uuid pointers.
I is worth mention that my interpretation of the NVMe spec is  UUID index 0, indicates the absence of UUID.
I still see a point on keeping consistency. If you still would prefer to keep consistency on the library API,  please make me know what would be the ideal function signature in this case and what error code I could return.

igaw: We have to check the return value anyway to decide to use it or not. Thus it doesn't really matter if it is 0 or -1. My point is about consistency as we use this pattern (hopefully) consistently everywhere. In hindsight, it would have been better not to follow the POSIX style here but well that's too late now. In this case if the function return -1, `errno` could be set to `ENOENT`.

About the idea with additional bits encoded in the return value: I'd like to avoid overloading values. This is something we should only do if speed and size really matters in an API. I'd rather choose simplicity instead (and use the compiler to catch problems). There are some ideas how to handle errors for 2.x but for 1.x the ship has sailed :(

lgdacunh: @igaw, Done!

igaw:
Thanks!

PR #760, opened 2023-12-19
build: construct path to config.h manually

igaw:
meson reports:

  internal/meson.build:25: DEPRECATION: Project uses feature that was
  always broken, and is now deprecated since '1.3.0': str.format: Value
  other than strings, integers, bools, options, dictionaries and lists
  thereof.

Thus just hardcode the config file path


igaw:
The absolute path is necessary to build libnvme as subproject in nvme-cli

PR #764, opened 2023-12-27
ioctl: Split latency outputs from debugging feature to show command

ikegami-t:


src/nvme/ioctl.c
```diff
@@ -112,8 +112,136 @@ static void nvme_show_command(struct nvme_passthru_cmd *cmd, int err, struct tim
 	printf("timeout_ms   : %08x\n", cmd->timeout_ms);
 	printf("result       : %08x\n", cmd->result);
 	printf("err          : %d\n", err);
-	printf("latency      : %lu us\n",
-	       (end.tv_sec - start.tv_sec) * 1000000 + (end.tv_usec - start.tv_usec));
+}
+
+const char *nvme_admin_to_string(__u8 opcode)
```
calebsander, line 117: Would these functions be useful to export from `libnvme`?

src/nvme/ioctl.c
```diff
@@ -112,8 +112,136 @@ static void nvme_show_command(struct nvme_passthru_cmd *cmd, int err, struct tim
...
+		return "Security Receive";
+	case nvme_admin_sanitize_nvm:
+		return "Sanitize";
+	case nvme_admin_get_lba_status:
+		return "Get LBA Status";
+	default:
+		break;
+	}
+
+	return NULL;
```
calebsander, line 178: Just `return NULL` instead of `break`?

src/nvme/ioctl.c
```diff
@@ -126,23 +254,36 @@ bool nvme_get_debug(void)
...
 
-	if (nvme_get_debug()) {
+	if (nvme_get_latency()) {
 		gettimeofday(&end, NULL);
-		nvme_show_command(cmd, err, start, end);
+		nvme_show_latency(ioctl_cmd == NVME_IOCTL_ADMIN_CMD, cmd->opcode, &start, &end);
 	}
 
+	if (nvme_get_debug())
+		nvme_show_command(cmd, err);
```
calebsander, line 285: Pre-existing, but curious why the 64-bit ioctls (`nvme_submit_passthru64()`) don't do this logging. Is it not useful for them? Does it just require more work to implement than it's worth?

src/nvme/ioctl.c
ikegami-t: Thanks for your comments. Yes the code basically copied from `nvme-cli: nvme_cmd_to_string()` and duplicated so I will do merge the functionality to export from `libnmve`.

src/nvme/ioctl.c
ikegami-t: In my understandig for the coding manner better to use the default case. By the way how about the following implementation instead? Or just I thought also to define the opcode strings as a table to use instead of the function using the switch case statements.
```c
	char *ret = NULL;

	switch (opcode) {
	case nvme_admin_delete_sq:
		ret = "Delete I/O Submission Queue";
		break;
...
	default:
		break;
	}

	return ret;
```

src/nvme/ioctl.c
ikegami-t: Yes will do implement as same for the 64-bit ioctls also.

src/nvme/ioctl.c
calebsander: I think it is terser and easier to follow to just return from the switch case, but doesn't really matter. A simple lookup table could definitely be faster. It might take up more memory unless you only stored entries up to the largest defined opcode.

ikegami-t: I see so will do try to use the table later.

src/nvme/ioctl.c
```diff
@@ -126,22 +245,25 @@ bool nvme_get_debug(void)
...
-
-	err = ioctl(fd, ioctl_cmd, cmd);
+	err = nvme_ioctl(fd, ioctl_cmd, cmd);
 
-	if (nvme_get_debug()) {
-		gettimeofday(&end, NULL);
-		nvme_show_command(cmd, err, start, end);
-	}
+	if (nvme_get_debug())
+		nvme_show_command(cmd, err);
```
calebsander, line 266: This could probably also be done in the 64-bit case?

calebsander (approved):

src/nvme/ioctl.c
```diff
@@ -42,14 +44,134 @@ static int nvme_verify_chr(int fd)
 	return 0;
 }
 
+static const char *nvme_cmd_to_string(unsigned long req, __u8 opcode)
+{
+	const char *cmd_name;
+
+	if (req == NVME_IOCTL_ADMIN_CMD || req == NVME_IOCTL_ADMIN64_CMD)
```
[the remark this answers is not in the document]
ikegami-t: Will do.

src/nvme/ioctl.c
```diff
@@ -42,14 +44,134 @@ static int nvme_verify_chr(int fd)
...
+	case NVME_IOCTL_ADMIN_CMD:
+		fallthrough;
+	case NVME_IOCTL_ADMIN64_CMD:
+		ioctl_cmd_admin = true;
+		break;
+	default:
+		break;
+	}
+
+	return ioctl_cmd_admin;
```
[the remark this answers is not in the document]
ikegami-t: Okay right.

src/nvme/ioctl.c
```diff
@@ -42,14 +44,116 @@ static int nvme_verify_chr(int fd)
...
+		printf("ioctl: %lx (%s) ", req, nvme_ioctl_to_string(req));
+
+	printf("latency: %lu us\n", latency.tv_sec * 1000000 + latency.tv_usec);
+}
+
+static bool nvme_ioctl_cmd(unsigned long req)
+{
+	bool ioctl_cmd = false;
+
+	switch (req) {
```
martin-belanger, line 100: Do you really need to use `fallthrough`? Why not just:
```
	switch (req) {
	case NVME_IOCTL_ADMIN_CMD:
	case NVME_IOCTL_IO_CMD:
	case NVME_IOCTL_ADMIN64_CMD:
	case NVME_IOCTL_IO64_CMD:
		ioctl_cmd = true;
		break;
```

src/nvme/ioctl.c
```diff
@@ -42,14 +44,116 @@ static int nvme_verify_chr(int fd)
...
+{
+	if (req == NVME_IOCTL_ADMIN_CMD || req == NVME_IOCTL_ADMIN64_CMD)
+		return true;
+
+	return false;
+}
+
+static bool nvme_ioctl_cmd_64(unsigned long req)
+{
+	if (req == NVME_IOCTL_ADMIN64_CMD || req == NVME_IOCTL_IO64_CMD)
```
martin-belanger, line 57: Could be simplified to a single line:
```
return req == NVME_IOCTL_ADMIN64_CMD || req == NVME_IOCTL_IO64_CMD;
```

src/nvme/util.c
```diff
@@ -1116,3 +1116,104 @@ void *__nvme_alloc(size_t len)
...
+
+const char *nvme_nvm_to_string(__u8 opcode)
+{
+	if (opcode < ARRAY_SIZE(nvm_opcode))
+		return nvm_opcode[opcode];
+
+	return NULL;
+}
+
+const char *nvme_ioctl_to_string(unsigned long req)
```
martin-belanger, line 1185: Could be simplified as follows:
```
const char *nvme_ioctl_to_string(unsigned long req)
{
	switch (req) {
	case NVME_IOCTL_ID:
		return "ID";
	case NVME_IOCTL_RESET:
		return "Reset";
	case NVME_IOCTL_SUBSYS_RESET:
		return "Subsystem reset";
	case NVME_IOCTL_RESCAN:
		return "Rescan";
	case NVME_IOCTL_ADMIN_CMD:
		return "Admin command";
	case NVME_IOCTL_IO_CMD:
		return "IO command";
	case NVME_IOCTL_ADMIN64_CMD:
		return "Admin64 command";
	case NVME_IOCTL_IO64_CMD:
		return "IO64 command";
	default:
		break;
	}

	return NULL;
}

```

src/nvme/ioctl.c
ikegami-t: The checkpatch.pl tool checks `fallthrough` and the tool is used by the PR check I think. Also usually used it in Kernel code I think also.
  <https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/checkpatch.pl#n7302>

src/nvme/ioctl.c
ikegami-t: Yes looks good. Thank you.

igaw:
I don't like the whole approach. This is adding debugging code to the library. The commit 17429f7578c76409af483cd14926618e88458de5 didn't even consider all the ioctl interfaces (64bit version) and prints just to stdout. I think we should remove it.

Why not just use existing syscall tracers instead?

Simple syscall latencies can be traced with `strace` and you can go fancy with `bpftrace`. E.g something like:

```bpf
#!/usr/bin/bpftrace
#include <sys/ioctl.h>

struct nvme_passthru_cmd {
	__u8	opcode;
	__u8	flags;
	__u16	rsvd1;
	__u32	nsid;
	__u32	cdw2;
	__u32	cdw3;
	__u64	metadata;
	__u64	addr;
	__u32	metadata_len;
	__u32	data_len;
	__u32	cdw10;
	__u32	cdw11;
	__u32	cdw12;
	__u32	cdw13;
	__u32	cdw14;
	__u32	cdw15;
	__u32	timeout_ms;
	__u32	result;
};

/* remember our own pid, so that we can filter it out from the handlers */
BEGIN { }

tracepoint:syscalls:sys_enter_ioctl /comm == "nvme" && args->__syscall_nr == 16/ {
       $cmd = (struct nvme_passthru_cmd *)args->arg;

       printf("opcode %d\n",$cmd->opcode);
       printf("cdw10 %x\n", $cmd->cdw10)
}

END { }
```

igaw:
With the recent refactoring in logging/debugging the annotation and measuring is done inside application. No need to clutter the library with debugging code.

PR #769, opened 2024-01-18
meson.build: fixup 'join' syntax

hreinecke:
The 'join' function only takes one argument, so use the correct syntax here.

Fixes: 0cbda86c ("build: handle patch level versioning")


hreinecke:
Something's messed up with the 'fixes' tag. Retrying with a new PR.

igaw:
I fixed the 'Fixed' tag bug. Works now.

PR #771, opened 2024-01-18
fabrics: add 'concat' option

hreinecke:
Add an option 'concat' to enable secure concatenation for TCP.

### DIFF
--- a/doc/config-schema.json.in
+++ b/doc/config-schema.json.in
@@ -169,6 +169,11 @@
 		    "type": "boolean",
 		    "default": false
 		},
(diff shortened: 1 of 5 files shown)


doc/rst/fabrics.rst
```diff
@@ -92,6 +93,9 @@ Fabrics-specific definitions.
 ``tls``
   Start TLS on the connection (TCP)
 
+``concat``
```
igaw, line 96: The `doc/rst/fabtrics.rst` file gets auto generated. No need to add it to the change.

PR #773, opened 2024-01-19
mi: set correct rc and errno when crc mismatch

jinliangw:
Before the fix, when we meet crc mismatch response, the errno is 0 and rc is 1. This combination will be mistaken as Admin Generic Command Status code value 0x1 (Invalid Command Opcode):
  $ nvme id-ctrl mctp:1,20
  crc mismatch
  NVMe status: Invalid Command Opcode: A reserved coded value or an unsupported value in the command opcode field(0x1

After the fix, the rc is -1, and errno is set to Bad message.
  $ nvme id-ctrl mctp:1,20
  crc mismatch
  identify controller: Bad message


jk-ozlabs (approved):
Looks good to me, thanks!

jinliangw:
Thank @igaw and @jk-ozlabs for the review, the checkpatch failure is fixed now.

igaw:
Thanks!

PR #776, opened 2024-02-03
types: Add controller properties CMBEBS, CMBSWTP and NSSD

ikegami-t:
Described by the NVM express base specification revision 2.0c.


igaw:
The documentation needs to be updated as well.

igaw:
Thanks!

PR #779, opened 2024-02-07
example: fix mi identify failed with error cntid

zhangjian3032:
This command failed when we try to identify a controller that the controller id is 1.

Refer to the `Figure 273: Identify - CNS Values`:
+-----------+-------+
| CNS Value | CNTID |
+-----------+-------+
| 01h       | N     |
+-----------+-------+
When CNS is 01h, the CNTID field is ignored.

See `Figure 270: Identify - Command Dword 10`:
If this field is not used as part of the Identify operation, then
* host software shall clear this field to 0h for backwards compatibility (0h is a valid controller identifier);
* and the controller shall ignore this field.

This filed is set to controller id in the example code, but it should be 0 when CNS is 1.

PS: The NVMe that we are testing does not ignore the CNTID field and returns an error when the CNTID field is not 0.

### DIFF
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -184,24 +184,24 @@ int do_identify(nvme_mi_ep_t ep, int argc, char **argv)
 	uint16_t ctrl_id;
 	char buf[41];
 	bool partial;
-	int rc, tmp;
+	int rc;
 
 	if (argc < 2) {
 		fprintf(stderr, "no controller ID specified\n");
 		return -1;
 	}
 
-	tmp = atoi(argv[1]);
-	if (tmp < 0 || tmp > 0xffff) {
+	ctrl_id = atoi(argv[1]);
+	if (ctrl_id < 0 || ctrl_id > 0xffff) {
 		fprintf(stderr, "invalid controller ID\n");
 		return -1;
 	}
 
-	ctrl_id = tmp & 0xffff;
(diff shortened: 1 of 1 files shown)


examples/mi-mctp.c
```diff
@@ -184,24 +184,24 @@ int do_identify(nvme_mi_ep_t ep, int argc, char **argv)
...
 
 	if (argc < 2) {
 		fprintf(stderr, "no controller ID specified\n");
 		return -1;
 	}
 
-	tmp = atoi(argv[1]);
-	if (tmp < 0 || tmp > 0xffff) {
+	ctrl_id = atoi(argv[1]);
+	if (ctrl_id < 0 || ctrl_id > 0xffff) {
```
jk-ozlabs, line 195: These checks are no longer possible if you're using `ctrl_id` directly there, as it's a `uint16_t`. Values outside of that 16-bit unsigned range will get silently dropped, instead of the check failing.

jk-ozlabs (changes requested):
The core of this change looks good (setting CNTID to 0), but you have unrelated changes that break input validation on the controller id passed from the command-line.

If you can remove those unrelated changes, then we should be fine.

examples/mi-mctp.c
zhangjian3032: yup, done

jk-ozlabs (approved):
Great, LGTM! Thanks for the fix.

PR #781, opened 2024-02-09
NBFT parser SSNS flags

tbzatek:
There are significant differences in different vendor implementations of the NVMe Boot Specification and as it turned out not everything is supported by the libnvme NBFT parser.


tbzatek:
This is pretty much work in progress and we might need some more flags. nvme-cli patches will follow.
I'm keeping future expansion with TP8029 in mind, though it's not a subject of this pull request.

The most important question: is the addition of new struct members at the end of structs okay with you?

igaw:
Hmm, as long we can figure out that we don't break existing users, you can do whatever is needed. That means older binary version of nvme-cli should still work against newer libnvme versions. If I get this right, this should work if you extend at the end of the struct. 

The other dependency order should be handled in nvme-cli. Which I think we do with the version check.

tbzatek:
Thanks, I'm not aware of any user other than nvme-cli. Potentially network management daemons (NetworkManager, wicked) may need to parse some extra info, but I'm not aware of such thing happening at the moment.

We do plan to further extend these convenient `nbft_info_*` structs with TP-8029 and TP-8027 (along with the usual 1:1 spec translation to C headers). Also, NBFT ACPI table is versioned and the parser is already covered by unit tests.

igaw:
Yes this code is pretty fresh so I am not really concerned about breaking stuff. Maybe we should mark certain features as experimental until we are happy with it. Just as idea.

tbzatek:
Okay, this is ready for review. We're far from done but this is an independent chunk of changes at least.

Cc: @mwilck

mwilck (approved):
LGTM. I didn't bother to review the "diffs" in detail.

johnmeneghini (approved):
Clean up the no space before tabs problem and commit - please.

PR #782, opened 2024-02-14
log: Introduce nvme_get_logging_level()

tbzatek:
This is essentially a getter for `nvme_init_logging()` since `nvme_root_t` is an opaque struct. Takes optional pointer to bool args to retrieve PID and timestamp logging values.

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,4 +1,9 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
+LIBNVME_1_9 {
+	global:
+		nvme_get_logging_level;
+};
+
 LIBNVME_1_8 {
 	global:
 		nvme_uuid_find;
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -58,6 +58,8 @@ __nvme_msg(nvme_root_t r, int lvl,
 
 	if (r && lvl > r->log_level)
 		return;
+	if (!r && lvl > DEFAULT_LOGLEVEL)
+		return;
 
 	if (r && r->log_timestamp) {
 		struct timespec now;
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -99,6 +99,19 @@ void nvme_init_logging(nvme_root_t r, int lvl, bool log_pid, bool log_tstamp)
 	r->log_timestamp = log_tstamp;
 }
 
+int nvme_get_logging_level(nvme_root_t r, bool *log_pid, bool *log_tstamp)
+{
+	if (!r)
+		r = root;
+	if (!r)
+		return DEFAULT_LOGLEVEL;
+	if (log_pid)
+		*log_pid = r->log_pid;
+	if (log_tstamp)
+		*log_tstamp = r->log_timestamp;
+	return r->log_level;
+}
+
 void nvme_set_root(nvme_root_t r)
 {
 	root = r;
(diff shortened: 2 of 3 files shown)


tbzatek:
The idea behind this is to temporarily pause logging (save-override-restore) for operations that are expected to fail.

tbzatek:
Added one more change to fix the scenario where e.g. nvme-cli plugins have not initialized logging (since there's no `root` available) and debugging clutter is being printed on stderr.

```
# nvme nbft show -o json
file /sys/firmware/acpi/tables/NBFT: SSNS 1 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 2 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 3 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 4 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 5 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 6 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 7 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 8 skipping duplicate HFI index 1
file /sys/firmware/acpi/tables/NBFT: SSNS 9 skipping duplicate HFI index 2
file /sys/firmware/acpi/tables/NBFT: SSNS 10 skipping duplicate HFI index 2
file /sys/firmware/acpi/tables/NBFT: SSNS 11 skipping duplicate HFI index 2
file /sys/firmware/acpi/tables/NBFT: SSNS 12 skipping duplicate HFI index 2
file /sys/firmware/acpi/tables/NBFT: SSNS 13 skipping duplicate HFI index 2
file /sys/firmware/acpi/tables/NBFT: SSNS 14 skipping duplicate HFI index 2
... <JSON follows>
```

igaw:
TBH, I think we creating a big mess with the logging. It's getting messier with every commit. Why should the default log level be checked when the root object exists? And what about disabling it on global level, what happens with any existing root object? 

These are just small 'smells' which tell me that this is going wrong. It's not this particular change. I don't have a good idea what to do. Maybe all function which don't take a root object not print anything at all.

tbzatek:
Both of my commits are rather nice-to-have hacks trying to sneak in an existing infrastructure. On a second thought, you're right, there's more weirdness going on.

For start, the symbol naming is inconsistent: `nvme_init_logging()`, `nvme_set_root()` and my `nvme_get_logging_level()`. Would be nice to have a common namespace. Then you have `nvme_set_debug()` and `nvme_get_debug()` for ioctls that are somewhat related as well.

Another issue is that while you may call `nvme_msg(root=NULL)`, you can't init logging without a root object. This is e.g. the case for stateless nvme-cli plugins.

> Maybe all function which don't take a root object not print anything at all.

We would still like to have e.g. the nbft parser high-severity messages printed. The global static root (`nvme_set_root()`) doesn't really help this use case either.

igaw:
I'll have a stab if we can improve this. Anyway, your changes itself look okay but as we discussed there is a bit of a mess around the logging infra. I guess we can't avoid having a global variable for the logging, because there is too much code which wants to print debug/error stuff.

igaw:
The situation in libnvme wasn't that bad I had in memory. Just the recently debug code was not good. I've dropped this and picked up these changes in #788.

tbzatek:
Nice, thanks!

PR #784, opened 2024-02-25
types: Add persistent event log reporting context definitions

ikegami-t:
The changes for persistent event log rci print function.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3340,6 +3340,41 @@ struct nvme_persistent_event_log {
 } __attribute__((packed));
 
 /**
+ * enum nvme_pel_rci - This field indicates the persistent event log reporting context
+ * @NVME_PEL_RCI_RCPID_SHIFT:	Shift amount to get the reporting context port identifier
+ *				from the &struct nvme_persistent_event_log.rci field.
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -3339,6 +3339,41 @@ struct nvme_persistent_event_log {
...
+ */
+enum nvme_pel_rci {
+	NVME_PEL_RCI_RCPID_SHIFT	= 0,
+	NVME_PEL_RCI_RCPIT_SHIFT	= 16,
+	NVME_PEL_RCI_RCE_SHIFT		= 18,
+	NVME_PEL_RCI_RSVD_SHIFT		= 19,
+	NVME_PEL_RCI_RCPID_MASK		= 0xffff,
+	NVME_PEL_RCI_RCPIT_MASK		= 0x3,
+	NVME_PEL_RCI_RCE_MASK		= 0x1,
+	NVME_PEL_RCI_RSVD_MASK		= 0x1fff,
```
igaw, line 3369: Let's not add getters for reserved fields. These bits should not be used anyway.

igaw: I just realized that nvme-cli prints the reserved bits... Seems wrong.

PR #785, opened 2024-02-26
types: Add PEL reporting context port identifier type definitions

ikegami-t:
The changes for PEL rci port identifier type print function.


igaw:
Thanks!

PR #786, opened 2024-02-28
Persistent TLS keys take #2

hreinecke:
Hey all,

this is my next attempt to implement persistent TLS key handling. The original idea of storing the key serial number in the JSON configuration file doesn't really work out as the key serial number is ephemeral, and might change whenever a key is updated/revoked or somesuch. So after a reboot the key won't be present, the we won't be able to load the keys at all.
With the pull request we now store the TLS keys in 'PSK interchange format' as mandated in the NVMe TCP transport specification, which then contains the actual key data. And, obviously, the parser has been updated to read the TLS keys
from the PSK interchange format, too.
The original pull request https://github.com/linux-nvme/libnvme/pull/783 would export all keys via the configuration file, thereby losing the distinction between keys selected from the user (ie via the '--tls_key' option) and keys auto-selected during TLS handshake (when only '--tls' was specified. To fix this this PR now also adds the functions nvme_scan_tls_keys() such that nvme-cli can add an 'export' and 'import' functionality for TLS keys.

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 LIBNVME_1.9 {
 	global:
+		nvme_export_tls_key;
 		nvme_get_logging_level;
 		nvme_read_key;
 		nvme_submit_passthru;
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -2,6 +2,7 @@
 LIBNVME_1.9 {
 	global:
 		nvme_get_logging_level;
+		nvme_read_key;
 		nvme_submit_passthru;
 		nvme_submit_passthru64;
 };
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -5,6 +5,7 @@ LIBNVME_1.9 {
 		nvme_read_key;
 		nvme_submit_passthru;
 		nvme_submit_passthru64;
+		nvme_update_key;
 };
 
 LIBNVME_1_8 {
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -3,6 +3,7 @@ LIBNVME_1.9 {
 	global:
 		nvme_export_tls_key;
 		nvme_get_logging_level;
+		nvme_import_tls_key;
 		nvme_read_key;
 		nvme_submit_passthru;
 		nvme_submit_passthru64;
--- a/src/nvme/crc32.c
+++ b/src/nvme/crc32.c
@@ -1,0 +1,111 @@
+/*
+ *  COPYRIGHT (C) 1986 Gary S. Brown.  You may use this program, or
+ *  code or tables extracted from it, as desired without restriction.
+ */
+
+/*
+ *  First, the polynomial itself and its table of feedback terms.  The
+ *  polynomial is
+ *  X^32+X^26+X^23+X^22+X^16+X^12+X^11+X^10+X^8+X^7+X^5+X^4+X^2+X^1+X^0
+ *
+ *  Note that we take it "backwards" and put the highest-order term in
+ *  the lowest-order bit.  The X^32 term is "implied"; the LSB is the
+ *  X^31 term, etc.  The X^0 term (usually shown as "+1") results in
+ *  the MSB being 1
+ *
+ *  Note that the usual hardware shift register implementation, which
+ *  is what we're using (we're merely optimizing it by doing eight-bit
+ *  chunks at a time) shifts bits into the lowest-order term.  In our
+ *  implementation, that means shifting towards the right.  Why do we
+ *  do it this way?  Because the calculated CRC must be transmitted in
+ *  order from highest-order term to lowest-order term.  UARTs transmit
+ *  characters in order from LSB to MSB.  By storing the CRC this way
+ *  we hand it to the UART in the order low-byte to high-byte; the UART
+ *  sends each low-bit to hight-bit; and the result is transmission bit
+ *  by bit from highest- to lowest-order term without requiring any bit
+ *  shuffling on our part.  Reception works similarly
+ *
+ *  The feedback terms table consists of 256, 32-bit entries.  Notes
+ *
+ *      The table can be generated at runtime if desired; code to do so
+ *      is shown later.  It might not be obvious, but the feedback
+ *      terms simply represent the results of eight shift/xor opera
+ *      tions for all combinations of data and CRC register values
+ *
+ *      The values must be right-shifted by eight bits by the "updcrc
+ *      logic; the shift must be unsigned (bring in zeroes).  On some
+ *      hardware you could probably optimize the shift in assembler by
+ *      using byte-swap instructions
+ *      polynomial $edb88320
+ *
+ *
+ * CRC32 code derived from work by Gary S. Brown.
+ */
+
+/* https://web.mit.edu/freebsd/head/sys/libkern/crc32.c */
+
+#include "crc32.h"
+
+const uint32_t crc32_tab[] = {
+	0x00000000, 0x77073096, 0xee0e612c, 0x990951ba, 0x076dc419, 0x706af48f,
+	0xe963a535, 0x9e6495a3,	0x0edb8832, 0x79dcb8a4, 0xe0d5e91e, 0x97d2d988,
+	0x09b64c2b, 0x7eb17cbd, 0xe7b82d07, 0x90bf1d91, 0x1db71064, 0x6ab020f2,
+	0xf3b97148, 0x84be41de,	0x1adad47d, 0x6ddde4eb, 0xf4d4b551, 0x83d385c7,
+	0x136c9856, 0x646ba8c0, 0xfd62f97a, 0x8a65c9ec,	0x14015c4f, 0x63066cd9,
+	0xfa0f3d63, 0x8d080df5,	0x3b6e20c8, 0x4c69105e, 0xd56041e4, 0xa2677172,
+	0x3c03e4d1, 0x4b04d447, 0xd20d85fd, 0xa50ab56b,	0x35b5a8fa, 0x42b2986c,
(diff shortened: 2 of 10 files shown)


src/nvme/crc32.c
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: FSFAP-no-warranty-disclaimer
```
igaw, line 1: My bad, but I think we can't add the SPDX tag because the license from Gary S. Brown doesn't show up in SPDX database. And the FSFAP is not the right one.

I tried to figure out how this handled by other project and found many copies of the crc32 code below. Some have added a GPL header in front (don't know how this should work but it's not our problem) and other project just use the crc32.c code with and without shortening the original code an don't mention any additional copyright statements.  

The Fedora project has a note on this https://fedoraproject.org/wiki/Licensing/CRC32

Given this, I'd say we don't add any SPDX header until the SPDX project has covered Brown's license.

I think we have done our homework on the license topic.

src/libnvme.map
```diff
@@ -3,6 +3,7 @@ LIBNVME_1_9 {
 	global:
 		nvme_read_key;
 		nvme_update_key;
+		nvme_export_tls_key;
```
igaw, line 6: Insert new values alphabetically sorted.

hreinecke: Okay.

src/nvme/crc32.c
igaw: I got once burned with not paying enough attention to license issues. So I am rather playing safe here. Sorry about that.

igaw:
Expect these couple nitpicks it looks good. I've to read up on our discussion about the ideas from @mwilck about the prefixing `tls_key`

hreinecke:
I'd rather not go that way. Prefixing it will be just adding complexity for no little gain. Especially as we don't have any real exploiter for any of these prefixes. My idea here is to rather delegate it to the kernel keyring; I've got a patchset converting the authentication stuff over to use 'struct key' and store all keys in the kernel keyring. Then one can use either a keys directly (ie the currrent interface) or just reference the key serial from the keystore. That way we can decouple the provisioning step from the nvme-cli interface, and we wouldn't need the prefix idea.

mwilck:
@hreinecke,

> My idea here is to rather delegate it to the kernel keyring; I've got a patchset converting the authentication stuff over to use 'struct key' and store all keys in the kernel keyring. Then one can use either a keys directly (ie the currrent interface) or just reference the key serial from the keystore. That way we can decouple the provisioning step from the nvme-cli interface, and we wouldn't need the prefix idea.

So you want a separate tool that loads the key into the keyring. This is fine. Still, this separate tool needs to be configured how to obtain the key material it feeds into the keyring. It will need to look up keys specific to host NQN, subsystem, NQN and possibly other parameters (like chosen HMAC function), and presumably it will be able to handle various types of key sources roughly along the lines I [listed earlier](https://github.com/linux-nvme/libnvme/pull/783#issuecomment-1961623808), whatever final syntax is chosen.

So the new tool will need to look up configuration items by host and subsystem NQN. 
IMO it'd make a lot of sense to have this tool use `config.json` for this purpose.
`config.json` has exactly the structure which is needed to look up NVMe connection-specific parameters, including library functions to do this comfortably in C and in Python. If we design the new tool with a separate configuration file, that file would necessarily duplicate the structure and a considerable part of the content that `config.json` already has. Using `config.json` would also be a usability benefit for end users, as they would find all configuration items related to a given host/subsys association in a single place.

*If* we decide to let the key provisioning tool use `config.json`, it would make sense to use either prefixing, or some other syntax that make it feasible to derive the method by which the key material will be obtained.

About "just reference the key serial from the keystore", I am not sure how that's supposed to work. You said in #783 that the key serial is "ephemeral", thus it's not suitable for being stored in a static configuration file. Users can figure it out by examining the key ring, but that is cumbersome and doesn't scale. I believe that key lookup in the kernel should work by host NQN/subsys NQN association exclusively. That should be possible already, as the key description contains all information about the NQNs.

hreinecke:
> @hreinecke,
> 
> > My idea here is to rather delegate it to the kernel keyring; I've got a patchset converting the authentication stuff over to use 'struct key' and store all keys in the kernel keyring. Then one can use either a keys directly (ie the currrent interface) or just reference the key serial from the keystore. That way we can decouple the provisioning step from the nvme-cli interface, and we wouldn't need the prefix idea.
> 
> So you want a separate tool that loads the key into the keyring. This is fine. Still, this separate tool needs to be configured how to obtain the key material it feeds into the keyring. It will need to look up keys specific to host NQN, subsystem, NQN and possibly other parameters (like chosen HMAC function), and presumably it will be able to handle various types of key sources roughly along the lines I [listed earlier](https://github.com/linux-nvme/libnvme/pull/783#issuecomment-1961623808), whatever final syntax is chosen.
> 
Yes. But we don't need to decide that now, and can discuss it once that tool is implemented.

> So the new tool will need to look up configuration items by host and subsystem NQN. IMO it'd make a lot of sense to have this tool use `config.json` for this purpose. `config.json` has exactly the structure which is needed to look up NVMe connection-specific parameters, including library functions to do this comfortably in C and in Python. If we design the new tool with a separate configuration file, that file would necessarily duplicate the structure and a considerable part of the content that `config.json` already has. Using `config.json` would also be a usability benefit for end users, as they would find all configuration items related to a given host/subsys association in a single place.
> 
But we will lose the ability to do a key rotation. IE the problem is that the command-line
nvme connect -t tcp -a <ip> -s <nqn> --tls
(ie select keys automatically from the keystore) will be converted into
nvme connect -t tcp -a <ip> -s <nqn> --tls_key <key_id>
when using the config.json file.
With the former we will be do an automatic key rotation. With the latter we can't.
Or, rather, will have to update the json file whenever a new key is selected.

> _If_ we decide to let the key provisioning tool use `config.json`, it would make sense to use either prefixing, or some other syntax that make it feasible to derive the method by which the key material will be obtained.
> 
... or just copy out all keys from the keystore upon shutdown, and read them in when starting up.
Which is what I tried to do here.

> About "just reference the key serial from the keystore", I am not sure how that's supposed to work. You said in #783 that the key serial is "ephemeral", thus it's not suitable for being stored in a static configuration file. Users can figure it out by examining the key ring, but that is cumbersome and doesn't scale. I believe that key lookup in the kernel should work by host NQN/subsys NQN association exclusively. That should be possible already, as the key description contains all information about the NQNs.

Which wouldn't it scale to look up from the keyring? Sure, lookup will take longer the more keys are stored, but they will be revoked upon key rotation so we shouldn't have any stale keys in the keystore.

mwilck:
Ok. So we'll have a separate tool, and this tool will probably have a separate configuration file.

I don't want to delay the inclusion of this PR with my comments. I just want to make sure we will be able to support multiple configurable ways to obtain key material in the future, without having to change the design.

src/nvme/linux.c
```diff
@@ -1295,3 +1429,106 @@ long nvme_insert_tls_key(const char *keyring, const char *key_type,
...
+	encoded_key = malloc(encoded_len);
+	if (!encoded_key) {
+		errno = ENOMEM;
+		return NULL;
+	}
+	memset(encoded_key, 0, encoded_len);
+	len = sprintf(encoded_key, "NVMeTLSkey-1:%02x:",
+		      key_len == 32 ? 1 : 2);
+	base64_encode(raw_secret, raw_len, encoded_key + len);
+	strcat(encoded_key, ":");
```
mwilck, line 1466: `base64_encode()` returns the number of bytes written, so you could avoid using the dreaded `strcat()` here.

mwilck (approved):
The code looks good to me. Just a few nits.

mwilck:
> Which wouldn't it scale to look up from the keyring?

Sorry for bad choice of words. Perhaps I am still missing something from the big picture. I was wondering how the *user* would figure out the serial number to feed into the connect command. I figured the user could examine the keyring e.g. with **keyctl** to derive the serial number. Doing this manually would be slow, that's what I meant with "doesn't scale".

mwilck:
> Well, he doesn't have to (which is kinda the point of my argumentation all along.) When you just specify '--tls' the kernel will lookup the first matching key from the keystore, irrespective of the key serial number.

Thanks. I got that part. I just misunderstood the sentence about "referencing the key serial from the keystore".

src/nvme/crc32.c
hreinecke: Ok, removed the SPDX line here.

src/nvme/json.c
```diff
@@ -25,10 +25,59 @@
...
+	const char *subsysnqn = nvme_ctrl_get_subsysnqn(c);
+	int key_len;
+	unsigned int hmac;
+	long key_id;
+	_cleanup_free_ unsigned char *key_data;
+
+	key_data = nvme_import_tls_key(key_str, &key_len, &hmac);
+	if (!key_data)
+		return;
+	if (!hostnqn || !subsysnqn) {
```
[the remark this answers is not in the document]
hreinecke: Fixed.

src/nvme/json.c
```diff
@@ -75,21 +124,24 @@ static void json_update_attributes(nvme_ctrl_t c,
...
+		if (!strcmp("tls_key", key_str) && cfg->tls_key == 0)
+			key_data = json_object_get_string(val_obj);
 	}
+
+	/*
+	 * We might need the keyring information from the above loop,
+	 * so we can only import the TLS key once all entries are
+	 * processed.
+	 */
+	if (key_data)
```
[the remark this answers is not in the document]
hreinecke: Renamed to 'encoded_key'.

src/nvme/linux.c
hreinecke: Replaced with explicit setting of characters.

igaw:
Just rebased to the current head to resolve the conflicts.

igaw:
Thanks!

PR #787, opened 2024-03-04
fabrics: check the read() return value

maurizio-lombardi:
Do not pass to the is_dmi_uuid_valid() function an invalid len value if read() fails.


igaw:
Thanks!

PR #788, opened 2024-03-06
Refactor and extend logging/debugging code

igaw:
Let's cleanup some of the debugging mess and make the logging a bit better. The main cleanup is in nvme-cli though. 

I've picked up the patches from #782 for this.


src/nvme/log.h
```diff
@@ -51,12 +51,16 @@ void nvme_set_root(nvme_root_t r);
 /**
  * nvme_set_debug - Set NVMe command debugging output
  * @debug:	true to enable or false to disable
+ *
+ * Don't use it, it's debricated.
```
tbzatek, line 69: -> _deprecated_?

PR #789, opened 2024-03-15
types: Add PEL header additional information definitions

ikegami-t:
The changes for persistent event entry ehai print function.


igaw:
Thanks.

PR #790, opened 2024-03-16
tree: Add nvme_root identify flag to scan NVMe topology

ikegami-t:
This is to avoid nvme identify namespace to scan NVMe topology.

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -281,6 +281,7 @@ LIBNVME_1_0 {
 		nvme_resv_report;
 		nvme_sanitize_nvm;
 		nvme_scan;
+		nvme_scan_no_id;
 		nvme_scan_ctrl;
 		nvme_scan_ctrls;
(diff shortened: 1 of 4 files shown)


ikegami-t:
The nvme identify namespace done by the sysfs ns nuse `ns_head_update_nuse()` and `ns_update_nuse()` as below.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/nvme/host/sysfs.c#n201

ikegami-t:
Okay will update the commit message later.

hreinecke:
Instead of passing flags we should drop the call to 'nvme_scan()' and use the device name directly. Much like we already do for eg 'nvme reset'.

igaw:
We don't need this anymore, fixed it differently: https://github.com/linux-nvme/nvme-cli/commit/a31081a3fa1476ee4e74ec928f51a2879f533963

PR #796, opened 2024-03-19
test: use diff to compare sysfs output

calebsander:
The sysfs tests are currently hand-rolling file comparison. Use `diff -u` instead to get a rich comparison if the comparison fails. This also simplifies the test code significantly.


igaw:
When I added this code I decided against using an external tool, so that we don't have a build dependency on the tool. Maybe add a check to meson an if available use it?

tbzatek:
> When I added this code I decided against using an external tool, so that we don't have a build dependency on the tool. Maybe add a check to meson an if available use it?

FYI, I have already wired `diff` in nbft tests this way:
https://github.com/linux-nvme/libnvme/blob/553c15e0ac3774d8eeb192d7ff3706b4e38945ff/test/nbft/meson.build#L64-L68

igaw:
Yes, and the test will not be executed when diff is missing. So this should handled in the same way.

igaw:
Thanks!

PR #797, opened 2024-03-19
sysfs: minimize heap allocations of sysfs paths

calebsander:
11a0918 ("nvme: allow to overwrite base sysfs path") added support for changing the sysfs path via an environment variable. Unfortunately, it added a heap allocation
every time a sysfs path was requested.

Modify the callers to not free the paths, which allows a string constant to be returned if the path isn't overridden, avoiding an allocation. Cache the path in a static variable so that if it is overridden, the heap-allocated string only needs to be constructed once and afterwards can be reused.

Create a file sysfs.c to consolidate this logic
instead of spreading them across 3 files.
Also introduce a helper to factor out the duplicated code.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -12,6 +12,7 @@ sources = [
     'nvme/ioctl.c',
     'nvme/linux.c',
     'nvme/log.c',
+    'nvme/sysfs.c',
     'nvme/tree.c',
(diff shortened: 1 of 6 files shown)


igaw:
Thanks!

Issue #728, opened 2023-10-10
Use __nvme_alloc helper to allocate payload


igaw:
To ensure the payloads buffer allocation are correctly aligned use the newly added `__nvme_alloc` helper. 

See also https://github.com/linux-nvme/nvme-cli/pull/2051

tbzatek:
Some more fixes: #731 

Not sure about `nvme_get_telemetry_log()` as the buffer gets realloc'd and there's more stuff in `mi.c` that I was not sure about either (`nvme_mi_ep_probe()`).

igaw:
Thanks. I am also not sure about `mi.c`. We could also introduce the realloc helper from nvme-cli if needed.

Another thing we could do is add a check to `nvme_passwthru64` which issue a warning if the alignment is not on a page.

igaw:
The kernel learned to handle unaligned data structures.  So there is no need for this anymore.

PR #798, opened 2024-03-20
linux: Allocate aligned payload for nvme_get_telemetry_log

liang3zy22:
continue to solve issue #728 . Add __nvme_realloc from nvme-cli codes.


igaw:
Thanks! 

Sorry it took so long.

PR #800, opened 2024-03-20
build: Switch default meson buildtype to 'debugoptimized'

tbzatek:
The meson 'debug' buildtype defaults to '-O0 -g' which misses some important compiler warnings (as of gcc-13). While there's no universal solution, the 'debugoptimized' buildtype supplies '-O2 -g' that appears to be a reasonable compromise for having important compiler warnings by default.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,7 +13,7 @@ project(
     default_options: [
         'c_std=gnu99',
         'warning_level=1',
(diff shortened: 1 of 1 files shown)


tbzatek:
Just a backport of my nvme-cli commit from yesterday.

PR #801, opened 2024-03-22
sysfs: Export NVMe controller sysfs directory return function

ikegami-t:
The path is returned with the controller name by the nvme scan. To return the path without the nvme scan nvme identify namespace.

### DIFF
--- a/src/libnvme.h
+++ b/src/libnvme.h
@@ -23,6 +23,7 @@ extern "C" {
 #include "nvme/tree.h"
 #include "nvme/util.h"
 #include "nvme/log.h"
+#include "nvme/sysfs.h"
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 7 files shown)


igaw:
I am not sure if we really need to do this. The sysfs is always mounted under `/sys`. The only reason the mount point is configurable in libnvme is for the unit tests. So we are able to run them against a sysfs mock. Is there any good use case for nvme-cli for this, e.g. do you want to add also some unit tests for nvme-cli which would rely on a sysfs mock? For this purpose I'd say it could make sense. But the current only place in nvme-cli where this export would be used is for mmaping of PCI registers.

So I tend to say no to this PR if we don't have a good use case why we should export it.

ikegami-t:
Okay I will do consider separately if the sysfs access functionality can be moved and combined from the nvme-cli to libnvme.

PR #802, opened 2024-03-24
types: Fix EOM header optional data present fields definition

ikegami-t:
Defined eye data present as bit 0 and printable eye present as bit 1. Fixed eye data present to bit 1 and printable eye present to bit 0. Add the optional data present reserved filed definition and macros to get.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3963,12 +3963,12 @@ struct nvme_phy_rx_eom_log {
 
 /**
  * enum nvme_eom_optional_data - EOM Optional Data Present Fields
- * @NVME_EOM_EYE_DATA_PRESENT:		Eye Data Present
  * @NVME_EOM_PRINTABLE_EYE_PRESENT:	Printable Eye Present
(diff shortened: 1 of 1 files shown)


ikegami-t:
Note: The existing NVME_EOM_EYE_DATA_PRESENT and NVME_EOM_PRINTABLE_EYE_PRESENT macros have been changed.

bpaupore-wdc (commented):
Change seem mostly good to me, though wouldn't removing the original enums cause backwards-compatibility issues?

I'm not sure if this feature is new enough to just do it anyways, or if renaming the macros and fixing but keeping the original definitions would be best.

igaw:
I agree, better leave the old enums around and avoid breakage.  Just add a 'deprecated' comment to them.

ikegami-t:
Thanks for your review comments. Just fixed the patch as mentioned.

bpaupore-wdc (approved):

igaw:
Thanks!

PR #804, opened 2024-03-31
types: Fix NVME_CSTS_SHST_MASK definition

ikegami-t:
Deprecate NVME_CSTS_SHS_MASK instead.


src/nvme/types.h
```diff
@@ -344,13 +344,13 @@ enum nvme_csts {
...
 	NVME_CSTS_CFS_MASK	= 0x1,
-	NVME_CSTS_SHN_MASK	= 0x3,
+	NVME_CSTS_SHST_MASK	= 0x3,
 	NVME_CSTS_NSSRO_MASK	= 0x1,
 	NVME_CSTS_PP_MASK	= 0x1,
 	NVME_CSTS_SHST_NORMAL	= 0,
 	NVME_CSTS_SHST_OCCUR	= 1,
 	NVME_CSTS_SHST_CMPLT	= 2,
-	NVME_CSTS_SHST_MASK	= 3,
+	NVME_CSTS_SHS_MASK	= NVME_CSTS_SHST_MASK, /* Deprecated */
```
igaw, line 353: I think there is a typo, shouldn't it be `NVME_CSTS_SHN_MASK`?

ikegami-t: Sorry for the error. Right just fixed as mentioned.

igaw:
Thanks!

PR #809, opened 2024-04-04
ioctl: Return EPROTO error in nvme_get_log_page()

hreinecke:
The ioctl for nvme_get_log_page() might return with status '0', but that only indicates that a CQE got received. The CQE could still indicate an error, and we should return EPROTO here.


hreinecke:
And I might indicate that this is a general thing; maybe we should modify the ioctl call itself ...

hreinecke:
Superseded by PR#810.

PR #810, opened 2024-04-04
Retry discovery log page on DNR

hreinecke:
The ioctl to issue an NVMe command might complete successfully, but that does not indicate that the command itself was successful. The CQE (ie the 'result' parameter) might be set to indicate the the NVMe command returned with an error.
In these cases we should not return '0' from the ioctl, but rather set 'EPROTO' to indicate that somthing has failed.
This then can be used when reading the discovery log page to check if it's safe to re-read the log page or if one should abort.


igaw:
Quite a mess with the nvme status and ioctl return values. Looks way saner now. Thanks!

src/nvme/fabrics.c
```diff
@@ -1113,6 +1122,9 @@ static struct nvmf_discovery_log *nvme_discovery_log(
 			nvme_msg(r, LOG_INFO,
 				 "%s: discover try %d/%d failed, error %d\n",
 				 name, retries, args->max_retries, errno);
+			if (*log_args.result &&
+			    !(*log_args.result & NVME_SC_DNR))
+				continue;
```
calebsander, line 1127: `continue` in a `do`-`while` loop jumps to the `while` condition: `while (genctr != le64_to_cpu(log->genctr) && ++retries < args->max_retries);`. Is accessing the header valid valid if the command returned an error status? Presumably no data would have been transferred in that case.

igaw: There were some nasty surprises with this series. I've reverted the whole series for now.

PR #811, opened 2024-04-10
ioctl: Add nvme_ns_mgmt_delete_timeout()

ikegami-t:
The existing nvme_ns_mgmt_delete() does not support the timeout.


hreinecke:
Please add it to libnvme.map to have it exported to external callers.

igaw:
These are just static inlines, thus no export map update needed.

igaw:
Thanks!

PR #814, opened 2024-04-14
types: Fix boot partition read offset (BPROF) mask value

ikegami-t:
The register bits defined 29:10 as 20 bits width mask.


igaw:
Thanks!

PR #815, opened 2024-04-15
tree: fix mem leak in nvme_ns_init()

martin-gpy:
Valgrind revealed a mem leak in nvme_ns_init() invoked by several nvme commands such as connect-all, list, disconnect-all, etc. Leak traced to the nvme_ns_id pointer not getting freed after use. Fix the same.


igaw:
Thanks!

PR #816, opened 2024-04-15
type: Add NVME_CMBEBS property definitions

ikegami-t:
Also add NVME_UNIT definitions for CMBSZU and PMRSZU.


src/nvme/types.h
```diff
@@ -491,6 +491,30 @@ enum nvme_cmbsts {
 
 #define NVME_CMBSTS_CBAI(cmbsts)	NVME_GET(cmbsts, CMBSTS_CBAI)
 
+enum nvme_unit {
+	NVME_UNIT_B,
```
igaw, line 495: Since these are types with fixed values, it's better to explicit assign them the values, e.g. `NVME_UNIT_B = 1` as we do in the other types.

ikegami-t: Fixed as mentioned. Thank you.

igaw:
Thanks!

PR #817, opened 2024-04-17
types: Add NVME_CMIC definitions

ikegami-t:
The existing NVME_CTRL_CMIC definitions are complex values.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -1114,6 +1114,25 @@ struct nvme_id_ctrl {
 	__u8			vs[1024];
 };
 
+enum nvme_cmic {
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -1114,6 +1114,25 @@ struct nvme_id_ctrl {
 	__u8			vs[1024];
 };
 
+enum nvme_cmic {
```
igaw, line 1117: I  think it  would be good to add documentation to all the new types you are adding. IIRC, I forgot to ask for this in the last few PRs. It's not a real blocker, though please consider to provide the documentation update in future. Thanks.

ikegami-t: Sure of course will to it later. Thank you.

PR #820, opened 2024-04-24
fabrics: Simple URI parser

tbzatek:
This is an initial implementation for a simple URI parser. Described in NVM Express Boot Specification (TP8012), used in `struct nbft_info_discovery` as a composite connection string, as opposed e.g. to separate struct fields in `struct nbft_info_subsystem_ns`.

Open questions/TODO:
- [x] handle escaped ("percent-encoded") sequences
- [ ] ~~any UTF-8 related tasks? (can't think of any right now)~~
- [x] handle query
- [x] handle fragment
- [x] handle username + password

i.e. as defined in RFC 3986:

```
      URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
```

Applied to NVMe:
```
nvme+tcp://user:pass@host:port/subsys_nqn/nid?name=ferret#nose
```

Since the above tasks may involve `struct nvme_fabrics_uri` element changes, marking as WIP, seeking initial feedback. Any other possible elements to extract that I missed?

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -2,6 +2,8 @@
 LIBNVME_1.10 {
 	global:
 		nvme_init_default_logging;
+		nvme_parse_uri;
+		nvme_free_uri;
 };
 
 LIBNVME_1.9 {
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1703,3 +1703,119 @@ int nvmf_register_ctrl(nvme_ctrl_t c, enum nvmf_dim_tas tas, __u32 *result)
 	 */
 	return nvmf_dim(c, tas, NVMF_TRTYPE_TCP, nvme_get_adrfam(c), "", NULL, result);
 }
+
+struct nvme_fabrics_uri *nvme_parse_uri(const char *str)
+{
+	struct nvme_fabrics_uri *uri;
+	_cleanup_free_ char *scheme = NULL;
+	_cleanup_free_ char *authority = NULL;
+	_cleanup_free_ char *path = NULL;
+	const char *host;
+	int i;
+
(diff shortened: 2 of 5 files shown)


tbzatek:
I think this is ready for review.

martin-belanger:
> What about nvme-stas, any plans for using URIs, @martin-belanger ?

Hi @tbzatek - nvme-stas currently supports URIs read from NBFT records. We convert URIs into the familiar `transport`, `traddr`, and `trsvcid` parameters needed by libnvme. Since nvme-stas is written in Python, I was able to simply use the function [urllib.parse.urlparse()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) without having to write my own URI parser. 

Ref: https://github.com/linux-nvme/nvme-stas/blob/main/staslib/conf.py#L804

tbzatek:
> I was able to simply use the function [urllib.parse.urlparse()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) without having to write my own URI parser.

I see, fair enough :-) I was originally considering using the [uriparser](https://uriparser.github.io/) C library but since it would be pulled in as dependency and we need to pack nvme-cli in initramfs, it was a no-go. Not to mention supporting old distros where we do regular rebases (RHEL 9).

igaw:
> My primary concern is whether this API is universal enough before it's released and frozen

I think we start with what you came up and see how it works out. In the worst case we introduce `nvme_parse_uri2`

igaw:
Thanks!

PR #821, opened 2024-04-25
tests: Add sample NBFT table with mpath IPv4+IPv6 discovery

tbzatek:
A sample table generated by latest EDK2	snapshot:
- two HFIs: static IPv4	and static IPv6
- two boot attempts defined for	each HFI (multipath) with the well-known discovery NQN
- second NIC link down

Result:
- the first boot attempt properly discovered
- the second boot attempt transformed in an SSNS record with the discovery NQN and marked as unavailable
- two Discovery	Descriptor records corresponding with the boot attempts and linked to corresponding SSNS records


igaw:
Thanks!

PR #823, opened 2024-04-26
types: Add enum nvme_unit/cmbebs/cmbswtp/flbas/cmic documentation

ikegami-t:
Since missed to add the documentatin.


igaw:
Thanks!

PR #824, opened 2024-04-26
split logging from root object

igaw:
Let's refactor the logging out from the root object and accept the full defeat against no global variables fight.

### DIFF
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -29,7 +29,7 @@
 static nvme_root_t root;
 
 void __attribute__((format(printf, 4, 5)))
-__nvme_msg(nvme_root_t r, int lvl,
+__nvme_msg(nvme_root_t r, int level,
 	   const char *func, const char *format, ...)
 {
 	FILE *fp = stderr;
@@ -56,9 +56,9 @@ __nvme_msg(nvme_root_t r, int lvl,
 	if (r)
 		fp = r->fp;
 
-	if (r && lvl > r->log_level)
+	if (r && level > r->log_level)
 		return;
-	if (!r && lvl > DEFAULT_LOGLEVEL)
+	if (!r && level > DEFAULT_LOGLEVEL)
 		return;
 
 	if (r && r->log_timestamp) {
--- a/src/nvme/log.c
+++ b/src/nvme/log.c
@@ -26,13 +26,18 @@
 #define LOG_CLOCK CLOCK_MONOTONIC
 #endif
 
-static nvme_root_t root;
+static struct nvme_log def_log = {
+	.fd = STDERR_FILENO,
+	.level = DEFAULT_LOGLEVEL,
+	.pid = false,
+	.timestamp = false,
+};
 
 void __attribute__((format(printf, 4, 5)))
 __nvme_msg(nvme_root_t r, int level,
 	   const char *func, const char *format, ...)
 {
-	FILE *fp = stderr;
+	struct nvme_log *l;
 	va_list ap;
 	char pidbuf[16];
 	char timebuf[32];
@@ -50,18 +55,15 @@ __nvme_msg(nvme_root_t r, int level,
 	_cleanup_free_ char *message = NULL;
 	int idx = 0;
 
-	if (!r)
-		r = root;
-
 	if (r)
-		fp = r->fp;
+		l = &r->log;
+	else
+		l = &def_log;
 
-	if (r && level > r->log_level)
-		return;
-	if (!r && level > DEFAULT_LOGLEVEL)
+	if (level > l->level)
 		return;
 
-	if (r && r->log_timestamp) {
+	if (l->timestamp) {
 		struct timespec now;
 
 		clock_gettime(LOG_CLOCK, &now);
@@ -71,7 +73,7 @@ __nvme_msg(nvme_root_t r, int level,
 	} else
 		*timebuf = '\0';
 
-	if (r && r->log_pid) {
+	if (l->pid) {
 		snprintf(pidbuf, sizeof(pidbuf), "%ld", (long)getpid());
 		idx |= 1 << 1;
 	} else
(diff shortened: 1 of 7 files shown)


src/nvme/private.h
```diff
@@ -160,15 +160,19 @@ struct nvme_fabric_options {
...
 struct nvme_root {
 	char *config_file;
 	char *application;
 	struct list_head hosts;
 	struct list_head endpoints; /* MI endpoints */
-	FILE *fp;
-	int log_level;
-	bool log_pid;
-	bool log_timestamp;
+	struct nvme_log log;
```
ikegami-t, line 175: Is this still needed since seems only the default log data added enough and duplicated. Also if needed seems nvme_root can have a pointer to the default log data.

igaw: Depends. If we want to support different log levels per root object then we should keep it if not then we could rip this part out as well and reduce the complexity. 

Not sure though as it is changing the behavior a bit.

src/nvme/mi.c
```diff
@@ -41,18 +41,26 @@ static bool nvme_mi_probe_enabled_default(void)
...
+	r = calloc(1, sizeof(*r));
 	if (!r) {
+		errno = ENOMEM;
 		return NULL;
 	}
-	r->log_level = log_level;
-	r->fp = stderr;
+
+	if (!fp)
+		fp = stderr;
```
ikegami-t, line 53: Seems usually the input argument `fp` not used as a local variable and also `r->log.fp` can be set `stderr` directly as below.
```c
    r->log.fp = fp ? fp : stderr;
```

src/nvme/tree.c
```diff
@@ -187,19 +187,24 @@ int nvme_scan_topology(struct nvme_root *r, nvme_scan_filter_t f, void *f_args)
...
 		errno = ENOMEM;
 		return NULL;
 	}
-	r->log_level = log_level;
-	r->fp = stderr;
-	if (fp)
-		r->fp = fp;
+
+	if (!fp)
+		fp = stderr;
```
ikegami-t, line 199: Same comment as stderr can be set directly.

src/nvme/private.h
ikegami-t: I see. To make sure let me confirm again. Multiple root objects supported or in future possible to be supported? (I would like confirm the possibility since if no possibility only the default log data enough and the behavior for use not changed I think.)

src/nvme/log.h
```diff
@@ -59,21 +71,24 @@ int nvme_get_logging_level(nvme_root_t r, bool *log_pid, bool *log_tstamp);
...
+ * nvme_init_logging instead.
  */
 void nvme_set_root(nvme_root_t r);
 
 /**
  * nvme_set_debug - Set NVMe command debugging output
  * @debug:	true to enable or false to disable
  *
- * Don't use it, it's debricated.
+ * This function is depricated. Use nvme_init_default_logging instead.
```
tbzatek, line 84: _depricated_ -> _deprecated_

src/nvme/log.h
```diff
@@ -59,21 +71,24 @@ int nvme_get_logging_level(nvme_root_t r, bool *log_pid, bool *log_tstamp);
  * will be set as well. This means the global root object is always pointing to
  * the latest created root object. Note the first @nvme_free_tree call will reset
  * the global root object.
+ *
+ * This funciton is depricated. Use nvme_init_default_logging or/and
```
tbzatek, line 75: _funciton_ -> _function_

src/nvme/log.h
tbzatek: What about adding `__attribute__((deprecated))` annotation for deprecated symbols?

src/nvme/mi.c
```diff
@@ -41,18 +41,26 @@ static bool nvme_mi_probe_enabled_default(void)
...
+
+	r->log.fp = fp;
+	r->log.level = log_level;
+	nvme_init_default_logging(fp, log_level, false, false);
```
tbzatek, line 57: Wait, this would overwrite whatever user has set previously. I understand something like this may be needed to maintain backwards compatibility though.

I believe `nvme_init_default_logging()` should never be called anywhere from the library. For instance, we want to initialize default libnvme logging once on libblockdev startup.

src/nvme/private.h
tbzatek: We do create multiple root objects in libblockdev since it's often used in threads and need separate memory objects. While not currently used, it might be useful to grab the log for certain operations and present it as a buffer back to the user (e.g. via D-Bus). Besides, as backwards compatibility is concerned, I would keep it there. Reconsider later for libnvme 2.0?

src/nvme/tree.c
```diff
@@ -187,19 +187,24 @@ int nvme_scan_topology(struct nvme_root *r, nvme_scan_filter_t f, void *f_args)
...
+
+	r->log.fp = fp;
+	r->log.level = log_level;
+	nvme_init_default_logging(fp, log_level, false, false);
```
tbzatek, line 203: Same concern as above.

tbzatek (commented):
Thanks, I like it. It always puzzled me to have a global root object with dubious memory ownership, though the code appeared safe.

Don't worry too much about global variables, I have a similar mental block for `goto`... :flushed:

src/nvme/private.h
ikegami-t: Understood. Thanks for your explanation.

igaw: @tbzatek thanks for the feedback. Good to know that you are using this feature. In this case, we keep it as it is, just have a global default fallback for all commands. Not nice, but I suppose the ship has sailed and we have to live with it.

Anyway, when we do libnvme 2.0 we could introduce a generic handle for all APIs. This should solve the issue for good.  The downside is that you always have to open/create a handle, though I think this is an acceptable requirement.

src/nvme/mi.c
igaw: I know this is really messy. This is basically in there from the initial attempt to everything working. I was not sure if we want to try to keep as close as possible to the old way. But maybe we should just break it now and the the default  logging API right.

In short, I agree we you, we should drop this hack.

igaw:
> We do create multiple root objects in libblockdev since it's often used in threads and need separate memory objects. While not currently used, it might be useful to grab the log for certain operations and present it as a buffer back to the user (e.g. via D-Bus). 

If I got you right, the idea is to introduce a global log buffer with multiple readers. This indeed would also solve the issue. 

> Besides, as backwards compatibility is concerned, I would keep it there. Reconsider later for libnvme 2.0?

Yeah, let's start with the default global logging fix here and see if anyone is really unhappy with it.  We can discuss/play with the two idea (either introducing a library handle or your idea with a log buffer) for libnvme 2.0.

tbzatek:
> > We do create multiple root objects in libblockdev since it's often used in threads and need separate memory objects. While not currently used, it might be useful to grab the log for certain operations and present it as a buffer back to the user (e.g. via D-Bus).
> 
> If I got you right, the idea is to introduce a global log buffer with multiple readers. This indeed would also solve the issue.

It was more an idea than any proposal and we don't plan to make use of anything like this in the near term.

To be precise, we need multiple root objects to be supported, due to threading. However logging is a different affair, at the moment we basically silence/disable libnvme logging in libblockdev and rely solely on error codes.

My only concern is that whenever a root object takes `fp` or `log_level` args via e.g. `nvme_mi_create_root()` or `nvme_create_root()` and global logging has been initialized already, it should not get overridden. I.e. either use the global logging only or keep it separate for each root object. Theoretically, in a process that pulls multiple libraries in that use libnvme for something, we don't want our global logging settings to be reset by a foreign library. No big deal though, just customers are sometimes picky with any single message on stderr. :expressionless:

src/nvme/log.c
```diff
@@ -26,13 +26,18 @@
 #define LOG_CLOCK CLOCK_MONOTONIC
 #endif
 
-static nvme_root_t root;
+static struct nvme_log def_log = {
+	.fp = NULL,
```
igaw, line 30: Or should we initialize this to stderr?

ikegami-t: If changed the `struct nvme_log` parameter: `FILE *fp` to `int fd` the `STDERR_FILENO` definition can be used instead. By the way the static data initialized by 0 so not necessary to initialize the parameters by `NULL` or `false` I think.

igaw: Good idea with the filedescriptor. Let's do this, so we have sane defaults.

I'll keep the explicit initializing in place, I'd prefer to have all members listed so it's easy to read what the defaults are.

PR #828, opened 2024-04-29
types: Fix PMRWBM register name

ikegami-t:
Incorrectly defined as PMRWMB.
Also add enum nvme_pmrcap documentation.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -561,21 +561,44 @@ enum nvme_crto {
 #define NVME_CRTO_CRIMT(crto)	NVME_GET(crto, CRTO_CRIMT)
 #define NVME_CRTO_CRWMT(crto)	NVME_GET(crto, CRTO_CRWMT)
 
+/**
+ * enum nvme_pmrcap - This field indicates the persistent memory region capabilities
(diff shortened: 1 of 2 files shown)


src/nvme/types.h
```diff
@@ -584,7 +603,7 @@ enum nvme_pmrcap {
 #define NVME_PMRCAP_WDS(pmrcap)		NVME_GET(pmrcap, PMRCAP_WDS)
 #define NVME_PMRCAP_BIR(pmrcap)		NVME_GET(pmrcap, PMRCAP_BIR)
 #define NVME_PMRCAP_PMRTU(pmrcap)	NVME_GET(pmrcap, PMRCAP_PMRTU)
-#define NVME_PMRCAP_PMRWMB(pmrcap)	NVME_GET(pmrcap, PMRCAP_PMRWMB)
+#define NVME_PMRCAP_PMRWBM(pmrcap)	NVME_GET(pmrcap, PMRCAP_PMRWBM)
```
igaw, line 606: Maybe keep a compat around to avoid unnecessary incompatibility?

ikegami-t: Changed the patch to remain the current definitions also as deprecated. Thank you.

igaw:
Thanks!

PR #833, opened 2024-05-03
linux: Fix uninitialized variables

tbzatek:
In file included from ../src/nvme/linux.c:40:
In function ‘freep’,
    inlined from ‘nvme_get_telemetry_log’ at ../src/nvme/linux.c:169:23:
../src/nvme/cleanup.h:24:9: warning: ‘log’ may be used uninitialized [-Wmaybe-uninitialized]
   24 |         free(*(void **)p);
      |         ^~~~~~~~~~~~~~~~~
../src/nvme/linux.c: In function ‘nvme_get_telemetry_log’:
../src/nvme/linux.c:169:30: note: ‘log’ was declared here
  169 |         _cleanup_free_ void *log;
      |                              ^~~

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -296,8 +296,8 @@ int nvme_get_new_host_telemetry(int fd, struct nvme_telemetry_log **log,
 
 int nvme_get_lba_status_log(int fd, bool rae, struct nvme_lba_status_log **log)
 {
+	_cleanup_free_ struct nvme_lba_status_log *buf = NULL;
 	__u32 size;
-	_cleanup_free_ struct nvme_lba_status_log *buf;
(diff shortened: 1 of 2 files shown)


tbzatek:
Funny that `-Wmaybe-uninitialized` alone would not trigger the compiler warning, it needs `-fexceptions` (which is not universally suitable for everyone). `gcc version 14.0.1 20240411`

igaw:
Though, the variable gets initialized before we exit the function. So this seems to be a recurring problem. Should we just make it the default to always to initialize all of them?

tbzatek:
Yeah, sounds like a good practice, though enforcing it may require a macro or something

PR #836, opened 2024-05-04
mi: fix the rc for nvme_mi_scan_ep

jinliangw:
nvme_mi_scan_ep should return error code of calling function.
This will help the upper layer know the reason of failure.

### DIFF
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -638,6 +638,7 @@ int nvme_mi_admin_admin_passthru(nvme_mi_ctrl_t ctrl, __u8 opcode, __u8 flags,
 	struct nvme_mi_admin_req_hdr req_hdr;
 	struct nvme_mi_resp resp;
 	struct nvme_mi_req req;
+	unsigned int timeout_save;
 	int rc;
 	int direction = opcode & 0x3;
 	bool has_write_data = false;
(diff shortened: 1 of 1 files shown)


src/nvme/mi.c
```diff
@@ -327,7 +327,7 @@ int nvme_mi_scan_ep(nvme_mi_ep_t ep, bool force_rescan)
 
 	rc = nvme_mi_mi_read_mi_data_ctrl_list(ep, 0, &list);
 	if (rc)
-		return -1;
+		return rc;
```
igaw, line 330: The library uses the errno style error handling approach. So the return code `-1` is correct. But I see that `nvme_mi_mi_read_mi_data_ctrl_list` is not following this approach. Thus I think you need to update that code.

igaw:
@jk-ozlabs

src/nvme/mi.c
igaw: Indeed :)

src/nvme/mi.c
```diff
@@ -700,7 +696,17 @@ int nvme_mi_admin_admin_passthru(nvme_mi_ctrl_t ctrl, __u8 opcode, __u8 flags,
 		resp.data_len = data_len;
 	}
 
+	if (timeout_ms != 0) {
+		/* When user forgets to set timeout_ms, the timeout_ms will be 0.
```
jk-ozlabs, line 700: The logic here is sound, but you're documenting the timeout == 0 case in the condition where the timeout is not zero.

Perhaps something like:

```
    /* if the user has specified a custom timeout, save the current timeout and override */
```

\- it's not necessarily an error ("user forgets") to pass zero here.

src/nvme/mi.c
```diff
@@ -638,6 +638,7 @@ int nvme_mi_admin_admin_passthru(nvme_mi_ctrl_t ctrl, __u8 opcode, __u8 flags,
 	struct nvme_mi_admin_req_hdr req_hdr;
 	struct nvme_mi_resp resp;
 	struct nvme_mi_req req;
+	unsigned int ori_timeout;
```
jk-ozlabs, line 641: Super minor, but maybe call this `timeout_save`

jk-ozlabs (commented):
Code looks good, I'd just suggest a couple of minor doc/naming changes.

As a more general (later) change, should we pass the timeout into `nvme_mi_submit`, rather than doing the save & restore?

src/nvme/mi.c
jinliangw: Done.

jk-ozlabs (approved):
LGTM, thanks!

igaw:
Thanks!

PR #837, opened 2024-05-04
mi: fix the return error code.

jinliangw:
When returning a system error code (i.e. rc < 0), the function should return -1 and set errno instead of returning the errno directly. It is very helpful for the client to handle the rc in an unified way.


igaw:
Thanks!

PR #839, opened 2024-05-09
types: add the nvme Cancel command opcode

maurizio-lombardi:
Defined by TP4097

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -8248,6 +8248,7 @@ enum nvme_data_tfr {
  * @nvme_cmd_resv_acquire:	Reservation Acquire
  * @nvme_cmd_io_mgmt_recv:	I/O Management Receive
  * @nvme_cmd_resv_release:	Reservation Release
+ * @nvme_cmd_cancel:		Cancel
  * @nvme_cmd_copy:		Copy
  * @nvme_cmd_io_mgmt_send:	I/O Management Send
  * @nvme_zns_cmd_mgmt_send:	Zone Management Send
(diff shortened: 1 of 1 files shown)


maurizio-lombardi:
It fails the build here on github but I've just added an enum... must be a problem in the tests.

martin-belanger:
> It fails the build here on github but I've just added an enum... must be a problem in the tests.

Could it be that the new enum is missing documentation?

maurizio-lombardi:
> > It fails the build here on github but I've just added an enum... must be a problem in the tests.
> 
> Could it be that the new enum is missing documentation?

You're probably right, the errors are a bit cryptic but in any case I have to add the new opcode to the docs

maurizio-lombardi:
much better

doc/man/nvme_io_opcode.2
```diff
@@ -56,6 +56,10 @@ enum nvme_io_opcode {
 , 
 .br
 .br
+.BI "    nvme_cmd_cancel"
+,
+.br
+.br
```
igaw, line 62: Please drop the generated doc changes. The docs get generated on release time anyway. Thanks

maurizio-lombardi: Done, is it ok now?

igaw:
Thanks! 

Should we also add the rest of the TP? Not sure if it would be useful from userspace.

maurizio-lombardi:
> Thanks!
> 
> Should we also add the rest of the TP? Not sure if it would be useful from userspace.

I am going to add some parts to nvme-cli because I have to fix things like this:

`
# nvme effects-log /dev/nvme0n1 -H
[...]
I/O Commands
IOCS9    [Dataset Management              ] 00000001  CSUPP+  LBCC-  NCC-  NIC-  CCC-  USS-  No command restriction
IOCS24   [Unknown                         ] 00000001  CSUPP+  LBCC-  NCC-  NIC-  CCC-  USS-  No command restriction
IOCS121  [Zone Management Send            ] 00000003  CSUPP+  LBCC+  NCC-  NIC-  CCC-  USS-  No command restriction

`

See this "Unknown" thing? It's the cancel command

PR #841, opened 2024-05-22
Add support for atomic, piecewise ANA log fetch and fix bugs around ANA log functions

calebsander:
The existing function `nvme_get_log_ana()` for fetching the ANA log
fills the provided ANA log page buffer in 4 KB increments.

This has two main drawbacks: it does not guarantee the log page is
fetched atomically, so the returned log page may be torn,
and it may read past the end of the ANA log page.
Not only does reading past the end of the log page entail
wasted commands and data transfers, the NVMe specification
says controllers should abort Get Log Page commands
that specify LPOL/LPOU values past the end of the log page.
nvme-cli sizes the ANA log page buffer based on the maximum
number of ANA groups and namespaces the controller supports,
which may be significantly larger than the current number.

Introduce a function `nvme_get_ana_log_atomic()` to fix these behaviors.
It parses the log page header and ANA group descriptors
and stops issuing Get Log Page commands once it has the full log page.
If multiple Get Log Page commands were required, it also fetches
the header again to verify the change count is still the same.
This is analogous to the mechanism implemented in `nvme_discovery_log()`
that checks the Discovery Log Page GENCTR to guarantee atomicity.

Add unit tests for this new function using the mock ioctl infrastructure

Also fix a few bugs found along the way:
- `nvme_get_log_ana_groups()` takes the wrong pointer type (ANA group descriptor instead of ANA log page header)
- `nvme_get_log_ana()` ignores the `rae` argument
- References to `struct nvme_ana_rsp_hdr` in the doc comments, which doesn't exist

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -355,6 +355,131 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
 	return 0;
 }
 
+static int read_ana_chunk(int fd, enum nvme_log_ana_lsp lsp, bool rae,
+			  __u8 *log, __u8 **read, __u8 *to_read, __u8 *log_end)
+{
+	if (to_read > log_end) {
+		errno = ENOSPC;
+		return -1;
+	}
+
+	while (*read < to_read) {
+		__u32 len = min(log_end - *read, NVME_LOG_PAGE_PDU_SIZE);
+		int ret;
+
+		ret = nvme_get_log_ana(fd, lsp, rae, *read - log, len, *read);
+		if (ret)
+			return ret;
+
+		*read += len;
+	}
+	return 0;
+}
+
+static int try_read_ana(int fd, enum nvme_log_ana_lsp lsp, bool rae,
+			struct nvme_ana_log *log, __u8 *log_end,
+			__u8 *read, __u8 **to_read, bool *may_retry)
+{
+	__u16 ngrps = le16_to_cpu(log->ngrps);
+
+	while (ngrps--) {
+		__u8 *group = *to_read;
+		int ret;
+		__le32 nnsids;
+
+		*to_read += sizeof(*log->descs);
+		ret = read_ana_chunk(fd, lsp, rae,
+				     (__u8 *)log, &read, *to_read, log_end);
+		if (ret) {
+			/*
+			 * If the provided buffer isn't long enough,
+			 * the log page may have changed while reading it
+			 * and the computed length was inaccurate.
+			 * Have the caller check chgcnt and retry.
+			 */
+			*may_retry = errno == ENOSPC;
+			return ret;
+		}
+
+		/*
+		 * struct nvme_ana_group_desc has 8-byte alignment
+		 * but the group pointer is only 4-byte aligned.
+		 * Don't dereference the misaligned pointer.
+		 */
+		memcpy(&nnsids,
+		       group + offsetof(struct nvme_ana_group_desc, nnsids),
+		       sizeof(nnsids));
+		*to_read += le32_to_cpu(nnsids) * sizeof(__le32);
+		ret = read_ana_chunk(fd, lsp, rae,
+				     (__u8 *)log, &read, *to_read, log_end);
+		if (ret) {
+			*may_retry = errno == ENOSPC;
+			return ret;
+		}
+	}
+
+	*may_retry = true;
+	return 0;
+}
+
+int nvme_get_ana_log_atomic(int fd, bool rgo, bool rae, unsigned int retries,
+			    struct nvme_ana_log *log, __u32 *len)
+{
+	const enum nvme_log_ana_lsp lsp =
+		rgo ? NVME_LOG_ANA_LSP_RGO_GROUPS_ONLY : 0;
+	/* Get Log Page can only fetch multiples of dwords */
+	__u8 * const log_end = (__u8 *)log + (*len & -4);
+	__u8 *read = (__u8 *)log;
+	__u8 *to_read;
+	int ret;
+
+	if (!retries) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	to_read = (__u8 *)log->descs;
+	ret = read_ana_chunk(fd, lsp, rae,
+			     (__u8 *)log, &read, to_read, log_end);
+	if (ret)
+		return ret;
+
+	do {
+		bool may_retry = false;
+		int saved_ret;
+		int saved_errno;
+		__le64 chgcnt;
+
+		saved_ret = try_read_ana(fd, lsp, rae, log, log_end,
+					 read, &to_read, &may_retry);
+		/*
+		 * If the log page was read with multiple Get Log Page commands,
+		 * chgcnt must be checked afterwards to ensure atomicity
+		 */
+		*len = to_read - (__u8 *)log;
+		if (*len <= NVME_LOG_PAGE_PDU_SIZE || !may_retry)
+			return saved_ret;
+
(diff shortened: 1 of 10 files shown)


src/nvme/ioctl.c
```diff
@@ -355,6 +355,130 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
 	return 0;
 }
 
+static int read_ana_chunk(int fd, enum nvme_log_ana_lsp lsp, bool rae,
+			  __u8 *log, __u8 **read, __u8 *to_read, __u8 *log_end)
+{
+	if (to_read > log_end) {
+		errno = ENOSPC;
+		return -1;
```
igaw, line 363: Internal function can just return the error code. We will drop the POSIX style of error codes in the next mayor library version.

src/nvme/ioctl.c
```diff
@@ -355,6 +355,130 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
...
+	}
+
+	while (*read < to_read) {
+		__u32 len = min(log_end - *read, NVME_LOG_PAGE_PDU_SIZE);
```
igaw, line 367: Isn't this also chucking the reads to 4k max?

src/nvme/ioctl.c
```diff
@@ -355,6 +355,130 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
...
+
+	to_read = (__u8 *)log->descs;
+	ret = read_ana_chunk(fd, lsp, rae,
+			     (__u8 *)log, &read, to_read, log_end);
+	if (ret)
+		return ret;
+
+	do {
+		bool may_retry = false;
+		int saved_ret = try_read_ana(fd, lsp, rae, log, log_end,
```
igaw, line 449: please have the `try_read_call` call not mix into the deceleration block.

src/nvme/ioctl.c
```diff
@@ -355,6 +355,130 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
...
+			  __u8 *log, __u8 **read, __u8 *to_read, __u8 *log_end)
+{
+	if (to_read > log_end) {
+		errno = ENOSPC;
+		return -1;
+	}
+
+	while (*read < to_read) {
+		__u32 len = min(log_end - *read, NVME_LOG_PAGE_PDU_SIZE);
+		int ret = nvme_get_log_ana(fd, lsp, rae, *read - log, len, *read);
```
igaw, line 368: please have the `nvme_get_log_ana` call not mixed into the variable deceleration block.

src/nvme/ioctl.c
```diff
@@ -355,6 +355,130 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
...
+			return ret;
+
+		if (log->chgcnt == chgcnt) {
+			/* Log hasn't changed; return try_read_ana() result */
+			errno = saved_errno;
+			return saved_ret;
+		}
+	} while (--max_retries);
+
+	errno = EAGAIN;
```
igaw, line 478: Why `EAGAIN`? Don't we end up here when max_retries hits 0?

src/nvme/ioctl.c
igaw: Forget it, I start to understand what this code is doing.

src/nvme/ioctl.c
```diff
@@ -355,6 +355,130 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
...
+		}
+	}
+
+	*may_retry = true;
+	return 0;
+}
+
+int nvme_get_ana_log_atomic(
+	int fd, bool rgo, bool rae, unsigned int max_retries,
+	struct nvme_ana_log *log, __u32 *len)
```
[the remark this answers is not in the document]
igaw: Just a bikeshed question: the name 'atomic' is heavily overloaded and I understood it wrong initially. I was expecting that the this code will read everything in one go, which it isn't. It carefully makes sure it reads the data with a understanding of the data types/sizes/alignments within the log page. Maybe we can come up with something else as name? But I know naming is diffucult...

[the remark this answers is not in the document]
calebsander: I removed the line break before the start of the arguments, hopefully this is what you were asking for?

src/nvme/ioctl.c
igaw: We do this in the rest of the code (or should), that is only the function which are exported should touch errno. The reason I want to keep it this is way, is that when we drop the errno usage, we don't have to review internal functions.

src/nvme/ioctl.c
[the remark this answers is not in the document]
igaw: aligment: all other functions in this file align the arguments with the open bracket. checkpatch is accepting different styles but doesn't report inconsistency. I don't care  too much about one or other style but I do care about consistency. 

naming: Whenever I read atomic I think in terms of cmpxchg. So we either do the whole operation or nothing. When I was reviewing it, it took me a while to figure out what you mean with atomic. The implementation makes sure it reads the individual parts of the log page in one go, respecting the type/size of each part. Sure, this can be considered 'atomic' but the whole function as such is not doing this as atomic operation. I got even more confused by 'chunk' function which I suppose should the the 'atomic' part? Anyway, I agree with Christoph and think the term atomic is heavily overloaded in it's meaning which can cause confusion.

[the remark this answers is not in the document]
calebsander: aligment (*sic*): I assume by "bracket" you mean the open parenthesis? I believe I have already made this change, can you please let me know if the updated review is to your aesthetic preferences?

naming: I don't mean to be pedantic here, but using a sequence number to ensure multiple reads are performed coherently certainly seems to qualify for the computer science definition of "atomic". "atomic" comes from a Greek root meaning roughly "indivisible". In computer science, it is widely used to describe any interaction between concurrent computations that is performed in an all-or-nothing way. The computation may involve multiple steps, in which case they are made to appear as a single unit. Typically this is done with some form of mutual exclusion or repeating a computation until the value it operates on doesn't change in the middle. cmpxchg is certainly an example of an atomic operation, and is the central building block for that "restartable computation" approach to perform more complex operations atomically. (Keep in mind that the processor's own implementation of cmpxchg requires multiple steps--read, compare, write--that have to be synchronized to make the operation appear atomic. So by your reasoning, a cmpxchg instruction wouldn't count as atomic either.) And in fact what we are doing here with the ANA change count is exactly a cmpxchg loop, except the cmpxchg is just an atomic read because there is no need to modify the change count:
```
1. Read change count
2. Read rest of ANA log page
3. Read change count again; if changed, goto 2
```
(If the whole log page fits in 4 KB, then only a single Get Log Page command is used, which is defined in the spec to be atomic. In that case the change count check is skipped.)
I am not wedded to "atomic", though. I have already suggested "untorn" as a possible alternative, and I'll suggest another: "coherent". Please let me know if you prefer one of those, or have your own suggestion. But "the term atomic...can cause confusion" is not really constructive feedback...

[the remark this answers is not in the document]
igaw: Sorry for the late response.

Formatting: I know, it sometimes a PITA but such things matter none the less.

Naming: The point I wanted to make is that the context is important. Not everyone understands one word the same way. Anyway, I am fine as it is. So let's go with it.

src/nvme/ioctl.c
igaw: Makes sense. Let's fix this after this merge.

PR #842, opened 2024-05-28
types: add enum for Protection Information Format

francispravin5:
Add enum for Protection Information Format (PIF).


igaw:
Thanks!

PR #843, opened 2024-05-28
mi: Probe quirks on first command submission

amboar:
Currently drive quirks are probed when opening a connection to the MI endpoint. The probe itself is the only IO that's required in the process of establishing the connection. We have an opportunity to make connection-establishment non-blocking if we can move the probe IO elsewhere.

We know the user is expecting IO to the endpoint when they issue a command. Given this, exploit the call to nvme_mi_submit() to establish the drive quirks.

### DIFF
--- a/src/nvme/mi-mctp.c
+++ b/src/nvme/mi-mctp.c
@@ -506,8 +506,6 @@ nvme_mi_ep_t nvme_mi_open_mctp(nvme_root_t root, unsigned int netid, __u8 eid)
 	 */
 	ep->timeout = 5000;
 
-	nvme_mi_ep_probe(ep);
-
 	return ep;
 
 err_free_rspbuf:
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -143,6 +143,20 @@ void nvme_mi_ep_probe(struct nvme_mi_ep *ep)
 	struct nvme_mi_ctrl *ctrl;
 	int rc;
 
+	/* Ensure the probe occurs at most once. This isn't just to mitigate doubling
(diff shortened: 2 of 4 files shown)


amboar:
@jk-ozlabs and I have been chatting about this, let me know what you think of the approach!

jk-ozlabs:
Looks good! I think removing the (potentially-doing-IO) quirk probe from the open is a net win here.

We may find that we need to move the probe to before the MIC calculation at some point in future, but that would be driven by a requirement that we don't yet have (eg, a quirk that affects MIC calculation :smile:  )

jk-ozlabs (approved):

igaw:
the test seems to break:

```
 25/27 mi                                               FAIL            0.15s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=235 LD_LIBRARY_PATH=/__w/libnvme/libnvme/.build-ci/src /__w/libnvme/libnvme/.build-ci/test/test-mi
```

amboar:
> the test seems to break:
> 
> ```
>  25/27 mi                                               FAIL            0.15s   killed by signal 6 SIGABRT
> >>> MALLOC_PERTURB_=235 LD_LIBRARY_PATH=/__w/libnvme/libnvme/.build-ci/src /__w/libnvme/libnvme/.build-ci/test/test-mi
> ```

Looking at that now.

amboar:
@igaw I've updated the MI test suite to disable the quirk probing by default, and then provided an explicit test for the behaviour.

igaw:
Thanks!

PR #844, opened 2024-05-31
ioctl: update nvme_set_var_size_tags() to properly use reftag

hasnpol:
be_reftag variable is in size of 64 bits, in case of reftag argument set to 1, be_reftag will be equal to 0x0100000000000000. (Since version 1.7 included cpu_to_be64() function) 

In case of NVME_NVM_PIF_16B_GUARD, bit masking the be_reftag above with 0xffffffff (size of cdw14) will clear the cdw14 to 0 causing some drives to not being able to set ILBRT/EILBRT.

I offer to shift be_reftag by 32 to the right so that drives can correctly set ILBRT/EILBRT field.

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -1670,32 +1670,30 @@ static int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw1
 		__u8 pif, __u8 sts, __u64 reftag, __u64 storage_tag)
 {
 	__u32 cdw2 = 0, cdw3 = 0, cdw14;
-	beint64_t be_reftag = cpu_to_be64(reftag);
-	beint64_t be_storage_tag = cpu_to_be64(storage_tag);
 
 	switch (pif) {
 	case NVME_NVM_PIF_16B_GUARD:
-		cdw14 = be_reftag & 0xffffffff;
-		cdw14 |= ((be_storage_tag << (32 - sts)) & 0xffffffff);
+		cdw14 = reftag & 0xffffffff;
+		cdw14 |= ((storage_tag << (32 - sts)) & 0xffffffff);
 		break;
 	case NVME_NVM_PIF_32B_GUARD:
-		cdw14 = be_reftag & 0xffffffff;
-		cdw3 = be_reftag >> 32;
-		cdw14 |= ((be_storage_tag << (80 - sts)) & 0xffff0000);
+		cdw14 = reftag & 0xffffffff;
+		cdw3 = reftag >> 32;
+		cdw14 |= ((storage_tag << (80 - sts)) & 0xffff0000);
 		if (sts >= 48)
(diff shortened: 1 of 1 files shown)


src/nvme/ioctl.c
```diff
@@ -1675,7 +1675,7 @@ static int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw1
 
 	switch (pif) {
 	case NVME_NVM_PIF_16B_GUARD:
-		cdw14 = be_reftag & 0xffffffff;
+		cdw14 = (be_reftag >> 32) & 0xffffffff;
```
bpaupore-wdc, line 1678: Yeah, this is a problem where the host system's 0xffffffff doesn't match the big-endian reference tag. Unfortunately I think the fix offered here would cause issues in big-endian environments by shifting the tag to 0.

Swapping to big-endian might need to happen after adjusting to fit the actual size/constraints of the PIF, for both the reference and storage tags.

igaw: indeed, the cpu_to_be64 should be done after the mask operation.

hasnpol: I just updated my PR for your comments. 

However, I believe we should be using cpu_to_be32 since the cdw sizes are in 32 bits. In case of using cpu_to_be64, even after the mask operation, the values will be cleared to 0.

Please let me know what you think.

src/nvme/ioctl.c
```diff
@@ -1670,42 +1670,40 @@ static int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw1
...
 {
 	__u32 cdw2 = 0, cdw3 = 0, cdw14;
-	beint64_t be_reftag = cpu_to_be64(reftag);
-	beint64_t be_storage_tag = cpu_to_be64(storage_tag);
 
 	switch (pif) {
 	case NVME_NVM_PIF_16B_GUARD:
-		cdw14 = be_reftag & 0xffffffff;
-		cdw14 |= ((be_storage_tag << (32 - sts)) & 0xffffffff);
+		cdw14 = (reftag >> 32) & 0xffffffff;
```
bpaupore-wdc, line 1676: The shift here would still reduce any valid 32-bit tag to 0, but otherwise with that removed the updated changes look good to me.

```suggestion
		cdw14 = reftag & 0xffffffff;
```

hasnpol: Thanks for pointing that out, PR is updated!

bpaupore-wdc (approved):

igaw:
Could you please merge all the commits into one commit and add your Signed-off-by, thanks!

hasnpol:
Merged all my commits into one, should be all set right now!

igaw:
Great! I presume I can add your sign-off-by tag?

igaw:
I see you figured it out, but just in case you need some more info, see https://github.com/linux-nvme/nvme-cli?tab=readme-ov-file#how-to-contribute

Thanks!

PR #846, opened 2024-06-13
types: add new field added in TP4141a

francispravin5:
Add the new field added in TP4141a and enums required for future implementation.


igaw:
Thanks!

PR #849, opened 2024-06-25
linux: avoid potential null pointer dereference

maurizio-lombardi:
if the "result" pointer is null because of a memory allocation failure, do not dereference it and do not free the old pointer.


igaw:
Thanks!

PR #850, opened 2024-06-25
mi-mctp: initialize the rc variable in handle_mctp_endpoint()

maurizio-lombardi:
Keep static analyzers happy.


igaw:
Though the checker is correct, there is a case where rc is evaluated even though it's never set.

igaw:
Thanks!

PR #851, opened 2024-06-28
types: add PEL vendor specific and TCG defined events definitions

ikegami-t:
The PEL feature added by NVMe revision 1.4 with the event types. But the event types were not implemented by the nvme-cli.


igaw:
Thanks!

PR #855, opened 2024-06-30
tree: fix segfault in nvme_free_tree()

martin-gpy:
Commands like nvme list & list-subsys currently segfault for --help or any invalid command option as shown below:

nvme list -h
Usage: nvme list <device> [OPTIONS]

Retrieve basic information for all NVMe namespaces

Options:
  [  --verbose, -v ]                    --- output verbosity
  [  --output-format=<FMT>, -o <FMT> ]  --- format: normal|json|binary

This is due to an invalid dereferencing of the nvme_root_t object in nvme_free_tree(). Fix this by checking whether this object is valid before dereferencing and freeing it. And while we are it, ensure r->options is also valid before freeing it.


igaw:
Thanks!

PR #856, opened 2024-07-01
linux: add nvme_revoke_tls_key

igaw:
Add a functio to revoke a TLS key from a keyring.

see #https://github.com/linux-nvme/nvme-cli/issues/2365


igaw:
I suppose this should be called revoke as there is no remove operation.

PR #857, opened 2024-07-03
tree: add helper to lookup hostnqn/hostid

igaw:
There are several places where the default hostnqn/hostid could be stored. Add a helper which ensures a canonical way to lookup these values. Also this function warns if the hostnqn/hostid do not match as the Linux kernel enforces this.


src/nvme/tree.c
```diff
@@ -117,16 +117,99 @@ static void cleanup_dirents(struct dirents *ents)
...
+	nvme_host_t h;
+
+	/* command line argumments */
+	if (hostid_arg)
+		hid = strdup(hostid_arg);
+	if (hostnqn_arg)
+		hnqn = strdup(hostnqn_arg);
+
+	/* JSON config: assume the first entry is the default host */
+	h = nvme_first_host(r);
```
igaw, line 147: This doesn't seem to do what I though it would do. Needs some more tinkering.

test/config/config-dump.c
```diff
@@ -0,0 +1,53 @@
...
+}
+
+int main(int argc, char *argv[])
+{
+	bool pass = true;
+
+	config_dump(argv[1]);
+	fflush(stdout);
+
+	exit(pass ? EXIT_SUCCESS : EXIT_FAILURE);
```
calebsander, line 52: Return value from `config_dump()` (`pass`) is ignored, so this will always succeed?

igaw: oops, that is obviously wrong. thanks for catching it!

PR #858, opened 2024-07-04
build: add cleanup.h install

ikegami-t:
To use the cleanup functions by nvme-cli.


igaw:
I'd rather not. This is a private header. We don't want to distribute.
We should only install those headers listed in `libnvme.h`

PR #860, opened 2024-07-10
types: add new field added in TP4090

nitin-sao:
As per TP4090, a new field support is added as NPDGL (Namespace Preferred Deallocate Granularity Large) and so the bit-field of NSFEAT, OPTPERF get extended by 1 bit.


igaw:
Thanks!

PR #861, opened 2024-07-11
types: add new field added in TP4099

nitin-sao:
As per TP4090, a new field is added as MEM (MDTS and Size Limits Exclude Metadata) under nvme_id_ctrl_ctratt.


igaw:
Thanks!

PR #863, opened 2024-07-13
tree: fix nvme_read_config() to not set errno if return 0

ikegami-t:
It should not touch errno but open() set value in json_read_config().


ikegami-t:
Fix for the bug mentioned by https://github.com/linux-nvme/nvme-cli/pull/2414#discussion_r1676296002.

ikegami-t:
Thank you too.

PR #864, opened 2024-07-16
tree: fix segfault in nvme_ctrl_lookup_phy_slot()

gjoyce-ibm:
Not all directories have an address. Verify addr before calling strcmp().

Fixes: 42ac45359635 ("tree: Add PCI physical slot number for controller")

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -1784,6 +1784,10 @@ static char *nvme_ctrl_lookup_phy_slot(nvme_root_t r, const char *address)
 				return NULL;
 			}
 			addr = nvme_get_attr(path, "address");
+
+			/* some directories don't have an address entry */
+			if (!addr)
+				continue;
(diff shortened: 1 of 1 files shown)


shroffni:
Looks good to me.

Thanks,
--Nilay

src/nvme/tree.c
```diff
@@ -1784,6 +1784,10 @@ static char *nvme_ctrl_lookup_phy_slot(nvme_root_t r, const char *address)
 				return NULL;
 			}
 			addr = nvme_get_attr(path, "address");
+
+			/* some directories don't have an address entry */
+			if (addr == NULL)
```
martin-gpy, line 1789: A cleaner way to do this would be ```if (!addr)```

gjoyce-ibm: I guess that I'm a K&R guy but if that's the style for the library, I can change it.

igaw:
Thanks!

PR #865, opened 2024-07-17
types: Added new controller attribute as per TP4104

nitin-sao:
As per TP4104, a new controller attribute support is added as HMBR (HMB Restrict Non-Operational Power State Access) under nvme_id_ctrl_ctratt.


igaw:
Thanks!

PR #867, opened 2024-07-18
linux: Remove the use of OpenSSL Engine API

tbzatek:
OpenSSL engines are not FIPS compatible and corresponding API is deprecated since OpenSSL 3.0. It appears this API is not actually used in the code, so remove it.

----

This is a wild shot and I wouldn't pretend I know what I'm doing. This simple change does however seem to do the trick. Tested only against openssl 3.0. Affects `nvme_gen_dhchap_key()`, tested through `nvme gen-dhchap-key`.

OpenSSL engine support is being removed from Fedora/CentOS Stream/RHEL.

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -18,7 +18,6 @@
 #include <unistd.h>
 
 #ifdef CONFIG_OPENSSL
-#include <openssl/engine.h>
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
 #include <openssl/kdf.h>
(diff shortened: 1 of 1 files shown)


igaw:
Okay, also no clue when it comes to this crypto stuff. I just wonder if we should/could add a simple unit test then we could try to hook up the CI build to build against 1.1.x and 3.0.

igaw:
> Tested only against openssl 3.0.

but you are touching the openssl 1 code path. no wonder it just works :)

hreinecke:
If it works with both OpenSSL 1.x and OpenSSL 3 I'm fine with this change.

tbzatek:
> > Tested only against openssl 3.0.
> 
> but you are touching the openssl 1 code path. no wonder it just works :)

Ahhh, looks like my recent vacation was a good brain-resetting one.

I've managed to find my old Gentoo machine, tested against `dev-libs/openssl-1.1.1t-r1` with no further changes to the patch:
```
$ ./nvme gen-dhchap-key
DHHC-1:00:BSuKcXtaRAEa3lQWkAmD1G2+2YGhZUwa652PdtQetiZKbjvu:
$ ldd /tmp/nvme-cli/.build/./subprojects/libnvme/src/libnvme.so.1
	libcrypto.so.1.1 => /usr/lib64/libcrypto.so.1.1 (0x00007fe0128b1000)
```

My original test was against Fedora `openssl-3.2.2-4.fc41.x86_64`

igaw:
Okay, let's merge it then. If you have time to work an unit test I wouldn't mind :)

PR #868, opened 2024-07-19
mi: add service control primitive command

zhangjian3032:
The `breaked` command will left the SSD in Transmit Mode.
Add API to send the `abort` command to exit the Transmit Mode to
avoid the SSD being stuck in the Transmit Mode.

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -1,3 +1,8 @@
+LIBNVME_MI_1_11 {
+	global:
+		nvme_mi_control;
+};
+
 LIBNVME_MI_1_10 {
 	global:
 		nvme_mi_admin_get_ana_log_atomic;
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -529,6 +529,31 @@ static void nvme_mi_admin_init_resp(struct nvme_mi_resp *resp,
 	resp->hdr_len = sizeof(*hdr);
 }
 
+static void nvme_mi_control_init_req(struct nvme_mi_req *req,
+				    struct nvme_mi_control_req *control_req,
+				    __u8 opcode, __u16 cpsp)
+{
+	memset(req, 0, sizeof(*req));
+	memset(control_req, 0, sizeof(*control_req));
+
+	control_req->hdr.type = NVME_MI_MSGTYPE_NVME;
+	control_req->hdr.nmp = (NVME_MI_ROR_REQ << 7) |
+		(NVME_MI_MT_CONTROL << 3); /* we always use command slot 0 */
+	control_req->opcode = opcode;
+	control_req->cpsp = cpu_to_le16(cpsp);
+
+	req->hdr = &control_req->hdr;
+	req->hdr_len = sizeof(*control_req);
+}
+
+static void nvme_mi_control_init_resp(struct nvme_mi_resp *resp,
+				    struct nvme_mi_control_resp *control_resp)
+{
+	memset(resp, 0, sizeof(*resp));
+	resp->hdr = &control_resp->hdr;
+	resp->hdr_len = sizeof(*control_resp);
+}
+
 static int nvme_mi_admin_parse_status(struct nvme_mi_resp *resp, __u32 *result)
 {
 	struct nvme_mi_admin_resp_hdr *admin_hdr;
@@ -578,6 +603,26 @@ static int nvme_mi_admin_parse_status(struct nvme_mi_resp *resp, __u32 *result)
 		*result = nvme_result;
 
 	return nvme_status;
+}
+
+static int nvme_mi_control_parse_status(struct nvme_mi_resp *resp, __u16 *cpsr)
+{
+	struct nvme_mi_control_resp *control_resp;
+
+	if (resp->hdr_len < sizeof(*control_resp)) {
+		errno = -EPROTO;
+		return -1;
+	}
+	control_resp = (struct nvme_mi_control_resp *)resp->hdr;
+
+	if (control_resp->status)
+		return control_resp->status |
+			(NVME_STATUS_TYPE_MI << NVME_STATUS_TYPE_SHIFT);
+
+	if (cpsr)
+		*cpsr = le16_to_cpu(control_resp->cpsr);
+
+	return control_resp->status;
 }
 
 int nvme_mi_admin_xfer(nvme_mi_ctrl_t ctrl,
@@ -808,6 +853,29 @@ int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
 		errno = EPROTO;
 		return -1;
 	}
+
+	return 0;
+}
+
+int nvme_mi_control(nvme_mi_ep_t ep, __u8 opcode,
+		    __u16 cpsp, __u16 *result_cpsr)
+{
+	struct nvme_mi_control_resp control_resp;
+	struct nvme_mi_control_req control_req;
+	struct nvme_mi_resp resp;
+	struct nvme_mi_req req;
+	int rc = 0;
+
+	nvme_mi_control_init_req(&req, &control_req, opcode, cpsp);
+	nvme_mi_control_init_resp(&resp, &control_resp);
+
+	rc = nvme_mi_submit(ep, &req, &resp);
+	if (rc)
+		return rc;
+
+	rc = nvme_mi_control_parse_status(&resp, result_cpsr);
+	if (rc)
+		return rc;
 
 	return 0;
 }
--- a/src/nvme/mi.h
+++ b/src/nvme/mi.h
@@ -368,6 +368,52 @@ struct nvme_mi_admin_resp_hdr {
 } __attribute__((packed));
 
 /**
+ * enum nvme_mi_control_opcode - Operation code for Control Primitives.
+ * @nvme_mi_control_opcode_pause: Suspend response transmission/timeout
+ * @nvme_mi_control_opcode_resume: Resume from a paused condition
+ * @nvme_mi_control_opcode_abort: Re-initialize a Command Slot to the Idle state
+ * @nvme_mi_control_opcode_get_state: Get the state of a Command Slot
+ * @nvme_mi_control_opcode_replay: Retransmit the Response Message
+ */
+enum nvme_mi_control_opcode {
+	nvme_mi_control_opcode_pause			= 0x00,
+	nvme_mi_control_opcode_resume			= 0x01,
+	nvme_mi_control_opcode_abort			= 0x02,
+	nvme_mi_control_opcode_get_state	= 0x03,
+	nvme_mi_control_opcode_replay			= 0x04,
+};
+
+/**
+ * struct nvme_mi_control_req - The Control Primitive request.
+ * @hdr: Generic MI message header
+ * @opcode: Control Primitive Opcodes (using &enum nvme_mi_control_opcode)
+ * @tag: flag - Opaque value passed from request to response
+ * @cpsp: Control Primitive Specific Parameter
+ *
+ */
+struct nvme_mi_control_req {
+	struct nvme_mi_msg_hdr hdr;
+	__u8	opcode;
+	__u8	tag;
+	__le16	cpsp;
+} __attribute((packed));
+
+/** struct nvme_mi_control_resp - The Control Primitive response.
+ * @hdr: Generic MI message header
+ * @status: Generic response code, non-zero on failure
+ * @tag: flag - Opaque value passed from request to response
+ * @cpsr: Control Primitive Specific Response
+ *
+ */
+
+struct nvme_mi_control_resp {
+	struct nvme_mi_msg_hdr hdr;
+	__u8	status;
+	__u8	tag;
+	__le16	cpsr;
+} __attribute((packed));
+
+/**
  * nvme_mi_status_to_string() - return a string representation of the MI
  * status.
  * @status: MI response status
@@ -1074,6 +1120,24 @@ static inline int nvme_mi_admin_identify(nvme_mi_ctrl_t ctrl,
 	return nvme_mi_admin_identify_partial(ctrl, args,
 					      0, NVME_IDENTIFY_DATA_SIZE);
 }
+
+/**
+ * nvme_mi_control() - Perform a Control Primitive command
+ * @ep: endpoint for MI communication
+ * @opcode: Control Primitive opcode (using &enum nvme_mi_control_opcode)
+ * @cpsp: Control Primitive Specific Parameter
+ * @result_cpsr: Optional field to return the result from the CPSR field
+ *
+ * Perform a Control Primitive command, using the opcode specified in @opcode
+ * Stores the result from the CPSR field in @result_cpsr if set.
+ *
+ * Return: 0 on success, non-zero on failure
+ *
+ * See: &enum nvme_mi_control_opcode
+ *
+ */
+int nvme_mi_control(nvme_mi_ep_t ep, __u8 opcode,
+		    __u16 cpsp, __u16 *result_cpsr);
 
 /**
  * nvme_mi_admin_identify_cns_nsid() - Perform an Admin identify command using
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -241,6 +241,104 @@ int do_identify(nvme_mi_ep_t ep, int argc, char **argv)
 	return 0;
 }
 
+int do_control_primitive(nvme_mi_ep_t ep, int argc, char **argv)
+{
+	int rc = 0;
+	char *action = NULL;
+	uint8_t opcode = 0xff;
+	uint16_t cpsr;
+
+	static const struct {
+		const char *name;
+		uint8_t opcode;
+	} control_actions[] = {
+		{ "pause", nvme_mi_control_opcode_pause },
+		{ "resume", nvme_mi_control_opcode_resume },
+		{ "abort", nvme_mi_control_opcode_abort },
+		{ "get-state", nvme_mi_control_opcode_get_state },
+		{ "replay", nvme_mi_control_opcode_replay },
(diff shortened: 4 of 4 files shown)


igaw:
@jk-ozlabs could have a look? Thanks!

jk-ozlabs:
Sure! Assigning.

src/nvme/mi.c
```diff
@@ -580,6 +603,36 @@ static int nvme_mi_admin_parse_status(struct nvme_mi_resp *resp, __u32 *result)
...
 
+static int nvme_mi_service_parse_status(struct nvme_mi_resp *resp, __le16 *cpsr)
+{
+	struct nvme_mi_service_resp_hdr *resp_hdr;
+
+	/* we have a few different sources of "result" here: the status header
+	 * in the MI response, the cdw3 status field, and (command specific)
+	 * return values in cdw0. The latter is returned in the result pointer,
+	 * the former two generate return values here
+	 */
```
jk-ozlabs, line 614: This comment looks to have been copy-pasted from elsewhere, but doesn't apply to the control primitive messages: there is no cdw0/cdw3 status.

src/nvme/mi.h
```diff
@@ -367,6 +367,34 @@ struct nvme_mi_admin_resp_hdr {
...
+ * @opcode: servicing command opcode (using &enum nvme_service_opcode)
+ * @tag: flag - be used to sent and return
+ * @cpsp: Control Primitive Specific Parameter
+ */
+struct nvme_mi_service_req_hdr {
+	struct nvme_mi_msg_hdr hdr;
+	__u8	opcode;
+	__u8	tag;
+	__le16	cpsp;
+} __attribute((packed));
```
jk-ozlabs, line 382: This isn't just the header, right? It's the structure of the entire control primitive message. In which case, we should name it accordingly - maybe just remove the `_hdr` suffix.

src/nvme/mi.h
```diff
@@ -367,6 +367,34 @@ struct nvme_mi_admin_resp_hdr {
...
+ * @tag: flag - Be used to sent and return
+ * @cpsr: Control Primitive Specific Response
+ */
+
+struct nvme_mi_service_resp_hdr {
+	struct nvme_mi_msg_hdr hdr;
+	__u8	status;
+	__u8	tag;
+	__le16	cpsr;
+} __attribute((packed));
```
jk-ozlabs, line 396: And same here, this is the entire response, not just the header.

src/nvme/mi.c
```diff
@@ -812,6 +865,38 @@ int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
...
+	struct nvme_mi_service_resp_hdr resp_hdr;
+	struct nvme_mi_service_req_hdr req_hdr;
+	struct nvme_mi_resp resp;
+	struct nvme_mi_req req;
+	int rc = 0;
+	__le16 cpsr = 0xffff;
+
+	nvme_mi_service_init_req(&req, &req_hdr, opcode);
+
+	nvme_mi_calc_req_mic(&req);
```
jk-ozlabs, line 879: The MIC should be calculated by `nvme_mi_submit()`, you don't need this.

src/nvme/mi.c
```diff
@@ -580,6 +603,36 @@ static int nvme_mi_admin_parse_status(struct nvme_mi_resp *resp, __u32 *result)
...
+	if (resp->hdr_len < sizeof(*resp_hdr)) {
+		errno = -EPROTO;
+		return -1;
+	}
+	resp_hdr = (struct nvme_mi_service_resp_hdr *)resp->hdr;
+
+	/* If we have a MI error, we can't be sure there's an admin header
+	 * following; return just the MI status, with the status type
+	 * indicator of MI.
+	 */
```
jk-ozlabs, line 625: This comment (and logic) is also incorrect, there is no admin header expected here.

Since we have checked that the size is valid to hold the entire response, there's no conditional parsing of the response required; just extract the status and cpsr values.

src/nvme/mi.c
```diff
@@ -812,6 +865,38 @@ int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
...
+	rc = nvme_mi_submit(ep, &req, &resp);
+	if (rc)
+		return rc;
+
+	rc = nvme_mi_service_parse_status(&resp, &cpsr);
+	if (rc)
+		return rc;
+
+	if (result_cpsr)
+		*result_cpsr = le16_to_cpu(cpsr);
```
jk-ozlabs, line 892: You're double-handling the cpsr field here. Probably best to perform the endian conversion in _parse_status, and set the cspr value even on non-zero status.

src/nvme/mi.c
```diff
@@ -812,6 +865,38 @@ int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
...
+
+	nvme_msg(ep->root, LOG_DEBUG,
+			 "servie primitive opcode=%#x, crsp is %#x\n", opcode, cpsr);
```
jk-ozlabs, line 895: two typos in this string, `servie` and `crsp`. You probably want to print this before the early return on non-zero status too, for easier error debugging.

src/nvme/mi.c
```diff
@@ -812,6 +865,38 @@ int nvme_mi_admin_identify_partial(nvme_mi_ctrl_t ctrl,
 	return 0;
 }
 
+int nvme_mi_service_primitive_control(nvme_mi_ep_t ep, __u8 opcode, __u16 *result_cpsr)
```
jk-ozlabs, line 868: Since this is new (public) API, we will need the cpsp value argument too, otherwise we will have to break compatibility this when introducing support for commands that pass a cpsp value.

src/libnvme-mi.map
```diff
@@ -61,6 +61,8 @@ LIBNVME_MI_1_1 {
 		nvme_mi_open_mctp;
 		nvme_mi_scan_mctp;
 		nvme_mi_scan_ep;
+		nvme_mi_service_primitive_control;
+
```
jk-ozlabs, line 65: minor: no need for this blank line

jk-ozlabs (changes requested):
Overall, it's nice to be adding support for the control primitive messages, so thanks for sending this through.

However, could you also include some information about your intended use-case, and how the command servicing states will interact with libnvme's own command interactions? From your commit message:

> The `breaked` command will left the SSD in Transmit Mode.
> Add API to send the `abort` command to exit the Transmit Mode to
> avoid the SSD being stuck in the Transmit Mode.

I'm not clear on what you mean by "breaked" command, could you elaborate on that?

If the device is in the wrong mode, then we may want libnvme to be performing that recovery, rather than passing the issue to the API caller. Which breakages are you seeing cause this problem? I have never needed an Abort to recover any failed commands in my testing, perhaps this is device-specific?

Also, the "service" naming you've suggested doesn't really match the spec. These are "control primitive" messages, which affect the message servicing model, but the latter applies to all command servicing anyway. I would suggest using the `nvme_mi_control_primitive` prefix (or similar) instead of `nvme_mi_service`.

A few changes requested inline too.

jk-ozlabs:
CC @amboar for the device recovery paths

src/nvme/mi.c
zhangjian3032: >  I'm not clear on what you mean by "breaked" command, could you elaborate on that?

yep, see NVM-Express-Management-Interface-Specification-1.2d-2024.01.10-Ratified.pdf, 
`Figure 34: Command Servicing State Diagram`

Here is the English translation:

For example, when NVMe transmits messages to the Management Controller (like BMC, hereinafter referred to as BMC), the state changes to "Transmit". It remains in this state until the message transmission completes, after which it reverts to "Idle".

In cases where multiple messages are needed to assemble a complete MI message, such as the admin Identify message with an MTU of 64, receiving 4096 bytes results in approximately 60-70 packets. If the BMC stops receiving halfway through, NVMe remains in the "Transmit" state without transitioning to "Idle", preventing further communication. Consultation with NVMe vendors such as Samsung and others suggests sending an abort command to return the NVMe controller to the `Idle` state.

Regarding the scenario where BMC stops receiving halfway:
1. BMC Reboot.
2. BMC uses commands like Ctrl-C during the read process to actively abort.

zhangjian3032: Overall, the "Transmit" mentioned above is just one possible case. There are many other scenarios that could cause NVMe to enter a paused state. For details, refer to Section 4.2.1.1 "Pause".

jk-ozlabs: > yep, see NVM-Express-Management-Interface-Specification-1.2d-2024.01.10-Ratified.pdf, `Figure 34: Command Servicing State Diagram`

OK, cool. I'd suggest referring to the states & transitions in your commit message then.

> In cases where multiple messages are needed to assemble a complete MI message, such as the admin Identify message with an MTU of 64, receiving 4096 bytes results in approximately 60-70 packets. If the BMC stops receiving halfway through, NVMe remains in the "Transmit" state without transitioning to "Idle", preventing further communication. Consultation with NVMe vendors such as Samsung and others suggests sending an abort command to return the NVMe controller to the `Idle` state.

In that example, libnvme should probably be responsible for detecting the loss of communication with the device and issuing the abort. We'll still need the patches you have here for the Control Primitive implementations, but we can't just hand the connectivity problem off to the caller.

@amboar may have some good ideas on where this recovery process may be best implemented.

> Regarding the scenario where BMC stops receiving halfway:
> 
>     1. BMC Reboot.
> 
>     2. BMC uses commands like Ctrl-C during the read process to actively abort.

Agree on (1), but not on (2): the BMC's i2c layer will still ACK all of the incoming MCTP packets.

zhangjian3032: >  but not on (2): the BMC's i2c layer will still ACK all of the incoming MCTP packets.

Maybe because of my hardware design, the NVMe under the I2C switch being preempted by other threads causes NVMe to fail to send messages to BMC fully.

src/libnvme-mi.map
igaw: And it should go to a new section, see the libnvme.map file as example.

src/nvme/mi.c
jk-ozlabs: Yes, that's a good point: if the flow is destroyed, we will release the lock on the multiplexer, which may then switch away from the NVMe endpoint, so the i2c transfer may not be acked.

zhangjian3032: > In that example, libnvme should probably be responsible for detecting the loss of communication with the device and issuing the abort. We'll still need the patches you have here for the Control Primitive implementations, but we can't just hand the connectivity problem off to the caller.

 my current implementation involves checking the `paused` flag both when the caller fails and during initialization. Perhaps my commit was a bit rough; my main intention was to raise this issue and gather opinions, particularly from the `mi Servicing` team.

In my view, for a CLI, it's feasible to check every time. However, for a BMC, which may be constantly monitoring, I'm considering implementing some kind of error handler logic on my end.

src/nvme/mi.c
zhangjian3032: okay, done

src/nvme/mi.c
zhangjian3032: remove this log

src/nvme/mi.c
zhangjian3032: done

src/libnvme-mi.map
zhangjian3032: done, put it in the 1_10

src/nvme/mi.c
zhangjian3032: yup, remove it

src/nvme/mi.c
zhangjian3032: yup, remove it

src/nvme/mi.h
zhangjian3032: done

src/nvme/mi.h
zhangjian3032: done

jk-ozlabs:
Thanks for the updates, all is looking pretty good now!

I had a chat with @amboar about the recovery parts, and I think regardless of what we want to do for that implementation, having the control primitives exposed through the API will still be useful.

Just one thing remaining though: the naming. The term "service" isn't used in the spec in relation to the control primitive commands, please replace the `nvme_mi_service_control` prefixes with `nvme_mi_control` (or even `nvme_mi_control_primitive` if you prefer something more verbose)

This is to align with the other namespacing we use at the moment:

 * `nvme_mi` for all of the MI support
 * `nvme_mi_mi` for the direct MI-level commands
 * `nvme_mi_admin` for the Admin-over-MI commands

So we would be adding `nvme_mi_control` for the Control Primitive commands, which I think is the most consistent with both the spec categorisation and the existing APIs.

zhangjian3032:
@jk-ozlabs super.
*  I using `nvme_mi_control` for all struct and function define, and fix some typo, kindly to review
* put `nvme_mi_control` to LIBNVME_MI_1_11 map, is right here?

src/nvme/mi.c
```diff
@@ -529,6 +529,30 @@ static void nvme_mi_admin_init_resp(struct nvme_mi_resp *resp,
 	resp->hdr_len = sizeof(*hdr);
 }
 
+static void nvme_mi_control_init_req(struct nvme_mi_req *req,
+				   struct nvme_mi_control_req *control_req, __u8 opcode, __u16 cpsp)
```
jk-ozlabs, line 533: Super minor, but looks like the wrapping has gone wrong here (and in following prototypes). Only need to address if you're re-rolling for other reasons though!

src/nvme/types.h
```diff
@@ -7452,6 +7452,14 @@ enum nvme_admin_opcode {
 	nvme_admin_get_lba_status	= 0x86,
 };
 
+enum nvme_control_opcode {
+	nvme_control_pause		= 0x00,
+	nvme_control_resume		= 0x01,
+	nvme_control_abort		= 0x02,
+	nvme_control_get_state	= 0x03,
+	nvme_control_replay		= 0x04,
+};
```
jk-ozlabs, line 7461: These declarations introduce some fairly generic names; I'd suggest making those mi-specific, similar to the MI-command opcode definitions: https://github.com/linux-nvme/libnvme/blob/55e29b85c9b85e69aff7b43dbd509b7be21b93ea/src/nvme/mi.h#L217-L222)

\- and I would suggest defining these in `mi.h` rather than the more-generic `types.h`

How about:

```c
enum nvme_mi_control_opcode {
	nvme_mi_control_opcode_pause		= 0x00,
	nvme_mi_control_opcode_resume		= 0x01,
	nvme_mi_control_opcode_abort		= 0x02,
	nvme_mi_control_opcode_get_state	= 0x03,
	nvme_mi_control_opcode_replay		= 0x04,
}
```

src/nvme/mi.h
```diff
@@ -367,6 +367,34 @@ struct nvme_mi_admin_resp_hdr {
...
+	__le16	cpsp;
+} __attribute((packed));
+
+/** struct nvme_mi_control_resp - The Control Primitive response.
+ * @hdr: Generic MI message header
+ * @status: Generic response code, non-zero on failure
+ * @tag: flag - Be used to sent and return, reserved now
+ * @cpsr: Control Primitive Specific Response
+ */
+
```
jk-ozlabs, line 390: minor: extra newline here

src/nvme/mi.h
```diff
@@ -367,6 +367,34 @@ struct nvme_mi_admin_resp_hdr {
 	__le32	cdw0, cdw1, cdw3;
 } __attribute__((packed));
 
+/**
+ * struct nvme_mi_control_req - The Control Primitive request.
+ * @hdr: Generic MI message header
+ * @opcode: Control Primitive Opcodes (using &enum nvme_control_opcode)
+ * @tag: flag - Be used to sent and return, reserved now
```
jk-ozlabs, line 374: Not sure this description hre (and below) makes sense. How about: 

```@tag: opaque value passed from request to response```

File: src/nvme/mi.h [hunk 14]
```diff
@@ -367,6 +367,34 @@ struct nvme_mi_admin_resp_hdr {
 	__le32	cdw0, cdw1, cdw3;
 } __attribute__((packed));
 
+/**
+ * struct nvme_mi_control_req - The Control Primitive request.
+ * @hdr: Generic MI message header
+ * @opcode: Control Primitive Opcodes (using &enum nvme_control_opcode)
```
jk-ozlabs, line 373: Don't forget to replace this `nvme_control_opcode` reference if you rename the enum.

jk-ozlabs (commented):
Looks good - just a minor request with the opcode definition, but otherwise all good.

@igaw : any preferences on the lib versioning for additions to the `.map`?

src/nvme/types.h
zhangjian3032: nice, done

src/nvme/mi.h
zhangjian3032: done

src/nvme/mi.h
zhangjian3032: done

src/nvme/mi.h
zhangjian3032: yup

src/nvme/mi.c
zhangjian3032: do you means the format, if yes, done~

igaw:
> @igaw : any preferences on the lib versioning for additions to the `.map`?

It looks alright now. We just had the 1.10 release, so adding the new symbol to 1.11 is correct thing.

igaw:
Thanks a lot!

src/nvme/mi.c
jk-ozlabs: Yep, that's what I was referring to.

jk-ozlabs:
Nice one, thanks for the contribution!

Issue #870, opened 2024-07-25
Wrong TLS v1 key generation


hreinecke:
For 'version 0' TLS PSKs the spec states:

The TLS PSK shall be derived as follows from an input PSK (i.e., either a retained PSK or a generated
PSK) and a PSK identity using the HKDF-Extract and HKDF-Expand-Label operations (refer to RFC 5869
and RFC 8446) where the hash function is the one specified by the hash specifier of the PSK identity:
1. PRK = HKDF-Extract(0, Input PSK); and
2. TLS PSK = HKDF-Expand-Label(PRK, “nvme-tls-psk”, PskIdentity, L),
where PskIdentity is the PSK identity and L is the output size in bytes of
the hash function (i.e., 32 for SHA-256 and 48 for SHA-384).

For 'version 1' TLS PSKs the spec states:
The TLS PSK shall be derived as follows from an input PSK (i.e., either a retained PSK or a generated
PSK) and a PSK identity using the HKDF-Extract and HKDF-Expand-Label operations (refer to RFC 5869
and RFC 8446) where the hash function is the one specified by the hash specifier of the PSK identity:
1. PRK = HKDF-Extract(0, Input PSK); and
2. TLS PSK = HKDF-Expand-Label(PRK, “nvme-tls-psk”, Context, L),
where Context is the hash identifier indicated in the PSK identity concatenated to a space
character and to the Base64 PSK digest (i.e., “`<hash> <PSK digest>`”) and L is the output size in bytes of
the hash function (i.e., 32 for SHA-256 and 48 for SHA-384).

IE for 'version 0' the 'label' argument for HKDF-Expand-Label is the PSK identity, and for 'version 1' the 'label' argument
for HKDF-Expand-Label is "`<hash> <PSK digest>`".

The current code always used the PSK identity as the 'label' argument.

PR #871, opened 2024-07-25
linux: update TLS version 1 PSK derivation

hreinecke:
The original implementation for handling 'version 1' TLS PSK derivation was incomplete as the 'label' argument for the HKDF-Expand-Label functionality was not according to the spec.
Update the implementation to match the spec.

Fixes: #870

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -551,22 +551,18 @@ static int derive_retained_key(int hmac, const char *hostnqn,
 	return -1;
 }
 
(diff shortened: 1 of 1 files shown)


hreinecke:
Kudos to Eric Biggers <ebiggers@kernel.org> for pointing this out.

igaw:
Updated the function arguments types, so the match all witch each other.  One of the cross builds failed because the types were not matching.

PR #872, opened 2024-08-01
types: add new fields added in TP4165

francispravin5:
Add the new fields added in TP4165 and enum for newly added action type. Also, modified the nvme_lba_status_atype enum documentation.


igaw:
Thanks!

PR #876, opened 2024-08-05
types: Added new field CSER in enum as per TP4167

nitin-sao:
As per TP4167, a new field is added as CSER (Command Submission and Execution Relaxations) under nvme_cmd_effects enum. Updated field value for NVME_CMD_EFFECTS_CSE_MASK as CSE takes 3 bits.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -3301,6 +3301,7 @@ struct nvme_cmd_effects_log {
  * @NVME_CMD_EFFECTS_NCC:	Namespace Capability Change
  * @NVME_CMD_EFFECTS_NIC:	Namespace Inventory Change
  * @NVME_CMD_EFFECTS_CCC:	Controller Capability Change
+ * @NVME_CMD_EFFECTS_CSER_MASK:	Command Submission and Execution Relaxations
  * @NVME_CMD_EFFECTS_CSE_MASK:	Command Submission and Execution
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -3310,7 +3311,8 @@ enum nvme_cmd_effects {
 	NVME_CMD_EFFECTS_NCC		= 1 << 2,
 	NVME_CMD_EFFECTS_NIC		= 1 << 3,
 	NVME_CMD_EFFECTS_CCC		= 1 << 4,
-	NVME_CMD_EFFECTS_CSE_MASK	= 3 << 16,
+	NVME_CMD_EFFECTS_CSER_MASK	= 3 << 14,
+	NVME_CMD_EFFECTS_CSE_MASK	= 7 << 16,
```
igaw, line 3315: ```NVME_CMD_EFFECTS_CSE_MASK``` update looks wrong. The bitfield lenght should stay at 3, no?

nitin-sao: Hi @igaw,
As per NVMe Base Spec 2.0d, Command Submission and Execution (CSE) field is of 3 bits (18:16). From which only 000b, 001b and 010b define something and 011b to 111b are reserved for now.
Older implementation implies bitfield length of 2, so I updated it to 7 to make the bitfield length 3.

igaw: You are right and even wrote it in the commit message. 'sorry for the noise.

PR #879, opened 2024-08-08
types: Add enum for Completion Condition of Get LBA status command

francispravin5:
Add enum for Completion Condition (CMPC) field of Get LBA status command.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -5193,6 +5193,22 @@ struct nvme_lba_status {
 };
 
 /**
+ * enum nvme_lba_status_cmpc - Get LBA Status Command Completion Condition
+ * @NVME_LBA_STATUS_CMPC_NO_CMPC:	No indication of the completion condition
+ * @NVME_LBA_STATUS_CMPC_INCOMPLETE:	Command completed, but additional LBA Status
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -5192,6 +5192,20 @@ struct nvme_lba_status {
...
+/**
+ * enum nvme_lba_status_cmpc - Get LBA Status Command Completion Condition
+ * @NVME_LBA_STATUS_CMPC_INCOMPLETE:	Command completed, but additional LBA Status
+ *					Descriptor Entries are available to transfer
+ *					or scan did not complete (if ATYPE = 10h)
+ * @NVME_LBA_STATUS_CMPC_COMPLETE:	Completed the specified action over the number
+ *					of LBAs specified in the Range Length field and
+ *					transferred all available LBA Status Descriptors
+ */
+enum nvme_lba_status_cmpc {
```
igaw, line 5204: Should we also add the `no indication of completion condition` value?

francispravin5: Yes, it may require. Will add it in the enum.
Thanks!

igaw:
Thanks!

PR #880, opened 2024-08-09
ioctl: refactoring set_features

sc108-lee:
Use nvme_set_features_data, if setfeature use cdw11 and data field Use NVME_UUID_NONE instead of 0


igaw:
Thanks!

PR #881, opened 2024-08-13
types: added new fields in nvme_nvme_id_ns

ankitvvsoni:
Adopted new fields from I/O Command Set Specific Identify Namespace Data Structure as per NVM Command Set Specification, Revision 1.1.


igaw:
Thanks!

PR #882, opened 2024-08-26
types: add kv opcodes

laiaferrer:
KV Command Opcodes added in a struct different from the I/O due to duplication of some of the opcodes

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -8351,6 +8351,32 @@ enum nvme_io_opcode {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


igaw (changes requested):
Please add something like `types: add kv opcodes` as commit title, instead of `KV Opcodes`. 

And use the same prefix for all the commands so we have a clear split between the `nvme_io_commands` and the `nvme_kv_opcodes`. Hmm maybe it also should be names `nvme_kv_commands` to be consistent?

laiaferrer:
I have created another pull request with the changes needed. Thank you!

PR #883, opened 2024-09-10
 Types: add kv opcodes 

laiaferrer:
KV Opcodes have been added. Changes suggested on the other pull request have all been done, everything is consistent now


igaw:
I've merged all commit into one and  updated the enum name `opcode` instead `commands`.  Pushed the result directly to master. Thanks!

PR #885, opened 2024-09-15
types: add NVME_CHECK() definition to check nvme register field value

ikegami-t:
This checks the value to compare the field value and returns the result.


igaw:
Thanks!

PR #887, opened 2024-09-18
libnvme: add lockdown log page support(LID : 0x14)

ankitvvsoni:
Adding command APIs for inband, out of band, structures and enums for lockdown log page (LID : 0x14).

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -4659,6 +4659,20 @@ struct nvme_supported_cap_config_list_log {
 };
 
 /**
+ * struct nvme_lockdown_log - Command and Feature Lockdown Log
+ * @cfila:	Contents of the Command and Feature Identifier List field in the log page.
+ * @rsvd1:	Reserved
+ * @lngth:	Length of Command and Feature Identifier List field
+ * @cfil:	Command and Feature Identifier List
+ */
+struct nvme_lockdown_log {
+	__u8	cfila;
+	__u8	rsvd1[2];
+	__u8	lngth;
+	__u8	cfil[508];
+};
+
+/**
  * struct nvme_resv_notification_log - Reservation Notification Log
  * @lpc:	Log Page Count
  * @rnlpt:	See &enum nvme_resv_notify_rnlpt.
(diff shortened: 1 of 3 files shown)


src/nvme/types.h
```diff
@@ -4958,6 +4972,46 @@ enum nvme_fdp_config_fdpa {
...
 
+/**
+ * enum nvme_lockdown_log_scope - lockdown log page scope attributes
+ * @ADMIN_CMD:		Scope value for Admin commandS
+ * @FEATURE_ID:		Scope value for Feature ID
+ * @MI_CMD_SET:		Scope value for Management Interface commandS
+ * @PCI_CMD_SET:	Scope value for PCI commands
+ */
+enum nvme_lockdown_log_scope {
+	ADMIN_CMD	= 0x0,
```
igaw, line 4983: Please prefix all the enums accordingly (also further down).

src/nvme/types.h
```diff
@@ -4658,6 +4658,20 @@ struct nvme_supported_cap_config_list_log {
...
 
+/**
+ * struct nvme_lockdown_log - Command and Feature Lockdown Log
+ * @csss:	Contents Selected and Scope Selected
+ * @rsvd1:	Reserved
+ * @length:	Length of Command and Feature Identifier List field
+ * @cfil:	Command and Feature Identifier List
+ */
+struct nvme_lockdown_log {
+	__u8	csss;
```
igaw, line 4669: The field is named `cfila` and the spec, please use this name here.

src/nvme/types.h
```diff
@@ -4658,6 +4658,20 @@ struct nvme_supported_cap_config_list_log {
...
+	__u8	rsvd1[2];
+	__u8	length;
```
igaw, line 4671: `lngth` is used as field name in the spec, I  think we should also use it here, just to be consistent.

igaw (changes requested):
same naming updates needed, rest looks good.

ankitvvsoni:
@igaw, addressed review comments.

ankitvvsoni:
hi @igaw ,
Changes you have suggested are addressed. can you provide your review?

Thanks

igaw:
Sorry for the late response. Lost in the inbox situation ... 
Anyway, thanks!

PR #893, opened 2024-10-10
libnvme: add EMVS support to sanitize command

francispravin5:
Add Enter Media Verification State (EMVS) support to sanitize command.
TP4152 - Post-Sanitize Media Verification 2024.04.01 Ratified.


src/nvme/api-types.h
```diff
@@ -523,6 +524,7 @@ struct nvme_sanitize_nvm_args {
 	__u8 owpass;
 	bool oipbp;
 	bool nodas;
+	bool emvs;
```
igaw, line 527: In order to make the `sizeof` trick work the newly added member of the struct needs to start on a new boundary (8 bytes in this case), and it looks like it does:
```
struct nvme_sanitize_nvm_args {
        __u32 *                    result;               /*     0     8 */
        int                        args_size;            /*     8     4 */
        int                        fd;                   /*    12     4 */
        enum nvme_sanitize_sanact  sanact;               /*    20     4 */
        __u32                      ovrpat;               /*    24     4 */
        _Bool                      ause;                 /*    28     1 */
        __u8                       owpass;               /*    29     1 */
        _Bool                      oipbp;                /*    30     1 */
        _Bool                      nodas;                /*    31     1 */

        /* size: 32, cachelines: 1, members: 10 */
        /* last cacheline: 32 bytes */
};
```
Nice :+1:

igaw:
Thanks!

Issue #890, opened 2024-09-27
Update JSON handling for TLS keys


martin-gpy:
As per Hannes:

JSON handling needs to be updated with the latest TLS changes upstream, which introduced a 'tls_configured_key' in addition to the existing 'tls_key' to differentiate between the TLS key used by the current connection (the 'tls_key' attribute) and the TLS key _specified_ on the command line (the 'tls_configured_key' attribute).

With that change we can format the JSON configuration correctly, and only specify the TLS key in the JSON configuration if it had been configured on the command line.

martin-gpy:
There are also other problems here. It seems the TLS identity version is hardcoded to 0 in the JSON TLS handling:

https://github.com/linux-nvme/libnvme/blob/master/src/nvme/json.c#L52

igaw:
The specs states under ```3.6.1.3 TLS PSK and PSK Identity Derivation```

```
The retained PSK is derived from the configured PSK (refer to section 3.6.1.4). The configured PSK shall
be destroyed as soon as the retained PSK is generated and stored.
```

The current code is actually doing this, it outputs the retained PSK to the JSON file after a successful `nvme connect`.

```shell
# nvme check-tls-key -i -I 1 -c nqn.io-1 -n nqn.2014-08.org.nvmexpress:uuid:befdec4c-2234-11b2-a85c-ca77c773af36 -d NVMeTLSkey-1:01:PMmRzMi+sRVtOSc6qpz8DaNszOufkp/+P1pp+M6r6/cGUjM9: 
# nvme connect -t tcp -a 192.168.154.148 -s 4420 -q nqn.2014-08.org.nvmexpress:uuid:befdec4c-2234-11b2-a85c-ca77c773af36 -n nqn.io-1 --tls --tls_key=811559380 -O
connecting to device: nvme1
[
  {
    "hostnqn":"nqn.2014-08.org.nvmexpress:uuid:2cd2c43b-a90a-45c1-a8cd-86b33ab273b6",
    "hostid":"2cd2c43b-a90a-45c1-a8cd-86b33ab273b6",
    "subsystems":[
      {
        "nqn":"nqn.io-1",
        "ports":[
          {
            "transport":"tcp",
            "traddr":"192.168.154.144",
            "trsvcid":"4421",
            "dhchap_key":"none"
          },
          {
            "transport":"tcp",
            "traddr":"192.168.154.144",
            "trsvcid":"4420",
            "dhchap_key":"none"
          }
        ]
      }
[...]

PR #894, opened 2024-10-10
TLS keys handling

igaw:
Refactor and extend the PSK API and decouple the key store operations from the JSON parsing.

Fixes: #890

### DIFF
--- a/doc/config-schema.json.in
+++ b/doc/config-schema.json.in
@@ -102,8 +102,12 @@
 		    "description": "Keyring for TLS key lookup",
 		    "type": "string"
 		},
+		"tls_key_identity": {
+		    "description": "PSK identity for the TLS PSK (tls_key)",
+		    "type": "string"
+		},
 		"tls_key": {
-		    "description": "TLS key for the connection",
+		    "description": "TLS PSK in PSK interchange format",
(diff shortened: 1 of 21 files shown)


hreinecke:
Please check also https://github.com/linux-nvme/libnvme/pull/896 ; we should probably combine these two pull requests.

igaw:
I found that the config.json and `connect-all` doesn't go well together. `connect-all` wants really to use discovery controllers to figure out what to connect. It uses parts of the info the config.json to create a discovery controller but with tls enabled this really doesn't work against the Linux target implementation.

Maybe `connect` should connect to all controllers listed in config.json without the discovery part. `connect-all` turned into a monster and we have to look at it later.  Anyway, these changes here are working fine, it's just nvme-cli needs to be updated.

igaw:
decided to rip out the retained key handling completely. nice feature which likely no one is going to use.

igaw:
So finally got all use cases working:
- connect using config.json only (no discovery etc, just the entries in the config)
- connect with key pass directly from command line 
- connect with key already in keystore

Note, nvme-cli also needs some changes which I am not sure how to handle in the details. I suggest we discuss this in the nvme-cli PR though.

PR #895, opened 2024-10-17
types: Added enum for SMVES event of PEL log

francispravin5:
Added enum for Sanitize Media Verification Event Support (SMVES) field in Persistent Event Log.


igaw:
Thanks!

PR #896, opened 2024-10-18
Handle TLS keys with HMAC type '0'

hreinecke:
The NVMe spec defines an HMAC type '0' for the PSK interchange format, which indicates that the specified ('configured') key should be used as 'retained' key, ie without any transformation.
This patchset adds handling for these keys.


igaw:
From a quick glance, the first patch can go on top of my PR. The second patch is already covered in my PR. I'd say we first merge my PR and pick the first patch then.

igaw:
I've picked the first patch and put it on top of my PR.

igaw:
I've merged my PR with the first patch in it.

PR #898, opened 2024-10-24
type: Added enums for ANSAN and RGCNS bit of OAES field

arbaz404:
Added enums for ANSAN and RGCNS bit of OAES field based on NVM Express Base Specification 2.1.


igaw:
Thanks!

PR #903, opened 2024-10-31
tests: mi: Add test for dlen/doff values through raw admin_xfer inteface

jk-ozlabs:
Ensure that we're setting the `dlen` and `doff` values correctly, where the semantics of dlen in particular will change on requests vs. responses.

Passes due to the fixes in https://github.com/linux-nvme/libnvme/pull/897


igaw:
Thanks a lot!

PR #907, opened 2024-10-31
build: Add token for coverage test

MaisenbacherD:
The code coverage action is silently failing partly because the required Codecov token is missing.
Reading the token from the GitHub secrets.

The missing gpg package must be fixed in
https://github.com/linux-nvme/ci-containers.

Furthermore, fail the coverage workflow if the codecov-action errors out.

Please add the `CODECOV_TOKEN` to the GitHub secrets :)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,4 +19,5 @@ jobs:
           scripts/build.sh coverage
       - uses: codecov/codecov-action@v4
         with:
-          fail_ci_if_error: false
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: true


MaisenbacherD:
The missing gpg package was fixed in https://github.com/linux-nvme/ci-containers/pull/3

igaw:
I'm trying to figure out to access the nvme-cli group on codedev. Hmm, it could be so easy...

igaw:
I couldn't figure out how to get ownership of the project but it seems I do have some sort of permission. I created a global token let's see how this  works.

MaisenbacherD:
It looks like the global token can't be associated with a repository:
```
error - 2024-10-31 17:21:16,837 -- Commit creating failed: {"message":"Repository not found"}
```
https://github.com/linux-nvme/libnvme/actions/runs/11615796884/job/32347235636
https://github.com/linux-nvme/nvme-cli/actions/runs/11615775763/job/32347164262

There is no obvious argument to specify the codecov repo. 
However, after looking closely at the argument list we might have to specify the `directory` of the `coverage.xml` if that is not found by default. But this should return a different error after finding the repo.
https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments

igaw:
Ah, so the problem is that they changed the permission model. The 'app install step' for the linux-nvme organization has to be done by 'This action must be performed by an organization owner'. I'll talk to Keith :)

PR #911, opened 2024-11-13
ioctl: Live Migration

NateThornton:
Continues efforts from the newly minted TP4159 PCIe Infrastructure for Live Migration specification by adding additional types and ioctl methods.

Depends on completion of #913

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -170,6 +170,12 @@ LIBNVME_1_0 {
 		nvme_default_host;
 		nvme_dev_self_test;
 		nvme_dim_send;
+		nvme_lm_cdq;
+		nvme_lm_track_send;
+		nvme_lm_migration_send;
+		nvme_lm_migration_recv;
+		nvme_lm_set_features_ctrl_data_queue;
+		nvme_lm_get_features_ctrl_data_queue;
 		nvme_directive_recv;
 		nvme_directive_send;
 		nvme_directive_send_id_endir;
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -2149,10 +2149,10 @@ int nvme_dim_send(struct nvme_dim_args *args)
 	__u32 cdw10 = NVME_SET(args->tas, DIM_TAS);
 
 	struct nvme_passthru_cmd  cmd = {
-		.opcode     = nvme_admin_discovery_info_mgmt,
-		.cdw10      = cdw10,
-		.addr       = (__u64)(uintptr_t)args->data,
-		.data_len   = args->data_len,
+		.opcode = nvme_admin_discovery_info_mgmt,
+		.cdw10 = cdw10,
+		.addr = (__u64)(uintptr_t)args->data,
+		.data_len = args->data_len,
 		.timeout_ms = args->timeout,
 	};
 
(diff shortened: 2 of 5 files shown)


src/libnvme.map
```diff
@@ -170,6 +170,12 @@ LIBNVME_1_0 {
 		nvme_default_host;
 		nvme_dev_self_test;
 		nvme_dim_send;
+		nvme_lm_cdq;
+		nvme_lm_track_send;
+		nvme_lm_migration_send;
+		nvme_lm_migration_recv;
+		nvme_lm_set_features_ctrl_data_queue;
+		nvme_lm_get_features_ctrl_data_queue;
```
igaw, line 178: Start a new section LIBNVME_1_12 and move them there.

src/nvme/ioctl.c
```diff
@@ -2149,10 +2149,100 @@ int nvme_dim_send(struct nvme_dim_args *args)
...
 	struct nvme_passthru_cmd  cmd = {
-		.opcode     = nvme_admin_discovery_info_mgmt,
-		.cdw10      = cdw10,
-		.addr       = (__u64)(uintptr_t)args->data,
-		.data_len   = args->data_len,
+		.opcode = nvme_admin_discovery_info_mgmt,
+		.cdw10 = cdw10,
+		.addr = (__u64)(uintptr_t)args->data,
+		.data_len = args->data_len,
+		.timeout_ms = args->timeout,
```
igaw, line 2156: Please avoid unrelated whitespace changes.

src/nvme/types.h
```diff
@@ -9007,4 +9025,324 @@ struct nvme_ns_mgmt_host_sw_specified {
 };
 #endif /* SWIG */
 
+/**
+ * enum nvme_lm_select - Select (SEL): This field specifies the type of management operation to
+ * perform.
```
igaw, line 9030: Does this render the documentation correct? Is the `perform` on the short description? I recenetly learned you can add a new line in the short description, but I think then you need to have a empty line between the header and the first paragraph.

src/nvme/types.h
```diff
@@ -9007,4 +9025,324 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_DELETE_CONTROLLER_DATA_QUEUE	= 1,
+	/* Track Send */
+	NVME_LM_LOG_USER_DATA_CHANGES		= 0,
+	NVME_LM_TRACK_MEMORY_CHANGES		= 1,
+	/* Migration Send */
+	NVME_LM_SUSPEND				= 0,
+	NVME_LM_RESUME				= 1,
+	NVME_LM_SET_CONTROLLER_STATE		= 2,
+	/* Migration Receive */
+	NVME_LM_GET_CONTROLLER_STATE		= 0,
```
igaw, line 9057: I fiend this very surprising that the `select` field from  different commands are mixed into one enum. Do we have this pattern already somewhere else?

src/nvme/types.h
```diff
@@ -9007,4 +9025,324 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_MIGRATION_RECV_CSUIDXP_SHIFT	= 24,
+	NVME_LM_MIGRATION_RECV_CSUIDXP_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CSUUIDI_SHIFT	= 16,
+	NVME_LM_MIGRATION_RECV_CSUUIDI_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CSVI_SHIFT	= 16,
+	NVME_LM_MIGRATION_RECV_CSVI_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED	= 1,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED_SHIFT = 0,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED_MASK = 0x1,
+};
```
igaw, line 9188: same comment as above, this seems to mix the definition from different commands into one enum.

src/nvme/types.h
NateThornton: Is does in the man page. Is there somewhere else I should look to verify?

src/nvme/types.h
NateThornton: All the LM commands share the same cdw10 structure. I thought it better organized in one select block.

Would you rather I have an enum per command, where all fields pertaining to that command are located in one block? Or an enum per command per field?

src/nvme/types.h
igaw: Check the output of `scripts/kernel-doc src/nvme/types.h`

src/nvme/ioctl.c
```diff
@@ -2163,3 +2253,73 @@ int nvme_dim_send(struct nvme_dim_args *args)
...
+		.nsid		= NVME_NSID_NONE,
+		.cdw11		= cdqid,
+		.data		= data,
+		.data_len	= sizeof(*data),
+		.timeout	= NVME_DEFAULT_IOCTL_TIMEOUT,
+		.result		= result,
+	};
+
+	return nvme_get_features(&args);
+}
```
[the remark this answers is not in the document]
NateThornton: Types moved to #913

src/nvme/types.h
igaw: While I understand the motivation to interpret the spec and make it a bit more usable I am not sure if this is a good thing in the long run. We didn't do this for the rest of the spec. Maybe @keithbusch has some input on this.

src/nvme/types.h
NateThornton: Looks OK to me

src/nvme/types.h
```diff
@@ -9007,4 +9025,324 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ */
+struct nvme_lm_io_submission_queue_data {
+	__u64 prp1;
+	__u16 qsize;
+	__u16 qid;
+	__u16 cqid;
+	__u16 attrs;
+	__u16 hp;
+	__u16 tp;
+	__u8  rsvd20[4];
```
igaw, line 9210: could you use the field names as in the spec used, e.g. iosqprp1, ...

src/nvme/types.h
```diff
@@ -9007,4 +9025,324 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_IOSQPRIO_SHIFT	= 1,
+	/* I/O Completion Queue */
+	NVME_LM_IOCQPC_MASK	= 0x1,
+	NVME_LM_IOCQPC_SHIFT	= 0,
+	NVME_LM_IOCQIEN_MASK	= 0x1,
+	NVME_LM_IOCQIEN_SHIFT	= 1,
+	NVME_LM_S0PT_MASK	= 0x1,
+	NVME_LM_S0PT_SHIFT	= 2,
+	NVME_LM_IOCQIV_MASK	= 0xffff,
+	NVME_LM_IOCQIV_SHIFT	= 16,
```
igaw, line 9326: the same comment as above, with prefixing it consistently and matching the names with the spec.

igaw:
Sorry for the late review, just a bit busy. And also sorry for being so anal here. But my point is that we shouldn't be too creative when typing down the spec. Note, we also plan to auto generate all the headers for the next version. In this case we really follow the spec exactly.

PR #913, opened 2024-11-14
types: Live Migration

NateThornton:
Continues efforts from the newly minted TP4159 PCIe Infrastructure for Live Migration specification by adding additional types.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -4460,6 +4460,7 @@ struct nvme_eg_event_aggregate_log {
  * @NVME_FID_SUPPORTED_EFFECTS_SCOPE_ENDGRP:	Endurance Group Scope
  * @NVME_FID_SUPPORTED_EFFECTS_SCOPE_DOMAIN:	Domain Scope
  * @NVME_FID_SUPPORTED_EFFECTS_SCOPE_NSS:	NVM Subsystem Scope
+ * @NVME_FID_SUPPORTED_EFFECTS_CDQSCP:		Controller Data Queue
  */
 enum nvme_fid_supported_effects {
 	NVME_FID_SUPPORTED_EFFECTS_FSUPP	= 1 << 0,
@@ -4476,6 +4477,7 @@ enum nvme_fid_supported_effects {
 	NVME_FID_SUPPORTED_EFFECTS_SCOPE_ENDGRP	= 1 << 3,
 	NVME_FID_SUPPORTED_EFFECTS_SCOPE_DOMAIN	= 1 << 4,
 	NVME_FID_SUPPORTED_EFFECTS_SCOPE_NSS	= 1 << 5,
+	NVME_FID_SUPPORTED_EFFECTS_CDQSCP	= 1 << 6,
 };
 
 /**
@@ -6012,6 +6014,8 @@ struct nvme_host_mem_buf_attrs {
  * @NVME_AER_ERROR:	Error event
  * @NVME_AER_SMART:	SMART / Health Status event
  * @NVME_AER_NOTICE:	Notice event
+ * @NVME_AER_IMMEDIATE:	Immediate
+ * @NVME_AER_ONESHOT:	One-Shot
  * @NVME_AER_CSS:	NVM Command Set Specific events
  * @NVME_AER_VS:	Vendor Specific event
  */
@@ -6019,6 +6023,8 @@ enum nvme_ae_type {
 	NVME_AER_ERROR				= 0,
 	NVME_AER_SMART				= 1,
 	NVME_AER_NOTICE				= 2,
+	NVME_AER_IMMEDIATE			= 3,
+	NVME_AER_ONESHOT			= 4,
 	NVME_AER_CSS				= 6,
 	NVME_AER_VS				= 7,
 };
@@ -7790,6 +7796,11 @@ static inline __u32 nvme_status_equals(int status, enum nvme_status_type type,
  * @nvme_admin_fabric_zoning_lookup:	Fabric Zoning Lookup
  * @nvme_admin_fabric_zoning_send:	Fabric Zoning Send
  * @nvme_admin_dbbuf:			Doorbell Buffer Config
+ * @nvme_admin_track_send:		Track Send
+ * @nvme_admin_track_receive:		Track Receive
+ * @nvme_admin_migration_send:		Migration Send
+ * @nvme_admin_migration_receive:	Migration Receive
+ * @nvme_admin_ctrl_data_queue:		Controller Data Queue
  * @nvme_admin_fabrics:			Fabrics Commands
  * @nvme_admin_format_nvm:		Format NVM
  * @nvme_admin_security_send:		Security Send
@@ -7826,6 +7837,11 @@ enum nvme_admin_opcode {
 	nvme_admin_lockdown		= 0x24,
 	nvme_admin_fabric_zoning_lookup	= 0x25,
 	nvme_admin_fabric_zoning_send	= 0x29,
+	nvme_admin_track_send		= 0x3d,
+	nvme_admin_track_receive	= 0x3e,
+	nvme_admin_migration_send	= 0x41,
+	nvme_admin_migration_receive	= 0x42,
+	nvme_admin_ctrl_data_queue	= 0x45,
 	nvme_admin_dbbuf		= 0x7c,
 	nvme_admin_fabrics		= 0x7f,
 	nvme_admin_format_nvm		= 0x80,
@@ -8022,6 +8038,7 @@ enum nvme_cmd_get_log_lid {
  * @NVME_FEAT_FID_SPINUP_CONTROL:	Spinup Control
  * @NVME_FEAT_FID_FDP:			Flexible Data Placement
  * @NVME_FEAT_FID_FDP_EVENTS:		FDP Events
+ * @NVME_FEAT_FID_CTRL_DATA_QUEUE:	Controller Data Queue
  * @NVME_FEAT_FID_ENH_CTRL_METADATA:	Enhanced Controller Metadata
  * @NVME_FEAT_FID_CTRL_METADATA:	Controller Metadata
  * @NVME_FEAT_FID_NS_METADATA:		Namespace Metadata
@@ -8060,6 +8077,7 @@ enum nvme_features_id {
 	NVME_FEAT_FID_SPINUP_CONTROL				= 0x1a,
 	NVME_FEAT_FID_FDP					= 0x1d,
 	NVME_FEAT_FID_FDP_EVENTS				= 0x1e,
+	NVME_FEAT_FID_CTRL_DATA_QUEUE				= 0x21,
 	NVME_FEAT_FID_ENH_CTRL_METADATA				= 0x7d,
 	NVME_FEAT_FID_CTRL_METADATA				= 0x7e,
 	NVME_FEAT_FID_NS_METADATA				= 0x7f,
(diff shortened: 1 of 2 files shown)


src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ *					controller associated with that User Data Migration Queue
+ *					processing commands.
+ * @NVME_LM_LACT_MASK:			Mask to set Logging Action (LACT)
+ * @NVME_LM_LACT_SHIFT:			Shift to set LACT
+ */
+enum nvme_lm_track_send_fields {
+	NVME_LM_LOG_USER_DATA_CHANGES			= 0,
+	NVME_LM_TRACK_MEMORY_CHANGES			= 1,
+	NVME_LM_TRACK_SEND_SELECT_SHIFT			= 0,
+	NVME_LM_TRACK_SEND_SELECT_MASK			= 16,
```
igaw, line 9088: `0xffff` ?

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ * @NVME_LM_CREATE_CDQ_CDQID_MASK:			Mask to get CNTLID field from Create
+ *							response in completion dword0
+ * @NVME_LM_DELETE_CDQ_CDQID_SHIFT:			Shift amount to set CDQID field for deletion
+ * @NVME_LM_DELETE_CDQ_CDQID_MASK:			Mask to set CDQID field for deletion
+ */
+enum nvme_lm_cdq_fields {
+	NVME_LM_CREATE_CDQ				= 0,
+	NVME_LM_DELETE_CDQ				= 1,
+	NVME_LM_CDQ_SELECT_SHIFT			= 0,
+	NVME_LM_CDQ_SELECT_MASK				= 16,
```
igaw, line 9052: `0xffff`?

igaw: Also please use the `SEL` instead the `Select`. This makes it consistent with the recently added field names (and usually better searchable in the spec).  Also is there a reason you don't add the `MOS` (looking at the 2.1 spec)

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ * @NVME_LM_CREATE_CDQ_CDQID_SHIFT:			Shift amount to get CDQID field from Create
+ *							response in completion dword0
+ * @NVME_LM_CREATE_CDQ_CDQID_MASK:			Mask to get CNTLID field from Create
+ *							response in completion dword0
+ * @NVME_LM_DELETE_CDQ_CDQID_SHIFT:			Shift amount to set CDQID field for deletion
+ * @NVME_LM_DELETE_CDQ_CDQID_MASK:			Mask to set CDQID field for deletion
+ */
+enum nvme_lm_cdq_fields {
+	NVME_LM_CREATE_CDQ				= 0,
+	NVME_LM_DELETE_CDQ				= 1,
```
igaw, line 9050: Use the infix `SEL`, e.g.`NVME_LM_SEL_CREATE_CDQ` . I prefer to keep the naming structure in sync with the spec. Having variations and interpretation is making it really hard to find the spot in the spec.

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_CDQ_SELECT_SHIFT			= 0,
+	NVME_LM_CDQ_SELECT_MASK				= 16,
+	NVME_LM_QUEUE_TYPE_USER_DATA_MIGRATION_QUEUE	= 0,
+	NVME_LM_QUEUE_TYPE_SHIFT			= 16,
+	NVME_LM_QUEUE_TYPE_MASK				= 0xffff,
```
igaw, line 9055: and here I'd say it should match the spec and be prefixed `NVME_LM_QT_...`

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_CREATE_CDQ_PC				= 1,
+	NVME_LM_CREATE_CDQ_CNTLID_SHIFT			= 16,
```
igaw, line 9057: Should this value not be `0`

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ *					User Data Migration Queue into that User Data Migration
+ *					Queue where those user data changes are caused by the
+ *					controller associated with that User Data Migration Queue
+ *					processing commands.
+ * @NVME_LM_LACT_MASK:			Mask to set Logging Action (LACT)
+ * @NVME_LM_LACT_SHIFT:			Shift to set LACT
+ */
+enum nvme_lm_track_send_fields {
+	NVME_LM_LOG_USER_DATA_CHANGES			= 0,
+	NVME_LM_TRACK_MEMORY_CHANGES			= 1,
```
igaw, line 9086: Again what happened to `MOS`?

igaw: `NVME_LM_SEL_...`

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ * @NVME_LM_SEQIND_SHIFT:		Shift amount to set SEQIND
+ * @NVME_LM_SEQIND_MASK:		Mask to set SEQIND
+ * @NVME_LM_MIGRATION_SEND_CSUUIDI_SHIFT: Shift amount to set Controller State UUID Index (CSUUIDI)
+ * @NVME_LM_MIGRATION_SEND_CSUUIDI_MASK: Mask to set CSUUIDI
+ * @NVME_LM_MIGRATION_SEND_CSVI_SHIFT:	Shift amount to set Controller State Version Index (CSVI)
+ * @NVME_LM_MIGRATION_SEND_CSVI_MASK:	Mask to set CSVI
+ */
+enum nvme_lm_migration_send_fields {
+	NVME_LM_SUSPEND				= 0,
+	NVME_LM_RESUME				= 1,
```
igaw, line 9133: `NVME_LM_SEL_...`

igaw: and 2.1 has a `2h` `Set Controller State`

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_MIGRATION_SEND_CNTLID_SHIFT	= 0,
+	NVME_LM_MIGRATION_SEND_CNTLID_MASK	= 0xffff,
+	NVME_LM_MIGRATION_SEND_UIDX_SHIFT	= 0,
+	NVME_LM_MIGRATION_SEND_UIDX_MASK	= 0x7f,
+
+	/* Migration Send - Suspend */
+	NVME_LM_DUDMQ				= 1,
+	NVME_LM_DUDMQ_SHIFT			= 31,
+	NVME_LM_DUDMQ_MASK			= 0x1,
+	NVME_LM_STYPE_NOTIFICATION		= 0,
```
igaw, line 9146: The 2.1 spec calls this `Suspend Notification`

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_STYPE_SUSPEND			= 1,
+	NVME_LM_STYPE_SHIFT			= 16,
+	NVME_LM_STYPE_MASK			= 0xff,
+
```
igaw, line 9150: What about `CNTLID`?

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_SEQIND_NOT_FIRST_NOT_LAST	= 0,
+	NVME_LM_SEQIND_FIRST			= 1,
+	NVME_LM_SEQIND_LAST			= 2,
+	NVME_LM_SEQIND_ENTIRE			= 3,
+	NVME_LM_SEQIND_SHIFT			= 16,
+	NVME_LM_SEQIND_MASK			= 0xf,
+	NVME_LM_MIGRATION_SEND_CSUUIDI_SHIFT	= 24,
+	NVME_LM_MIGRATION_SEND_CSUUIDI_MASK	= 0xff,
+	NVME_LM_MIGRATION_SEND_CSVI_SHIFT	= 16,
+	NVME_LM_MIGRATION_SEND_CSVI_MASK	= 0xff,
```
igaw, line 9161: `CNTLID` is also mentioned in 2.1

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+ *						 completion dword0 of Migration Receive command
+ * @NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED_MASK:  Mask to get Controller Suspended value in
+ *						 completion dword0 of Migration Receive command
+ */
+enum nvme_lm_migration_recv_fields {
+	NVME_LM_GET_CONTROLLER_STATE		= 0,
+	NVME_LM_MIGRATION_RECV_SEL_SHIFT	= 0,
+	NVME_LM_MIGRATION_RECV_SEL_MASK		= 0xff,
+	NVME_LM_MIGRATION_RECV_CNTLID_SHIFT	= 0,
+	NVME_LM_MIGRATION_RECV_CNTLID_MASK	= 0xffff,
```
igaw, line 9191: shouldn't this be grouped withe the `CSUIDXP` and `CSIUUDI` fields?

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	NVME_LM_MIGRATION_RECV_CSUIDXP_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CSUUIDI_SHIFT	= 16,
+	NVME_LM_MIGRATION_RECV_CSUUIDI_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CSVI_SHIFT	= 16,
+	NVME_LM_MIGRATION_RECV_CSVI_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED	= 1,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED_SHIFT = 0,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED_MASK = 0x1,
+};
+
```
igaw, line 9206: One general comment here is that it would be way easier to review if the definition order would match the order in the spec. E.g `CSVI`, `CSUIDXP`, `CSUUIDI`, `CNTLID`.

src/nvme/types.h
```diff
@@ -9007,4 +9025,350 @@ struct nvme_ns_mgmt_host_sw_specified {
...
+	/* Migration Receive - Get Controller State */
+	NVME_LM_MIGRATION_RECV_CSUIDXP_SHIFT	= 24,
+	NVME_LM_MIGRATION_RECV_CSUIDXP_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CSUUIDI_SHIFT	= 16,
+	NVME_LM_MIGRATION_RECV_CSUUIDI_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CSVI_SHIFT	= 16,
+	NVME_LM_MIGRATION_RECV_CSVI_MASK	= 0xff,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED	= 1,
+	NVME_LM_MIGRATION_RECV_CTRL_SUSPENDED_SHIFT = 0,
```
igaw, line 9203: Is `CTRL_SUSPENDED` called `CSUP` in the spec?

NateThornton:
Comments reviewed - thanks for detailed scrutiny. I hope to have the changes ready by end of week. Will resolve with #916 too.

igaw:
Thanks a lot. Yeah, I went down the rabbit hole :)

src/nvme/types.h
NateThornton: This is where I wasn't sure - should they be absolute bit position in the CDW, or relative to the parent field - Management Specific Operation (MOS) in this case.

I'll proceed more closely to the spec - if the fields are presented as relative to MOS then the shift and mask values will reflect that as well.

src/nvme/types.h
NateThornton: For Migration Send, CNTLID is common to all SEL types. See `NVME_LM_MIGRATION_SEND_CNTLID_SHIFT` above

igaw: I think it would make sense to keep it KISS and follow the spec pretty close. The long time plan is to generate these headers directly from the spec and having a clever trick here will make any transition to generated header more painful than it needs to be.

igaw:
Sorry for the late response... (insert my usual busy excuse...)

src/nvme/types.h
NateThornton: Got it. I broke CNTLID out into per SEL-type definitions.

igaw:
Looks good. Sorry that I enforce these rules on the API. I am more relaxed when it comes to the implementation but APIs tend to be PITA if we get them wrong (e.g. see the struct args stuff... something for 2.0) Anyway, thanks for the patience.

PR #916, opened 2024-11-24
types: add NVME_ID_CTRL_NVM_LBAMQF definitions

ikegami-t:
The field values added by NVM command set specification 1.1.


igaw:
It seems @NateThornton and you are working on the same stuff. Just in case you didn't see it check #913

PR #918, opened 2024-11-26
types: add enum for the fields added in TP4160

francispravin5:
Added the enum for MAXWZD and NSZS bits of ONCS field in Identify Controller Data Structure and the enum for NSZ bit in the Write Zeroes command. TP4160 - WZSL Limit Modification.

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -2085,6 +2085,15 @@ enum nvme_id_ctrl_cqes {
  * @NVME_CTRL_ONCS_ALL_FAST_COPY:	If set, then all copy operations for
  *					the Copy command are fast copy
  *					operations.
+ * @NVME_CTRL_ONCS_WRITE_ZEROES_DEALLOCATE: If MAXWZD bit set, then the maximum data
+ *					size for Write Zeroes command depends on the
+ *					value of the Deallocate bit in the Write Zeroes
+ *					command and the value in the WZDSL field in the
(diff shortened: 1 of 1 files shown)


src/nvme/types.h
```diff
@@ -2085,6 +2085,15 @@ enum nvme_id_ctrl_cqes {
...
  *					operations.
+ * @NVME_CTRL_ONCS_WRITE_ZEROES_DEALLOCATE: If set, then the maximum data size for
+ *					Write Zeroes command depends on the value of
+ *					the Deallocate bit in the Write Zeroes command
+ *					and the value in the WZDSL field in the I/O
+ *					Command Set specific Identify Controller data
+ *					structure.
+ * @NVME_CTRL_ONCS_NAMESPACE_ZEROES:	If set, then the controller supports the
+ *					Namespace Zeroes (NSZ) bit in the NVM Command
+ *					Set Write Zeroes command.
```
igaw, line 2096: could you add the 'unique' field name here in the comments, e.g. MAXWZD  and  NSZS. The ship has already sailed for the enum names, so we use the long version. But generally I'd prefer to use the short ones in the enums and have the long ones in the documenetion. This short ones are pretty good to use as key to find the right spot in the spec ;)

francispravin5: @igaw Addressed the review comment.
Thanks!

igaw:
Thanks!

PR #919, opened 2024-11-26
test: Add library dependency for dlsym

sc108-lee:
Fix build failure for test

[27/79] Linking target test/ioctl/libmock-ioctl.so FAILED: test/ioctl/libmock-ioctl.so
cc  -o test/ioctl/libmock-ioctl.so test/ioctl/libmock-ioctl.so.p/mock.c.o test/ioctl/libmock-ioctl.so.p/util.c.o -Wl,--as-needed -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libmock-ioctl.so -Wl,--end-group test/ioctl/libmock-ioctl.so.p/mock.c.o: In function `ioctl': /root/Desktop/repo/libnvme_upstream/.build/../test/ioctl/mock.c:154: undefined reference to `dlsym'

### DIFF
--- a/test/ioctl/meson.build
+++ b/test/ioctl/meson.build
@@ -1,6 +1,8 @@
+dl_dep = dependency('dl')
 mock_ioctl = library(
     'mock-ioctl',
     ['mock.c', 'util.c'],
+    dependencies: [dl_dep],
 )
 
 # Add mock-ioctl to the LD_PRELOAD path so it overrides libc.


igaw:
Ooops, though have to look into why the CI build fails horrible.

igaw:
Which version of meson are you using? IIRC some libraries are added to the dependency list automatically and the dl lib might one of them. That might be the reason why it was failing in CI and on my machine.

igaw:
I've merged my fix from https://github.com/linux-nvme/libnvme/pull/923 which works around the problem when dl is not available. Thanks for the report!

PR #922, opened 2024-11-29
ioctl,types: telemetry log update based NVMe v2.1

sc108-lee:
MCDA added for Telemetry Host-Initiated Log creation

Log member variable added
Host-Initiated Log - THS
Controller-Initiated Log - TCS

### DIFF
--- a/src/nvme/ioctl.h
+++ b/src/nvme/ioctl.h
@@ -1526,14 +1526,32 @@ static inline int nvme_get_log_device_self_test(int fd,
 }
 
 /**
- * nvme_get_log_create_telemetry_host() - Create host telemetry log
+ * enum nvme_telemetry_da - Telemetry Log Data Area
+ * @NVME_TELEMETRY_DA_CTRL_DETERMINE:
+ * @NVME_TELEMETRY_DA_1:		Data Area 1
+ * @NVME_TELEMETRY_DA_2:		Data Area 2
+ * @NVME_TELEMETRY_DA_3:		Data Area 3
+ * @NVME_TELEMETRY_DA_4:		Data Area 4
+ */
+enum nvme_telemetry_da {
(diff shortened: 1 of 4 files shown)


src/nvme/ioctl.h
```diff
@@ -1548,7 +1564,7 @@ static inline int nvme_get_log_create_telemetry_host(int fd,
 		.nsid = NVME_NSID_NONE,
 		.csi = NVME_CSI_NVM,
 		.lsi = NVME_LOG_LSI_NONE,
-		.lsp = NVME_LOG_TELEM_HOST_LSP_CREATE,
+		.lsp = (__u8)((mcda << 1) | NVME_LOG_TELEM_HOST_LSP_CREATE),
```
igaw, line 1567: I assume this is the thing which `5.1.12.1.8 Telemetry Host-Initiated (Log Page Identifier 07h)` talks about:

```
If the MCDAS bit (refer to Figure 215) is set to ‘1’ and a Get Log Page command with the Create Telemetry
Host-Initiated Data bit is set to ‘1’, then the maximum data area to be created in the Telemetry Host-Initiated
log page shall be less than or equal to the MCDA field in the Log Specific Parameter field in Command
Dword 10 (refer to Figure 213).
```

Is my understanding correct?

igaw:
looks good, just a bit confused how this api works.

src/nvme/ioctl.h
sc108-lee: yes it is

sc108-lee:
Actually I was considering api problem, same as you.
So I updated the commit as I first modified.

remain : old function to call new function
new : nvme_get_log_create_telemetry_host_mcda include mcda argument.

igaw:
Thanks!

PR #923, opened 2024-12-02
tests: do not unconditionally depend on dlsym

igaw:
Older versions of glibc do not have dl as builtin in. Thus include libdl when it's available.

While at it also make the mock library to depend hard on the presents of dlsym. It's only necessary for s390.

Fixes: cbbfdd0ddfa2 ("test/mock: pass thru unknown ioctls")
Reported-by: Steven Seungcheol Lee <sc108.lee@samsung.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -246,6 +246,12 @@ conf.set(
     ),
     description: 'Is ioctl the glibc interface (rather than POSIX)'
 )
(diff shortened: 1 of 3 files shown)


sc108-lee:
sorry for late reply, one of my pc that use older version has meson version 1.2.0 :)
nice fix!

PR #924, opened 2024-12-03
mi: telemetry log update based NVMe v2.1

sc108-lee:
MCDA added for Telemetry Host-Initiated Log creation

sorry for missing mi api, Just updated :)


igaw:
I didn't catch it either :)

PR #925, opened 2024-12-06
types: use __le16 instead of __u16 for data fields

sc108-lee:
NVMe specification specifies data in a little endian format


igaw:
Thanks! 

This is likely to cause some fallouts, but it's just during compile time.

PR #926, opened 2024-12-06
ioctl: Live Migration

NateThornton:
Continues efforts from the newly minted TP4159 PCIe Infrastructure for Live Migration specification with new ioctls.

Same as #911 but adopts new types.h definitions from #913 and removes MOS derived arguments in favor of generic `mos` field. I figured this was more "future proof".

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,4 +1,14 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
+LIBNVME_1.12 {
+	global:
+		nvme_lm_cdq;
+		nvme_lm_track_send;
+		nvme_lm_migration_send;
+		nvme_lm_migration_recv;
+		nvme_lm_set_features_ctrl_data_queue;
+		nvme_lm_get_features_ctrl_data_queue;
+};
+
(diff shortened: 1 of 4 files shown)


src/libnvme.map
```diff
@@ -170,6 +180,12 @@ LIBNVME_1_0 {
 		nvme_default_host;
 		nvme_dev_self_test;
 		nvme_dim_send;
+		nvme_lm_cdq;
+		nvme_lm_track_send;
+		nvme_lm_migration_send;
+		nvme_lm_migration_recv;
+		nvme_lm_set_features_ctrl_data_queue;
+		nvme_lm_get_features_ctrl_data_queue;
```
igaw, line 188: These symbols are already in the 1.12 section. A bit too much copy pasta? :)

igaw:
Rest looks good. Good we haven't released the `struct args` changes yet, so there is no problem with changing them now.

NateThornton:
Appreciate your careful eye and thoughtful feedback through all of this! The changes are much better than my original set.

The nvme-cli changes are next.

PR #927, opened 2024-12-07
ioctl: get_log_page by nvme uring cmd

996refuse:
hello libnvme master,

io_uring on linux is pretty mature now, for the system with liburing, do the batch nvme_get_log by nvme uring cmd.
This change could gain about 10% performance improvement for some large log pages on my lab.

i believe the performance can be better if implement producer and consumer in two separated threads, if it's worth and acceptable, i can spend some time on this later.

thanks

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -8,6 +8,9 @@
  */
 #include <errno.h>
 #include <fcntl.h>
+#ifdef CONFIG_LIBURING
+#include <liburing.h>
+#endif
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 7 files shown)


src/nvme/ioctl.c
```diff
@@ -8,6 +8,7 @@
  */
 #include <errno.h>
 #include <fcntl.h>
+#include <liburing.h>
```
igaw, line 11: This include needs to be guarded, no?

src/nvme/ioctl.c
```diff
@@ -344,14 +431,30 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
 		args->len = xfer;
 		args->log = ptr;
 		args->rae = offset + xfer < data_len || retain;
+#ifdef CONFIG_LIBURING
+		if (n >= NVME_URING_ENTRIES) {
+			nvme_uring_cmd_wait_complete(&ring, n);
```
igaw, line 436: I am fine with using uring but would it be possible to have it fallback during runtime to the ioctl when uring is not available, like on older kernels?

src/nvme/ioctl.c
```diff
@@ -323,6 +323,11 @@ int nvme_get_log(struct nvme_get_log_args *args)
 #ifdef CONFIG_LIBURING
 static int nvme_uring_cmd_setup(struct io_uring *ring)
 {
+	struct io_uring_probe *probe = io_uring_get_probe();
+	if (!io_uring_opcode_supported(probe, IORING_OP_URING_CMD)) {
+		perror("IORING_OP_URING_CMD is not supported for the kernel version below 5.19, failback ioctl interface.");
+		return -1;
```
igaw, line 329: Could we hook this up on library load time, so it only tests once if the io_uring is available?

Also it should not print an error at all, this would spam the logs when io_uring is not available. Internal function shouldn't use the POSIX style of error propagation (setting errno and return -1). The plan is to fix this once we work on the next major version.

src/nvme/ioctl.c
```diff
@@ -407,14 +412,13 @@ int nvme_get_log_page(int fd, __u32 xfer_len, struct nvme_get_log_args *args)
 
 	args->fd = fd;
 
+	bool failback = true;
```
igaw, line 415: Looks like a typo, `fallback` ?

src/nvme/ioctl.c
996refuse: thanks for your clarification.

> Could we hook this up on library load time, so it only tests once if the io_uring is available?

correct me if i'm wrong, i think it's not possible to catch the error on .so library loading time in an ordinary way.
`nvme: error while loading shared libraries: liburing.so.2: cannot open shared object file: No such file or directory`
ld.so looks for the .so from the ELF dynamic section then load the .so, then run the ELF .text code.
when the error is occurs, the program has not started yet.
we can hack to load the .so on program runtime, but it's not recommended.

what if add a parameter for get-log to give an explicit instruction to use io_uring? and by default, this feature is not enabled, what do you think?
`nvme get-log /dev/nvme2 --log-id=0x00 --log-len=0x1000 --uring-cmd`

igaw: Indeed, my idea doesn't work. We don't have a `nvme_init_lib` function yet. So the simplest thing is to introduce a global static variable which tracks if io_uring is available. That doesn't need to be race free, so no locks or so are necessary.  Worst case if two threads are figuring out that io_uring is available or not at the same time and both should come to the same conclusion.

996refuse: what the two threads are, could help explain it please? two threads do same task is inefficient, increasing the libnvme runtime complexity.

maybe we too overcomplicated the thing. by default, compile time `disabled` the feature, remove liburing.so and io_uring syscall dependencies. waiting for liburing being accepted by main linux distributions widely, then switch it to `auto`.

igaw: Just leave the meson auto detection in place. This part is fine.

No, it's very simple. If a library is loaded into process context, the application could run several threads using libnvme. This is perfectly fine, e.g. if there are multiple devices. So we don't want to have the detection everytime for each single command send out. So this part just needs to be done once. The io_uring detection though can't be made 'thread safe' without using global locks which is not really a good idea. So my point, it doesn't matter. We just need [eventually consistency](https://en.wikipedia.org/wiki/Eventual_consistency). 

With this in mind, something like this is good enough:

```c
enum io_uring_state {
  io_uring_state_detect, 
  io_uring_state_available,
  io_uring_state_not_available,
};
enum io_uring_state io_uring_state = io_uring_state_detect;


if (io_uring_state == io_uring_detetct) {
  if (io_uring_opcode_supported(probe, IORING_OP_URING_CMD))
    io_uring_state =  io_uring_available;
  else
    io_uring_state = io_uring_not_available;
}
```

996refuse: brilliant idea, got it

996refuse: i prefer `constructor`, which is a gcc specific attribute, the function attached is called when libnvme.so is loaded.

i tested the fallback mechanism manually by io_uring_disabled sysctl, it works well
`sysctl -w kernel.io_uring_disabled=2`

igaw:
This should be just one commit. I'll squash the changes into one when I am reviewing and testing it. We need also to update our CI build accordingly.

996refuse:
thanks for your help !

there're 2 failed testcase. it requires mock liburing api
```
 8/34 ana                                              FAIL            0.01s   killed by signal 6 SIGABRT
10/34 discovery                                        FAIL            0.01s   killed by signal 6 SIGABRT
```
besides, it doesn't work on SPDK NVME CUSE devices. FYI

igaw:
> there're 2 failed testcase. it requires mock liburing api

Ah, I've missed this. Unfortunately, this is important, I really don't want the tests cases to break when liburing is used.

> besides, it doesn't work on SPDK NVME CUSE devices. FYI

That is likely due the missing support of the CUSE devices.

igaw:
I suppose, for an initial attempt the `io_uring_opcode_supported(probe, IORING_OP_URING_CMD)` should just fail. 

`io_uring_opcode_supported` calls `__sys_io_uring_register`, this means you need to mock the `io_uring_register`:

https://github.com/axboe/liburing/blob/6aeac2041987649c32a8f15de104ec4339725c49/src/arch/syscall-defs.h#L64C1-L69C2

```c
static inline int __sys_io_uring_register(unsigned int fd, unsigned int opcode,
					  const void *arg, unsigned int nr_args)
{
	return (int) __do_syscall4(__NR_io_uring_register, fd, opcode, arg,
				   nr_args);
}
```

I haven't looked into liburing at all, maybe it supports some sort of mocking for testing?

996refuse:
no, io_uring_get_probe returning NULL could avoid breaking the cases perfectly. after a quick study, think there are no official mocking support of liburing.

igaw:
Just squashed the changes into one patch. Thanks!

PR #929, opened 2024-12-13
Update mctp APIs to mctp codeconstruct version 2.0

ThuBaNguyen:
The mctp codeconstruct released the mctp APIs version 2.0 as [1]. Update code to use new APIs.

[1] https://github.com/CodeConstruct/mctp/blob/v2.0/docs/mctpd.md

### DIFF
--- a/examples/mi-conf.c
+++ b/examples/mi-conf.c
@@ -25,9 +25,9 @@
 
 #include <dbus/dbus.h>
 
-#define MCTP_DBUS_NAME "xyz.openbmc_project.MCTP"
-#define MCTP_DBUS_PATH "/xyz/openbmc_project/mctp"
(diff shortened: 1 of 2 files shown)


jk-ozlabs:
LGTM, thanks!

As far as I am aware, there is minimal external usage of the dbus-based device enumeration at present, so a changeover should have minimal impact. Might be worth a mention in the libnvme changelog though.

ThuBaNguyen:
> LGTM, thanks!
> 
> As far as I am aware, there is minimal external usage of the dbus-based device enumeration at present, so a changeover should have minimal impact. 
> Might be worth a mention in the libnvme changelog though.
What change log file do you mean?

jk-ozlabs:
That was more a note if @igaw does any release notes / related news on releases.

igaw:
Thanks!

PR #933, opened 2024-12-25
types: add NVMe 2.1 identify power state descriptor fields

ikegami-t:


src/nvme/types.h
```diff
@@ -1162,7 +1168,12 @@ struct nvme_id_psd {
 	__u8			rsvd19;
 	__le16			actp;
 	__u8			apws;
-	__u8			rsvd23[9];
+	__u8			epfrt;
+	__u8			fqvt;
+	__u8			epfvt;
+	__u8			epfr_fqv_ts;
+	__u8			epfvts;
```
igaw, line 1175: The official released spec 2.1 doesn't list the last two fields in Figure 313. Instead there is

- epfrts
- fqvts
- epfvts

Are you sure you got the official released spec?

ikegami-t: Yes those three fields are 4 btis size field so the __u8 epfr_fqv_ts variable combined epfrts and fqvts fields. (The exsiting __u8 apws variable is also combined 4 bits fields APS and APW as same.) Note: The epfvts field also 4 bits size filed but remaining 4 bits of the __u8 epfvts variable are reseved bit.

igaw: Ufff, that is ugly! Very inconsistently done by the spec. Alright, in this case you solution looks reasonable. Sorry, I haven't seen your documentation update, which says it.

igaw:
Thanks!

PR #935, opened 2024-12-31
types: error information log entry status field definitions

ikegami-t:
This is to use libnvme register definitions instead of the shift operator.


igaw:
Thanks!

PR #936, opened 2025-01-04
ioctl: add nvme_get_egid_log()

ikegami-t:
This is to use for the endurance group scope log page.

### DIFF
--- a/src/nvme/ioctl.h
+++ b/src/nvme/ioctl.h
@@ -2136,6 +2136,24 @@ static inline int nvme_get_log_rotational_media_info(int fd, __u16 endgid, __u32
 }
 
 /**
+ * nvme_get_log_dispersed_ns_participating_nss() - Retrieve Dispersed Namespace Participating NVM
+ * Subsystems Log
+ * @fd:		File descriptor of nvme device
+ * @nsid:	Namespace Identifier
+ * @len:	The allocated length of the log page
+ * @log:	User address to store the log page
+ *
+ * Return: The nvme command status if a response was received (see
+ * &enum nvme_status_field) or -1 with errno set otherwise
+ */
+static inline int nvme_get_log_dispersed_ns_participating_nss(int fd, __u32 nsid, __u32 len,
+	struct nvme_dispersed_ns_participating_nss_log *log)
+{
+	return nvme_get_nsid_log(fd, false, NVME_LOG_LID_DISPERSED_NS_PARTICIPATING_NSS, nsid, len,
+				 log);
+}
+
+/**
  * nvme_get_log_phy_rx_eom() - Retrieve Physical Interface Receiver Eye Opening Measurement Log
  * @fd:		File descriptor of nvme device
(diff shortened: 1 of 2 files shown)


src/nvme/ioctl.h
```diff
@@ -1370,6 +1370,22 @@ static inline int nvme_get_nsid_log(int fd, bool rae,
...
+				    __u32 len, void *log)
+{
+	struct nvme_get_log_args args = {
+		.log = log,
+		.args_size = sizeof(args),
+		.fd = fd,
+		.lid = lid,
+		.len = len,
+		.lsi = endgid,
+		.rae = rae,
```
igaw, line 1383: You should probably also set the timeout value, no?

src/nvme/ioctl.h
```diff
@@ -2195,6 +2195,34 @@ static inline int nvme_get_log_phy_rx_eom(int fd, __u8 lsp, __u16 controller,
...
+{
+	struct nvme_get_log_args args = {
+		.log = log,
+		.args_size = sizeof(args),
+		.fd = fd,
+		.lid = NVME_LOG_LID_REACHABILITY_GROUPS,
+		.len = len,
+		.nsid = NVME_NSID_ALL,
+		.lsp = rgo,
+		.rae = rae,
```
igaw, line 2220: Timeout missing?

src/nvme/ioctl.h
```diff
@@ -2223,6 +2223,34 @@ static inline int nvme_get_log_reachability_groups(int fd, __u32 len, bool rgo,
...
+{
+	struct nvme_get_log_args args = {
+		.log = log,
+		.args_size = sizeof(args),
+		.fd = fd,
+		.lid = NVME_LOG_LID_REACHABILITY_ASSOCIATIONS,
+		.len = len,
+		.nsid = NVME_NSID_ALL,
+		.lsp = rao,
+		.rae = rae,
```
igaw, line 2248: timeout?

src/nvme/ioctl.h
```diff
@@ -2305,6 +2305,34 @@ static inline int nvme_get_log_discovery(int fd, bool rae,
...
+					     struct nvme_host_discover_log *log)
+{
+	struct nvme_get_log_args args = {
+		.log = log,
+		.args_size = sizeof(args),
+		.fd = fd,
+		.lid = NVME_LOG_LID_HOST_DISCOVER,
+		.len = len,
+		.nsid = NVME_NSID_ALL,
+		.lsp = allhoste,
```
igaw, line 2329: timeout?

igaw:
Looks all good. Just my comment on the timeout.

ikegami-t:
The default timeout value is zero so not needed to set as initialized the variable args timeout as zero. Should we set the default value explicitly? Thank you.

ikegami-t:
Fixed the patch to set the timeout and other values unset before. I just understood again as before you mentioned for the nvme-cli commands config values as same. Thank you.

igaw:
Thanks!

PR #939, opened 2025-01-06
mi_xfer: Added nvme_mi_mi_xfer API

chorkin:
Exposed an nvme_mi_mi_xfer for MI messages.  This is similar to the Admin passthrough, nvme_mi_admin_xfer.  

Added new API and test cases in test-mi
Signed-off-by: Chuck Horkin <chorkin@microsoft.com>

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -1,3 +1,8 @@
+LIBNVME_MI_1_12 {
+	global:
+		nvme_mi_mi_xfer;
+};
+
 LIBNVME_MI_1_11 {
 	global:
 		nvme_mi_control;
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -1599,6 +1599,72 @@ static int nvme_mi_read_data(nvme_mi_ep_t ep, __u32 cdw0,
 	return 0;
 }
 
+int nvme_mi_mi_xfer(nvme_mi_ep_t ep,
+		       struct nvme_mi_mi_req_hdr *mi_req,
+		       size_t req_data_size,
+		       struct nvme_mi_mi_resp_hdr *mi_resp,
+		       size_t *resp_data_size)
+{
+	int rc;
+	struct nvme_mi_req req;
+	struct nvme_mi_resp resp;
+
+	/* There is nothing in the spec to define this limit but going with the limits
+	 * from the admin message types for DLEN seems like a reasonable starting point
+	 * to check for coding errors
+	 */
+	const size_t mi_data_xfer_size_limit = 4096;
+
+	/* length/offset checks. The common _submit() API will do further
+	 * checking on the message lengths too, so these are kept specific
+	 * to the requirements of the particular command set
+	 */
+
+	if (*resp_data_size > mi_data_xfer_size_limit) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	/* request and response lengths & offset must be aligned */
+	if ((req_data_size & 0x3) ||
+	    (*resp_data_size & 0x3)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	/* bidirectional not permitted */
+	if (req_data_size && *resp_data_size) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	mi_req->hdr.type = NVME_MI_MSGTYPE_NVME;
+	mi_req->hdr.nmp = (NVME_MI_ROR_REQ << 7) |
+				(NVME_MI_MT_MI << 3);
+
+	memset(&req, 0, sizeof(req));
+	req.hdr = &mi_req->hdr;
+	req.hdr_len = sizeof(*mi_req);
+	req.data = mi_req + 1;
+	req.data_len = req_data_size;
+
+	nvme_mi_calc_req_mic(&req);
+
+	memset(&resp, 0, sizeof(resp));
+	resp.hdr = &mi_resp->hdr;
+	resp.hdr_len = sizeof(*mi_resp);
+	resp.data = mi_resp + 1;
+	resp.data_len = *resp_data_size;
+
+	rc = nvme_mi_submit(ep, &req, &resp);
+	if (rc)
+		return rc;
+
+	*resp_data_size = resp.data_len;
+
+	return 0;
+}
+
 int nvme_mi_mi_read_mi_data_subsys(nvme_mi_ep_t ep,
 				   struct nvme_mi_read_nvm_ss_info *s)
 {
--- a/src/nvme/mi.h
+++ b/src/nvme/mi.h
@@ -730,6 +730,36 @@ char *nvme_mi_endpoint_desc(nvme_mi_ep_t ep);
 /* MI Command API: nvme_mi_mi_ prefix */
 
 /**
+ * nvme_mi_mi_xfer() -  Raw mi transfer interface.
+ * @ep: endpoint to send the MI command to
+ * @mi_req: request data
+ * @req_data_size: size of request data payload
+ * @mi_resp: buffer for response data
+ * @resp_data_size: size of response data buffer, updated to received size
+ *
+ * Performs an arbitrary NVMe MI command, using the provided request data,
+ * in @mi_req. The size of the request data *payload* is specified in
+ * @req_data_size - this does not include the standard header length (so a
+ * header-only request would have a size of 0). Note that the Management
+ * Request Doublewords are considered part of the header data.
+ *
+ * On success, response data is stored in @mi_resp, which has an optional
+ * appended payload buffer of @resp_data_size bytes. The actual payload
+ * size transferred will be stored in @resp_data_size. This size does not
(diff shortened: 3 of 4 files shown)


igaw:
@jk-ozlabs could you have a look? Thanks!

src/nvme/mi.c
```diff
@@ -1599,6 +1599,67 @@ static int nvme_mi_read_data(nvme_mi_ep_t ep, __u32 cdw0,
...
+	struct nvme_mi_req req;
+	struct nvme_mi_resp resp;
+
+	/* length/offset checks. The common _submit() API will do further
+	 * checking on the message lengths too, so these are kept specific
+	 * to the requirements of the Admin command set
+	 */
+
+	/* NVMe-MI v1.2 imposes a limit of 4096 bytes on the dlen field */
+	if (*resp_data_size > 4096) {
```
jk-ozlabs, line 1618: The DLEN field is not present on MI commands. It's sensible to have a limit here, but is 4096 still appropriate?

src/nvme/mi.h
```diff
@@ -729,6 +729,36 @@ char *nvme_mi_endpoint_desc(nvme_mi_ep_t ep);
...
+ * @ep: endpoint to send the admin command to
+ * @mi_req: request data
+ * @req_data_size: size of request data payload
+ * @mi_resp: buffer for response data
+ * @resp_data_size: size of response data buffer, updated to received size
+ *
+ * Performs an arbitrary NVMe MI command, using the provided request data,
+ * in @admin_req. The size of the request data *payload* is specified in
+ * @req_data_size - this does not include the standard header length (so a
+ * header-only request would have a size of 0).
```
jk-ozlabs, line 743: Would be good to clarify here whether the standard NVME Management Request Dwords are included in this size (ie, that they are considered part of the header, not the request data).

src/nvme/mi.h
```diff
@@ -729,6 +729,36 @@ char *nvme_mi_endpoint_desc(nvme_mi_ep_t ep);
...
+ *
+ * On success, response data is stored in @admin_resp, which has an optional
+ * appended payload buffer of @resp_data_size bytes. The actual payload
+ * transferred will be stored in @resp_data_size. These sizes do not include
+ * the Admin request header, so 0 represents no payload.
```
jk-ozlabs, line 748: This paragraph seems to be copied from the admin_xfer comment; could probably use some correcting for the mi xfer implemetation.

test/mi.c
```diff
@@ -629,6 +629,44 @@ static void test_admin_invalid_formats(nvme_mi_ep_t ep)
...
+	struct {
+		struct nvme_mi_mi_req_hdr hdr;
+		uint8_t data[4];
+	} req = { 0 };
+	struct nvme_mi_mi_resp_hdr resp = { 0 };
+	nvme_mi_ctrl_t ctrl;
+	size_t len;
+	int rc;
+
+	test_set_transport_callback(ep, test_admin_invalid_formats_cb, NULL);
```
jk-ozlabs, line 643: Might be good to rename the callback if we're no longer just using it for admin.

Maybe something like `test_rejected_command_cb` ?

src/nvme/mi.c
```diff
@@ -1599,6 +1599,67 @@ static int nvme_mi_read_data(nvme_mi_ep_t ep, __u32 cdw0,
...
+	}
+
+	/* request and response lengths & offset must be aligned */
+	if ((req_data_size & 0x3) ||
+	    (*resp_data_size & 0x3)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	/* bidirectional not permitted (see DLEN definition) */
```
jk-ozlabs, line 1630: The DLEN definition doesn't apply here, but I suspect we cannot perform bidirectional commands anyway. There doesn't seem to be any indication in the spec about this, have you seen a reference somewhere?

src/nvme/mi.c
```diff
@@ -1599,6 +1599,67 @@ static int nvme_mi_read_data(nvme_mi_ep_t ep, __u32 cdw0,
...
+		       struct nvme_mi_mi_resp_hdr *mi_resp,
+		       size_t *resp_data_size)
+{
+	int rc;
+	struct nvme_mi_req req;
+	struct nvme_mi_resp resp;
+
+	/* length/offset checks. The common _submit() API will do further
+	 * checking on the message lengths too, so these are kept specific
+	 * to the requirements of the Admin command set
```
jk-ozlabs, line 1614: There are no Admin command set requirements applicable here :)

jk-ozlabs (commented):
I'm fine with the concept here, but could you elaborate on what the intended use-case is here? If it's just for coverage of specified MI commands, then we may want to add those (rather than leaking the implementations out the caller). Or is this for vendor-defined MI interactions?

It may be helpful to include a bit of a rationale/use-case in the commit message, if that's applicable - your call on that.

I have a couple of comments inline. You'll also need to update the .map file, otherwise the new function will not be visible.

src/nvme/mi.c
chorkin: Yes.  I realized this.  What do you think a good limit is?  I didn't see anything documented so I just reused this.

src/nvme/mi.c
chorkin: I have not.  Shall I remove this check for now?

chorkin:
> I'm fine with the concept here, but could you elaborate on what the intended use-case is here? If it's just for coverage of specified MI commands, then we may want to add those (rather than leaking the implementations out the caller). Or is this for vendor-defined MI interactions?
> 
> It may be helpful to include a bit of a rationale/use-case in the commit message, if that's applicable - your call on that.
> 
> I have a couple of comments inline. You'll also need to update the .map file, otherwise the new function will not be visible.

Thanks for the review.  Regarding the use-case, we have a service we're writing and it's meant to be mostly pass through to allow multiple processes to talk to the device we are working on.  The device will also have some custom vendor-defined MI interactions.  For both of those reasons I wanted to expose this pass through API.

Noted on the .map.  I will fix this.

src/nvme/mi.c
jk-ozlabs: My thinking is that there are facilities that wouldn't work with a bidirectional MI command (eg, the in-band MI tunneling), so let's reject those for now. Always better for a future change to make things more permissive than less - so we don't break things that used to work.

So, I'd say leave the check in, but remove the DLEN reference in the comment.

src/nvme/mi.c
jk-ozlabs: I can't see any limits mentioned in the spec either, and there is no length value encoded anywhere (which would impose a programmatic limit). 4096 is probably reasonable. I'd suggest making this a `const size_t` somewhere, with a bit of a comment on rationale.

src/nvme/mi.h
chorkin: Can you be more specific.  To me it reads like this is already included "this does not include the standard header length (so a header-only request would have a size of 0)"

src/nvme/mi.h
chorkin: Cleaned up

src/nvme/mi.c
chorkin: Sounds good.

src/nvme/mi.c
chorkin: This is a weird one, given we have no hard requirement on the length checks.  Let me know if you're okay with my update here.

src/libnvme-mi.map
```diff
@@ -1,6 +1,7 @@
 LIBNVME_MI_1_11 {
 	global:
 		nvme_mi_control;
+		nvme_mi_mi_xfer;
```
igaw, line 4: 1.11 is already released. You need to open up a new section `LIBNVME__MI_1_12`.

src/nvme/mi.c
```diff
@@ -1599,6 +1599,72 @@ static int nvme_mi_read_data(nvme_mi_ep_t ep, __u32 cdw0,
...
+		       struct nvme_mi_mi_resp_hdr *mi_resp,
+		       size_t *resp_data_size)
+{
+	int rc;
+	struct nvme_mi_req req;
+	struct nvme_mi_resp resp;
+
+	/* There is nothing in the spec to define this limit but going with the limits
+	 * from the admin message types for DLEN seems like a reasonable starting point
+	 * to check for erronious coding errors
```
jk-ozlabs, line 1614: erronious errors! :D

jk-ozlabs:
We're at the stage of me doing minor tweaks in comments, so all looks pretty good to me with the .map fix that @igaw has mentioned.

src/nvme/mi.c
jk-ozlabs: yeah, looks good to me.

src/libnvme-mi.map
chorkin: Done

jk-ozlabs (approved):

PR #940, opened 2025-01-07
types: add NVME_FEAT_FDPE_*** definitions

ikegami-t:
Also mark struct nvme_feat_fdp_events_cwd11 as deprecated instead.


igaw:
Thanks!

PR #942, opened 2025-01-10
compile: avoid min macro pollution

zhangjian3032:
The generic min macro in util.h is too common and can conflict with other libraries. Rename it to nvme_min to avoid pollution.

Before renaming the macro, the cpp's std::min will fail to compile.
```
./test/cpp.cc: In function ‘int min_compile_test()’:
../src/nvme/util.h:563:19: error: expected unqualified-id before ‘(’ token
  563 | #define min(x, y) ((x) > (y) ? (y) : (x))
```


src/nvme/util.h
```diff
@@ -560,7 +560,7 @@ char *kv_keymatch(const char *kv, const char *key);
  */
 char *startswith(const char *s, const char *prefix);
 
-#define min(x, y) ((x) > (y) ? (y) : (x))
+#define nvme_min(x, y) ((x) > (y) ? (y) : (x))
```
igaw, line 563: I think we should not expose this macro at all. And there should be already the min macro from the ccan 'library'. Can you drop this one and include the ccan/minmax.h header in the c files?

test/cpp.cc
```diff
@@ -6,9 +6,20 @@
...
 
+static int min_compile_test()
+{
+	return std::min(1, 2);
+}
+
+static int max_compile_test()
+{
+	return std::max(1, 2);
+}
```
igaw, line 21: This is unrelated with the above change.

zhangjian3032: yup, I think min and max should always appear in pairs, so I just added it, remove now.

src/nvme/util.h
zhangjian3032: yup, we can, but we need cast the type, see updated code.

src/nvme/ioctl.c
```diff
@@ -503,7 +504,7 @@ static int read_ana_chunk(int fd, enum nvme_log_ana_lsp lsp, bool rae,
 	}
 
 	while (*read < to_read) {
-		__u32 len = min(log_end - *read, NVME_LOG_PAGE_PDU_SIZE);
+		__u32 len = min((int)(log_end - *read), NVME_LOG_PAGE_PDU_SIZE);
```
igaw, line 507: I think it would be better to use `min_t` here:

    `min_t(int, log_end - *read, NVME_LOG_PAGE_PDU_SIZE);`

test/cpp.cc
```diff
@@ -6,9 +6,15 @@
...
  */
 
+#include <algorithm>
 #include <iostream>
 #include <libnvme.h>
 
+static int min_compile_test()
+{
+	return std::min(1, 2);
+}
```
igaw, line 16: My point is that the whole change to `test/cpp.c` should be in a different patch. I don't mind if you add the `min` and `max` to catch build errors, just in a separate patch.

src/nvme/ioctl.c
zhangjian3032: thanks, make sense.

test/cpp.cc
zhangjian3032: got it.

src/nvme/ioctl.c
igaw: Sorry, thinko on my side. The final type is `__u32`, so we should also cast to `__u32` and not `int`.

zhangjian3032: the result of subtracting two u8* pointers is a signed integer,  use `int` also makes sense

igaw:
Thanks!

PR #945, opened 2025-01-13
examples: fix print warning on 32-bit systems

zhangjian3032:
Changed the format specifier in snprintf to use PRIu64 for compatibility with 32-bit systems.


igaw:
This looks indeed a bit funky. Not really sure what the goal is of the function but if wants to use a timestamp as dir name, we should create proper name.

PR #946, opened 2025-01-16
types: Update id-ns fields based on NVMe 2.1 spec

francispravin5:
Add Key Per I/O Status (KPIOS) and Key Per I/O Data Access Alignment and Granularity (KPIODAAG) fields of Identify Namespace Data Structure added in NVM Command Set Specification 1.1.


igaw:
Thanks!

PR #949, opened 2025-01-20
types: Update nvm-id-ns fields based on NVM Command Set Spec 1.1

francispravin5:
Add Key Per I/O Capabilities (KPIOCAP) and Write Zeroes With Deallocate Size Limit (WZDSL) fields of I/O Command Set Specific Identify Controller Data Structure in NVM Command Set Specification 1.1.


igaw:
Thanks!

PR #952, opened 2025-02-05
types,util: Added Enums for missing status codes

arbaz404:
Added enums for missing status codes based on NVM Express Base Specification 2.1.


igaw:
Thanks!

PR #953, opened 2025-02-06
types,util: add Asynchronous Event Configuration NVMe 2.1 definitions

francispravin5:
Add shift and mask for Asynchronous Event Configuration feature fields based on NVM Express Base Specification 2.1.


igaw:
Thanks!

PR #954, opened 2025-02-11
ioctl: fix nvme_get_log_reachability_groups() parameters order

ikegami-t:
Since the len parameter should be moved before the log parameter.


igaw:
Thanks!

PR #955, opened 2025-02-11
types: fix nvme_reachability_association_desc member variable name

ikegami-t:
Since the ngid variable name should be as rgid correctly.


igaw:
Thanks!

PR #956, opened 2025-02-11
fabrics: Use /etc/machine-id as a fallback source for system UUID

tbzatek:
In case when system DMI data are unavailable or bogus, fall back to reading `/etc/machine-id` before resorting to generating a random UUID.

The machine-id file typically contains persistent and unique machine UUID stable for the OS instance.

See https://www.freedesktop.org/software/systemd/man/latest/machine-id.html

Note: while this appears to be part of systemd, the file is present on my OpenRC system, likely created by elogind.

TODO:
- [ ] The specs say the value is confidential and is supposed to be hashed before used publicly. The suggested [sd_id128_get_machine_app_specific(3)](https://www.freedesktop.org/software/systemd/man/latest/sd_id128_get_machine_app_specific.html#) function just performs SHA256 hash with an arbitrary application ID value, however given that openssl use is optional and this should work in a minimal environment, we could perhaps xor the value with some const. Not sure how useful would that be though.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -44,6 +44,7 @@
 
 #define NVMF_HOSTNQN_FILE	SYSCONFDIR "/nvme/hostnqn"
 #define NVMF_HOSTID_FILE	SYSCONFDIR "/nvme/hostid"
+#define MACHINE_ID_FILE		SYSCONFDIR "/machine-id"
 
 const char *nvmf_dev = "/dev/nvme-fabrics";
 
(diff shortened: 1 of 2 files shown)


igaw:
libnvme already has a dependency to openssl, so it would be possible to use the sha algo from there, see the `_tls_key` functions. It wouldn't be too hard to get this working if openssl is present.

igaw:
Closing it, no progress. Feel free to reopen when you start to work on it again.

Issue #963, opened 2025-02-27
fallthrough define clash in included headers


tbzatek:
I believe we'll still need #506 due to naming clash in included headers. When `#define fallthrough` gets included through `config.h` (supplied as `-include libnvme/.build/internal/config.h` directly on $CC commandline by `build.ninja`) and something else uses `__attribute__((fallthrough))` in an included header, it will expand to something uncompilable.

```
In file included from /usr/include/python3.14/Python.h:63,
                 from libnvme/nvme_wrap.c:203:
/usr/include/python3.14/pyport.h:629:35: error: missing ')' after "__has_attribute"
  629 | #if _Py__has_attribute(fallthrough)
      |                                   ^
/builddir/build/BUILD/libnvme-1.11-build/libnvme-1.11/redhat-linux-build/internal/config.h:62:35: error: missing binary operator before token "("
   62 | #define fallthrough __attribute__((__fallthrough__))
```

In this case it's the swig-generated `nvme_wrap.c` code that includes `Python.h` that uses the `fallthrough` symbol:
```
#if _Py__has_attribute(fallthrough)
#  define _Py_FALLTHROUGH __attribute__((fallthrough))
#else
#  define _Py_FALLTHROUGH do { } while (0)
#endif
```

Originally reported as https://bugzilla.redhat.com/show_bug.cgi?id=2325194

igaw:
I don't like to introduce a `libnvme_fallthrough`.

The suggestion from the original report ` libnvme should undefine the "fallthrough" macro before including <Python.h>` seems a better workaround.

PR #964, opened 2025-02-27
build: do not include config.h globally

igaw:
The python build breaks because of libnvme and the swig generator define variations of fallthrough. Let's only include the config.h for the libraries.

Fixes: #963


tbzatek:
Ah, thanks!

PR #970, opened 2025-03-13
python, swig: add missing controller attributes

martin-belanger:
Found that some controller attributes (e.g. tls_key) were missing from the SWIG wrapper code. Also found that for some of the attributes, SWIG was generating code that accessed to attributes directly instead of using the getter/setter methods.

For example, SWIG would generate code like this:

  name = c->name;

Instead of that:

  name = nvme_ctrl_get_name(c);

This may have been OK in most cases, but the Python code should really use the getter/setter methods instead of accessing the controller object directly.

SWIG is a weird beast to tame. For example, all the setter/getter methods must exactly match the syntax:

  [class]_[member]_[set|get] (e.g. nvme_ctrl_traddr_get)

However, most of the libnvme functions follow this syntax:

  [class]_[set|get]_[member] (e.g. nvme_ctrl_get_traddr)

The trick to force SWIG to use the setter/getter methods is to define the attributes in a %extend block and provide explicit translation methods for all the setter/getter methods.

### DIFF
--- a/libnvme/meson.build
+++ b/libnvme/meson.build
@@ -62,9 +62,9 @@ if build_python_bindings
     # Set the PYTHONPATH so that we can run the 
     # tests directly from the build directory.
     test_env = environment()
-    test_env.append('MALLOC_PERTURB_', '0')
-    test_env.append('PYTHONPATH', join_paths(meson.current_build_dir(), '..'))
-    test_env.append('PYTHONMALLOC', 'malloc')
+    test_env.set('MALLOC_PERTURB_', '90')
+    test_env.prepend('PYTHONPATH', join_paths(meson.current_build_dir(), '..'))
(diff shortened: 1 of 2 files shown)


igaw:
Just to understand it, these `[class][member][set|get]` function wont be visible in the Python code? It's just internal stuff?

igaw:
BTW, we should keep this in mind: https://github.com/linux-nvme/libnvme/issues/206

martin-belanger:
> Just to understand it, these `[class][member][set|get]` function wont be visible in the Python code? It's just internal stuff?

Yes. It's all internal. The interface has not changed. From the Python code, everything is accessed as a property just like before, but internally the setter/getter methods will now be invoked instead of directly accessing the class members.

igaw:
Thanks!

PR #971, opened 2025-03-20
types: add performance attribute structures

ikegami-t:
These for the performance characteristics feature: 1Ch.


src/nvme/types.h
```diff
@@ -21,6 +21,8 @@
  * NVMe standard definitions
  */
 
+#define NVME_UUID_LEN 16
```
igaw, line 24: I've just fixed up the includes path directives. So here you could just include the util.h header instead.

```c
#include <nvme/util.h>
```

src/nvme/types.h
```diff
@@ -5997,6 +5999,75 @@ enum nvme_apst_entry {
...
+ * @rsvd3:	Reserved
+ * @id_list:	Performance attribute identifier list
+ * @rsvd1024:	Reserved
+ */
+struct nvme_perf_attr_id_list {
+	__u8 attrtyp;
+	__u8 msvspa;
+	__u8 usvspa;
+	__u8 rsvd3[13];
+	struct nvme_perf_attr_id id_list[1008];
```
igaw, line 6037: Isn't the range from [0xc, 0xff], so 63?

src/nvme/types.h
```diff
@@ -5997,6 +5999,75 @@ enum nvme_apst_entry {
...
+
+/**
+ * struct nvme_vs_perf_attr - Vendor specific performance attribute structure
+ * @paid:	Performance attribute identifier
+ * @rsvd16:	Reserved
+ * @attrl:	Attribute Length
+ * @vs:		Vendor specific
+ */
+struct nvme_vs_perf_attr {
+	struct nvme_perf_attr_id paid;
```
igaw, line 6049: Are you sure we want to model it this way? From the spec I would just use ` __u8 paid[NVME_UUID_LEN]` directly. 

Should `nvme_perf_attr_id` exist as it is not really directly as data struct defined in the spec? I'd like to stick as close as possible to the spec and don't come up with our own data structs. While I agree in this case here it would make sense (and it is cleaner), I think it would be better to stick to the spec. 

I can be convinced to do add it though. Thoughts? @keithbusch

src/nvme/types.h
ikegami-t: Just tried but still the error output as the struct nvme_ctrl_list used in `nvme/util.h` but the structure data itself defined in the `nvme/types.h` including the header file.
```sh
tokunori@tokunori-desktop:~/nvme-cli/subprojects/libnvme$ make -j32
meson compile -C .build
ninja: Entering directory `/home/tokunori/nvme-cli/subprojects/libnvme/.build'
[1/77] Compiling C object test/test-psk.p/psk.c.o
FAILED: test/test-psk.p/psk.c.o 
cc -Itest/test-psk.p -Itest -I../test -I. -I.. -Iccan -I../ccan -Isrc -I../src -Iinternal -I../internal -I/usr/local/include -I/usr/local/include/json-c -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=gnu99 -O2 -g -fomit-frame-pointer -D_GNU_SOURCE -MD -MQ test/test-psk.p/psk.c.o -MF test/test-psk.p/psk.c.o.d -o test/test-psk.p/psk.c.o -c ../test/psk.c
In file included from ../src/nvme/types.h:25,
                 from ../src/nvme/ioctl.h:16,
                 from ../src/nvme/linux.h:14,
                 from ../test/psk.c:7:
../src/nvme/util.h:107:33: warning: ‘struct nvme_ctrl_list’ declared inside parameter list will not be visible outside of this definition or declaration
  107 | void nvme_init_ctrl_list(struct nvme_ctrl_list *cntlist, __u16 num_ctrls,
      |                                 ^~~~~~~~~~~~~~
...
```

src/nvme/types.h
ikegami-t: Yes sorry for the error.

ikegami-t:
Fixed the review comments and rebased but the UUID changes still remained for the build error as mentioned. Thank you.

src/nvme/types.h
igaw: Oh, so it's a cyclic dependency. Hmm, maybe in this case we should indeed move it to types.h but I think we should also move the other definition (`NVME_UUID_LEN_STRING`) in this case. I'll look into it tomorrow.

Issue #980, opened 2025-04-02
`nvme sanitize-log` differences with io_uring enabled


tbzatek:
Bisected to #927 - we've started seeing first differences with libnvme-1.12 and io_uring enabled. Testing against local target and `loop` transport:

with io_uring:
```
# nvme sanitize-log /dev/nvme2
Sanitize Progress                      (SPROG) :  0
Sanitize Status                        (SSTAT) :  0
Sanitize Command Dword 10 Information (SCDW10) :  0
Estimated Time For Overwrite                   :  0
Estimated Time For Block Erase                 :  0
Estimated Time For Crypto Erase                :  0
Estimated Time For Overwrite (No-Deallocate)   :  0
Estimated Time For Block Erase (No-Deallocate) :  0
Estimated Time For Crypto Erase (No-Deallocate):  0
```

without io_uring, libnvme-1.11:

```
# nvme sanitize-log /dev/nvme2
NVMe status: Invalid Field in Command: A reserved coded value or an unsupported value in a defined field(0x6002)
```

Similar situation with `nvme self-test-log`.

igaw:
FWIW, I plan to release soon a new version. Too many issues with 2.12...

PR #981, opened 2025-04-03
ioctl: return status code when using liburing

igaw:
The status code should be propagated like it is done for the ioctl interface.

Fixes: #980


tbzatek (approved):
Thanks, from a quick testing this fixes the two issues we've been seeing.

PR #982, opened 2025-04-03
Revert various related big-endian PI commits

brandon-paupore-sndk:
This reverts a series of related changes that mapped storage and reference tags into the shared region over multiple CDWords in big-endian ordering. Keith confirmed the CDWords are always in little-endian ordering, and that the MSB-LSB representation in the spec is for the metadata payload itself.

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -1920,9 +1920,9 @@ static int nvme_set_var_size_tags(__u32 *cmd_dw2, __u32 *cmd_dw3, __u32 *cmd_dw1
 		return -1;
 	}
 
-	*cmd_dw2 = cpu_to_be32(cdw2);
-	*cmd_dw3 = cpu_to_be32(cdw3);
-	*cmd_dw14 = cpu_to_be32(cdw14);
+	*cmd_dw2 = cdw2;
+	*cmd_dw3 = cdw3;
+	*cmd_dw14 = cdw14;
 	return 0;
 }
 
--- a/src/nvme/util.c
+++ b/src/nvme/util.c
@@ -413,10 +413,8 @@ const char *nvme_status_to_string(int status, bool fabrics)
 
 static inline void nvme_init_copy_range_elbt(__u8 *elbt, __u64 eilbrt)
 {
-	int i;
-
-	for (i = 0; i < 8; i++)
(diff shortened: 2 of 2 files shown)


src/nvme/ioctl.c
```diff
@@ -1944,9 +1944,9 @@ int nvme_io(struct nvme_io_args *args, __u8 opcode)
 	cdw15 = args->apptag | (args->appmask << 16);
 
 	if (args->args_size == size_v1) {
-		cdw2 = (args->storage_tag >> 32) & 0xffff;
-		cdw3 = args->storage_tag & 0xffffffff;
-		cdw14 = args->reftag;
+		cdw2 = cpu_to_le32((args->storage_tag >> 32) & 0xffff);
```
keithbusch, line 1947: This is still wrong, you should never have to do any byte swapping for the command dwords. The driver will take care of that if necessary. Just store the command values in native cpu format.

brandon-paupore-sndk: Good to know, will update the commit message as well.

igaw:
@keithbusch this PR also updates the `nvme_init_copy_range_elbt` function. I think I take this version (after updating the commit message slightly) if this is okay with you.

keithbusch:
The copy command's elbt field is part of the payload, so that part does need the host to do whatever encoding manipulation is required. The patch will means big vs little endian CPU's will produce different payloads, so I don't think it's right. But I am not sure exactly what the format is supposed to be here, the spec is a bit confusing on this. I'd just drop that part from the patch.

igaw:
Thanks, so that means your patch on the mailing list the likely the only fix needed. Gosh, I already thought I am too stupid to read the spec. This part of the spec is completely bonkers.

brandon-paupore-sndk:
Seems good to me, thanks for the details Keith. 

Feel free to close this, or if you prefer I could update to just have the non-payload updates here.

igaw:
Okay, I'll take the version from Keith. Thanks for everyone helping out with this one. Highly appreciated!

keithbusch:
> Gosh, I already thought I am too stupid to read the spec. This part of the spec is completely bonkers.

And we didn't even have a copy-offload talk at LSFMM this year. We're losing our touch...

PR #984, opened 2025-04-07
types: Update nvme_id_independent_id_ns fields based on NVMe 2.1 spec

francispravin5:
Update the I/O Command Set Independent Identify Namespace Data Structure fields based on NVM Express Base Specification 2.1


igaw:
Thanks!

PR #986, opened 2025-04-08
types: Fix Live Migration SEQIND Shift and Mask

DanceDance:
This change fixes Sequence Indicator (SEQIND) Shift and Mask to match the NVM Express Base Specification, Revision 2.1.

The Mask and Shift are defined relative to Mos:

```
 * @NVME_LM_SEQIND_SHIFT:			Shift amount to set Sequence Indicator (SEQIND)
 *						field relative to MOS
 * @NVME_LM_SEQIND_MASK:			Mask to set SEQIND field relative to MOS
```

Relative to Mos, the `SHIFT` should be 0 and `MASK` should be 0b11 (0x3).
Screenshot 2025-04-08 at 20 32 07


igaw:
Thanks!

PR #987, opened 2025-04-08
fabrics: retry connect command on EINTR

igaw:
When the system call fails with EINTR, retry the operation.

Fixes: #https://github.com/linux-nvme/nvme-cli/issues/2760

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -243,6 +243,20 @@ else
   conf.set('fallthrough', 'do {} while (0) /* fallthrough */')
 endif
 
+if cc.has_function('TEMP_FAILURE_RETRY', prefix : '#include <errno.h>')
+  conf.set('TFR', 'TEMP_FAILURE_RETRY')
+else
+  conf.set('TFR(exp)', '''					\
+({								\
+	long int __result = 0; 					\
+	do {							\
+		__result = (long int)(exp);			\
+	} while ((__result == -1) && (errno == EINTR));		\
+	__result;						\
+})
+''')
+endif
+
 ################################################################################
 substs = configuration_data()
 substs.set('NAME',    meson.project_name())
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -243,6 +243,20 @@ else
...
 
+if cc.has_function('TEMP_FAILURE_RETRY', prefix : '#include <errno.h>')
+  conf.set('TFR', 'TEMP_FAILURE_RETRY')
+else
+  conf.set('TFR(exp)', '''					\
+({								\
+	long int __result = 0; 					\
+	do {							\
+		__result = (long int)(exp);			\
+	} while ((__result == -1) && (errno == EINTR));		\
```
martin-belanger, line 254: Could we be stuck here for a very long time?

`EINTR` happens when a signal is received. For example, if someone types CTRL-C (`SIGINT`) to stop a process that seems stuck. With this, CTRL-C will have no effect. Just wondering if that's what we want.

igaw: Yes, that is a problem. Though I am not sure what /dev/nvme-fabrics does on SIGINT.

igaw: > > Yes, this is a problem. Though I am not sure what /dev/nvme-fabrics does on SIGINT.
> 
> Right! I remember being stuck on `/dev/nvme-fabrics` before and there is no way to get out. The kernel just doesn't allow interrupting operations on that device. So, maybe my comment does not apply here.

Indeed, though I think it should be possible to handle SIGINTs, it's just not there yet.

igaw: After a bit of refreshing my memory on signals, I don't think there is a real problem. The ctrl-c is translated to SIGINT which is then delivered to the process. The default handler for SIGINT is to abort the process. That means when the syscall returns the default handler gets executed (terminate process) and thus the loop will not be run anymore.

igaw: Removed the TFR macro from the open call, as this call shouldn't be a blocking call.

igaw:
FTR, I've tested this by adding a EINTR return into to the admin queue connect code path which fired every other time. Userspace saw the EINTR and retried. So that part seems to work as expected.

PR #1000, opened 2025-04-22
mi: Add support for multiple csi buffers

chorkin:
These are used to send up to two concurrent requests over MI to the endpoint.

### DIFF
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -522,7 +522,20 @@ int nvme_mi_submit(nvme_mi_ep_t ep, struct nvme_mi_req *req,
 	return 0;
 }
 
-static void nvme_mi_admin_init_req(struct nvme_mi_req *req,
+int nvme_mi_set_csi(nvme_mi_ep_t ep, uint8_t csi)
+{
+	uint8_t csi_bit = (csi) ? 1 : 0;
+
+	if (nvme_mi_ep_has_quirk(ep, NVME_QUIRK_CSI_1_NOT_SUPPORTED) && csi_bit)
+		return -1;
+
+	ep->csi = csi_bit;
+
+	return 0;
+}
+
+static void nvme_mi_admin_init_req(nvme_mi_ep_t ep,
+				   struct nvme_mi_req *req,
 				   struct nvme_mi_admin_req_hdr *hdr,
 				   __u16 ctrl_id, __u8 opcode)
 {
@@ -531,7 +544,8 @@ static void nvme_mi_admin_init_req(struct nvme_mi_req *req,
 
 	hdr->hdr.type = NVME_MI_MSGTYPE_NVME;
 	hdr->hdr.nmp = (NVME_MI_ROR_REQ << 7) |
-		(NVME_MI_MT_ADMIN << 3); /* we always use command slot 0 */
+		(NVME_MI_MT_ADMIN << 3) |
+		(ep->csi & 1);
 	hdr->opcode = opcode;
 	hdr->ctrl_id = cpu_to_le16(ctrl_id);
 
@@ -547,7 +561,8 @@ static void nvme_mi_admin_init_resp(struct nvme_mi_resp *resp,
 	resp->hdr_len = sizeof(*hdr);
 }
 
-static void nvme_mi_control_init_req(struct nvme_mi_req *req,
+static void nvme_mi_control_init_req(nvme_mi_ep_t ep,
+				    struct nvme_mi_req *req,
 				    struct nvme_mi_control_req *control_req,
 				    __u8 opcode, __u16 cpsp)
 {
@@ -556,7 +571,8 @@ static void nvme_mi_control_init_req(struct nvme_mi_req *req,
 
 	control_req->hdr.type = NVME_MI_MSGTYPE_NVME;
 	control_req->hdr.nmp = (NVME_MI_ROR_REQ << 7) |
-		(NVME_MI_MT_CONTROL << 3); /* we always use command slot 0 */
+		(NVME_MI_MT_CONTROL << 3) |
+		(ep->csi & 1);
 	control_req->opcode = opcode;
 	control_req->cpsp = cpu_to_le16(cpsp);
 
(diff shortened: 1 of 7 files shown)


chorkin:
I believe there is a few more places we need to update with the CSI, but this is just an first pass of the idea.

chorkin:
@jk-ozlabs what do you think of this approach?

src/nvme/mi.h
```diff
@@ -478,6 +478,8 @@ struct nvme_mi_ep;
  */
 typedef struct nvme_mi_ep * nvme_mi_ep_t;
 
+void set_csi(nvme_mi_ep_t ep, uint8_t csi);
```
chorkin, line 481: This will need a prefix

src/nvme/mi.c
```diff
@@ -929,6 +936,7 @@ static int __nvme_mi_admin_get_log(nvme_mi_ctrl_t ctrl,
 	ndw = (len >> 2) - 1;
 
 	nvme_mi_admin_init_req(&req, &req_hdr, ctrl->id, nvme_admin_get_log_page);
+	req_hdr.hdr.nmp |= (ctrl->ep->csi & 1);
```
chorkin, line 939: I believe there are multiple places in the code something similar will need to be added.

src/nvme/mi.h
chorkin, line 481: This needs docs headers

src/nvme/mi.c
```diff
@@ -505,6 +505,11 @@ int nvme_mi_submit(nvme_mi_ep_t ep, struct nvme_mi_req *req,
 	return 0;
 }
 
+void set_csi(nvme_mi_ep_t ep, uint8_t csi)
+{
+	ep->csi = csi & 0b1;
+}
```
jk-ozlabs, line 511: We're probably going to need our second quirk for this - some devices in the field do not implement multiple command-slots.

For those, we probably want this to return an error if that quirk is set for this endpoint.

src/nvme/mi.c
jk-ozlabs: Yep, should be fairly straightforward to check uses of the `nmp` field though.

src/nvme/mi.c
chorkin: I haven't really looked at the quirks, but I'll investigate.

jk-ozlabs:
> I believe there is a few more places we need to update with the CSI, but this is just an first pass of the idea.

OK, maybe mark as a draft then, so that we're not bugging @igaw too much :smile:

src/nvme/mi.c
chorkin: I've added handling for if the new quirk is set, but haven't explicitly set the quirk.

src/nvme/mi.c
chorkin: @jk-ozlabs, There's a lot of different ways the NMP field is populated in the codebase.  I'm wondering if it makes the most sense to |= this bet when we get to nvme_mi_submit().  Any concerns here?   Otherwise, there's a lot of places and also a lot of risk of bugs popping up in the future when this isn't added to new code as well.

chorkin: On that note, I see that nvme_mi_calc_req_mic() is called in submit conditionally, but unconditionally in other places.  Is this intentional?  Perhaps a redundancy?  That would be _my_ only concern in moving the CSI bitset to nvme_mi_submit() since there are cases the unconditional mic could be set and not the conditional one after the bit is changed.

jk-ozlabs: >  There's a lot of different ways the NMP field is populated in the codebase

Looking at those, it's either in one of the request-init functions (`nvme_mi_admin_init_req` or `nvme_mi_control_init_req`), or open-coded in the init of a MI command. Perhaps we would just need an init helper for MI too (`nvme_mi_mi_init_req`), to consolidate those.

For the MIC:

> Is this intentional? Perhaps a redundancy?

Looks like it's redundant. The only one we should need there is in the submit path.

But I'd suggest that we set `nmp` in the command inits, if possible, to contain the responsibility for setting up the header fields to the one place.

Do you want me to look at any of those prerequisite changes (consolidating the inits, and removing the redundant MIC settings)?

jk-ozlabs: Yes, the latter: we'd have a helper for the MI command inits (which are all currently open-coded), and do the `nmp` setup there.

That means we would just have the three locations for setting the CSI: each of the three command-type specific `_req_init` functions. If necessary, we could also add a helper (that those three helpers would themselves call) for just the common MI header too, but I'm not sure that's justified.

chorkin:
I took a stab at it.  Because the way some passthrough stuff was handled not everything uses the init functions.  Let me know if you feel strongly and I can make more changes.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

chorkin:
@jk-ozlabs , I think this is ready for full review at this point.

test/mi.c
```diff
@@ -753,6 +753,13 @@ static void test_resp_csi(nvme_mi_ep_t ep)
 
 	rc = nvme_mi_mi_read_mi_data_subsys(ep, &ss_info);
 	assert(rc != 0);
+
+	nvme_mi_set_csi(ep, 1);//Change CSI
+
+	rc = nvme_mi_mi_read_mi_data_subsys(ep, &ss_info);
+	assert(rc != 0);
```
jk-ozlabs, line 760: Should we also add a test that the CSI is received as expected (ie, in the callback)?

src/nvme/mi.c
```diff
@@ -539,7 +554,8 @@ static void nvme_mi_control_init_req(struct nvme_mi_req *req,
 
 	control_req->hdr.type = NVME_MI_MSGTYPE_NVME;
 	control_req->hdr.nmp = (NVME_MI_ROR_REQ << 7) |
-		(NVME_MI_MT_CONTROL << 3); /* we always use command slot 0 */
+		(NVME_MI_MT_CONTROL << 3)|
```
jk-ozlabs, line 557: minor: space around the `|` operator, to be consistent with the line above

(and elsewhere in the change)

jk-ozlabs (commented):
Looks good, those new helpers make things cleaner.

A couple of comments inline.

igaw:
I see the two MI PRs getting ready to be merged. When both are ready, is there an order I have to pull them or they are completely independent?

chorkin:
> I see the two MI PRs getting ready to be merged. When both are ready, is there an order I have to pull them or they are completely independent?

They are independent.

jk-ozlabs (approved):

igaw:
rebased to resolve conflict in map file.

igaw:
Thanks!

PR #1001, opened 2025-04-23
Update the release script to update linker mapping file

igaw:
It is not obvious to start a new section after a release. Add an explicit new UNRELEASED section so contributers know where to put a new symbol.

I tried to get this done in awk and I really don't have a lot of experience with it. Yes, I asked a LLM to support me with it...

### DIFF
--- a/src/libnvme-mi.map
+++ b/src/libnvme-mi.map
@@ -1,3 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+LIBNVME_UNRELEASED {
+};
+
 LIBNVME_MI_1_12 {
 	global:
 		nvme_mi_mi_xfer;
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -132,6 +132,42 @@ if [ "$build_doc" = true ]; then
     git commit -s -m "doc: Regenerate all docs for $VERSION"
 fi
 
+declare -A maps
+maps=(
+    [src/libnvme.map]=LIBNVME
+    [src/libnvme-mi.map]=LIBNVME_MI
+)
+
+lib_ver="${ver//./_}"
+
+for map_file in "${!maps[@]}"
+do
+    lib_name=${maps[$map_file]}
+
+    if [ ! -f "${map_file}" ]; then
+       continue
+    fi
+
+    lib_unreleased="${lib_name}_UNRELEASED"
+
+    # Check if UNRELEASED has symbols
+    if ! awk -v lib_unreleased="$lib_unreleased" '
+        $0 ~ "^"lib_unreleased { in_section = 1; next }
+        in_section && $0 ~ /\}/ { exit }
+        in_section && $0 !~ /^[[:space:]]*($|\/|\/\*|\*|#)/ { found = 1; exit }
+        END { exit !found }
+    ' "${map_file}"; then
+        continue
+    fi
+
+    sed -i \
+        -e "s/^${lib_unreleased}\s*{/&\n};\n\n${lib_name}_${lib_ver} {/" \
+        "$map_file"
+
+    git add "${map_file}"
+    echo "${map_file} updated."
+done
+
 # update meson.build
 sed -i -e "0,/[ \t]version: /s/\([ \t]version: \).*/\1\'$ver\',/" meson.build
 if [[ -n "$libnvme_VERSION" ]] && [[ -f subprojects/libnvme.wrap ]]; then
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -1,4 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
+LIBNVME_UNRELEASED {
+};
+
 LIBNVME_1.13 {
 	global:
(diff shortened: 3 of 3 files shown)


igaw:
If this is utterly shit, we could also just use a python script...

jk-ozlabs:
You probably don't want the release version to have an `UNRELEASED` section, right? Should those just be created (say, manually) on the first symbol change following a release?

In which case, all release.sh needs to do is `sed -e s/UNRELEASED/LIBNVME_$version`

jk-ozlabs:
.. I guess that doesn't address the "where do contributors add their changes" thing though

jk-ozlabs:
note that the library base names change between libnvme (`LIBVNME_ver`) and libnvme-mi (`LIBNVME_MI_ver`)

jk-ozlabs:
hm, and we seem to have different version syntax in these. ie:

```
LIBNVME_1.12
```

vs

```
LIBNVME_MI_1_12
```

libnvme had changed from `_` to `.` in 1.9

igaw:
My idea was always have a UNRELEASED section at the beginning of the linker script, because it happens very often  new symbols are placed somewhere. Though if we can't have the 'global' label it's kind of useless. Maybe we should just add a proper comment at the beginning and have the `sed` rename thing.

Though one think I do like at this awk script is the sorting of the new symbol names.

igaw:
> libnvme had changed from _ to . in 1.9

Argh! That's why a release script is needed

jk-ozlabs:
> My idea was always have a UNRELEASED section at the beginning of the linker script

OK, makes sense. So we want something along the lines of:

- if there is an UNRELEASED section that contains symbols, rename it to the new version, and add a new UNRELEASED section
- if there are no symbols in UNRELEASED, leave it as-is

is that correct?

igaw:
Yes that is what I tried to do with the awk script. It's not easy to read I agree.

jk-ozlabs:
Something like this?

```bash
declare -A maps
maps=(
    [src/libnvme.map]=LIBNVME
    [src/libnvme-mi.map]=LIBNVME_MI
)

lib_ver="${ver//./_}"

for map_file in "${!maps[@]}"
do
    lib_name=${maps[$map_file]}

    if [ ! -f "${map_file}" ]; then
       continue
    fi

    lib_unreleased="${lib_name}_UNRELEASED"

    # Check if UNRELEASED has symbols
    if ! awk -v lib_unreleased="$lib_unreleased" '
        $0 ~ "^"lib_unreleased { in_section = 1; next }
        in_section && $0 ~ /\}/ { exit }
        in_section && $0 !~ /^[[:space:]]*($|\/|\/\*|\*|#)/ { found = 1; exit }
        END { exit !found }
    ' "${map_file}"; then
        continue
    fi

    sed -i -e "s/${lib_unreleased}/${lib_name}_${lib_ver}/" "$map_file"

    git add "${map_file}"
    echo "${map_file} updated."
done
```

src/libnvme-mi.map
```diff
@@ -1,3 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+LIBNVME_UNRELEASED {
```
jk-ozlabs, line 2: `LIBNVME_MI_UNRELEASED`

jk-ozlabs:
Ah, plus adding the new UNRELEASED section, just a minute.

jk-ozlabs:
```bash
declare -A maps
maps=(
    [src/libnvme.map]=LIBNVME
    [src/libnvme-mi.map]=LIBNVME_MI
)

lib_ver="${ver//./_}"

for map_file in "${!maps[@]}"
do
    lib_name=${maps[$map_file]}

    if [ ! -f "${map_file}" ]; then
       continue
    fi

    lib_unreleased="${lib_name}_UNRELEASED"

    # Check if UNRELEASED has symbols
    if ! awk -v lib_unreleased="$lib_unreleased" '
        $0 ~ "^"lib_unreleased { in_section = 1; next }
        in_section && $0 ~ /\}/ { exit }
        in_section && $0 !~ /^[[:space:]]*($|\/|\/\*|\*|#)/ { found = 1; exit }
        END { exit !found }
    ' "${map_file}"; then
        continue
    fi

    sed -i \
        -e "s/^${lib_unreleased}\s*{/&\n};\n\n${lib_name}_${lib_ver} {/" \
        "$map_file"

    git add "${map_file}"
    echo "${map_file} updated."
done
```

jk-ozlabs:
(only the sed expression has changed there)

igaw:
I've gave it a quick test and it seems to do the trick! Way better than my awk hack...

igaw:
I'd like to change the author credits to you if this is okay with you.

jk-ozlabs:
No need, all I did was write a bit of sed there :)

igaw:
Okay, thanks anyway! :)

jk-ozlabs:
I think this still had the LIBNVME_UNRELEASED tag for libnvme-mi.map?

igaw:
Oops. Fix it directly. Thanks!

PR #1002, opened 2025-04-23
tree: add routine to fetch subsys serial number

martin-gpy:
Add a new routine to fetch the subsystem serial number.


igaw:
Thanks!

PR #967, opened 2025-03-04
mi: Introduce asynchronous event message handling

chorkin:
Added new functionality to mi.c and mi-mctp.c to handle AEMs.  Included new example mi-mctp-ae.c for usage.

PR #1004, opened 2025-04-24
build: trigger manual upload PyPi releases

igaw:
Sometimes the upload fails for $reasons. Have something in place to manually be able to upload PyPi releases.

#967

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -27,6 +27,12 @@ jobs:
         run: |
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
+      - name: Check out specified tag
+        if: ${{ github.event.inputs.tag != '' }}
+        run: |
+          git fetch --tags
+          git checkout ${{ github.event.inputs.tag }}
+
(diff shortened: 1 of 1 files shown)


glimchb (changes requested):
i don't think you need it
when you trigger manually., you can just specify the branch/tag and action will already run there, see

image

igaw:
I triggered a manual build for the v1.12 tag from the master branch. The checkout is from master and not v1.12:

https://github.com/linux-nvme/libnvme/actions/runs/14658644341/job/41138093843

The documentation on `workflow_dispatch` indicates that the workflow can be run on any branch, but that doesn't seem to be the case here. https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow

There must be something off in the current workflow...

igaw:
Do we just need to update the initial checkout step to consider `github.event.inputs.tag`?

igaw:
I really don't want to spend any more time on this topic. Thus I'll drop this, I really don't care if PyPi gets updates or not.

PR #1006, opened 2025-05-02
types: Change LM CDQ sz argument from u8 to u32.

DanceDance:
According to NVMe 2.1 Base Specificaiton, Controller Data Queue Size (CDQSIZE) occupies bits 31:00 of CDW12, so it should be using u32.

Screenshot 2025-05-02 at 15 31 01

Note that in other places, like nvme-cli, the correct u32 is used: 

* [nvme-cli/plugins/lm/lm-nvme.c at master · linux-nvme/nvme-cli](https://github.com/linux-nvme/nvme-cli/blob/master/plugins/lm/lm-nvme.c#L67)
* [libnvme/src/nvme/ioctl.c at master · linux-nvme/libnvme](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/ioctl.c#L2325)

### DIFF
--- a/src/nvme/api-types.h
+++ b/src/nvme/api-types.h
@@ -976,7 +976,9 @@ struct nvme_dim_args {
  *		by the controller if no error is present. For Delete CDQ, this field is the CDQID
  *		to delete.
  * @sel:	Select (SEL): This field specifies the type of management operation to perform.
- * @sz:		For Create CDQ, specifies the size of CDQ, in dwords
+ * @sz_u8:	For Create CDQ, specifies the size of CDQ, in dwords - 1 byte
+ * @rsvd1:	Reserved
+ * @sz:		For Create CDQ, specifies the size of CDQ, in dwords - 4 byte
  */
 struct nvme_lm_cdq_args {
 	__u32	*result;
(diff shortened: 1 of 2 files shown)


igaw:
The problem is that this is a stable API, which means all changes should be made with backward compatibility in mind. This change will break existing users which will update libnvme but not nvme-cli. 

The current way to handle this is to add padding (the new struct size needs to be multiple of the largest member, due to alignment constraints) and than add a new membber. In the function the arg type size is then check and the different versions used, e.g.  5ac91b716ca9 ("types: Add ns-mgmt host software specified fields").

DanceDance:
Thanks, @igaw, for the explanation. Please take a look at the latest change. I did some manual testing as well, but with this approach we would need another change to nvme_cli to use the newly created sz_u32 as well.

src/nvme/api-types.h
```diff
@@ -989,6 +991,8 @@ struct nvme_lm_cdq_args {
 	__u16	cdqid;
 	__u8	sel;
 	__u8	sz;
```
igaw, line 993: The good news is that we don't promise source code backwards compatibility. That means you could rename this one and name the new member `sz`. Then we don't have to modify nvme-cli, only recompile it.

DanceDance: That's a good point! I've updated the patch. I did test locally without modifying nvme_cli and it worked too.

igaw:
just renamed `sz_u32` to `sz`.

igaw:
Thanks!

PR #1008, opened 2025-05-02
tests: Add some more sample NBFT tables

tbzatek:
Two corner cases, useful for testing the parser itself and the network management framework.


igaw:
Thanks!

PR #1009, opened 2025-05-05
types,util: add enums for Boot Partition Write Protection Config feature

francispravin5:
Add enums for Boot Partition Write Protection Config feature (FID = 85h)


igaw:
Thanks!

PR #1014, opened 2025-05-20
Make examples build selectable

giuliobenetti:
Sometimes we don't need to build examples and increase some dependency we don't have(i.e. bits/pthreadtypes.h in musl) so let's add examples option true by default.


igaw:
Added a prefix to commit title and applied the patch directly. Thanks.

PR #1015, opened 2025-05-22
tree: handle the case when host_iface/host_traddr are set to "none"

martin-belanger:
When trying to determine if a candidate configuration matches an existing controller connection, treat the candidate's host_iface or host_traddr set to "none" as a NULL pointer.

This is to avoid using host_iface/host_traddr in the comparison with existing connections and falsely concluding that a new connection needs to be created.

This should fix issue: https://github.com/linux-nvme/nvme-cli/issues/2824


abhineet99 (approved):
Verified this on the setup where I could reproduce the error without the fix, thanks.

igaw:
Thanks!

PR #1016, opened 2025-05-23
examples: remove unnecessary include of <bits/pthreadtypes.h>

dne:
The <bits/*.h> header files are GNU libc specific, and should not be used directly from application code. This one in particular is already included by <pthread.h> on glibc.

Fixes build on musl.

See also #1014.


igaw:
Thanks!

PR #1020, opened 2025-06-01
mi: add arbitration feature get function

ikegami-t:
The set feature is prohibited for the managemnet endpoint support.


jk-ozlabs (approved):
Looks good, thanks!

igaw:
Thanks!

igaw:
The wrap file in nvme-cli needs to be updated, so it gets the libnvme changes.

PR #1021, opened 2025-06-03
Expanding test coverage for ioctl.h functions

MaisenbacherD:
This PR adds tests for log functions defined in ioctl.h and adds the missing tests for the identify functions defined in ioctl.h.


igaw:
Nice! Thanks a lot!

calebsander:
These look great! Glad the mock ioctl infrastructure is useful.

PR #1026, opened 2025-06-13
types: Fix nvme_ns_id_desc packing.

thieule742:
The namespace identifier length (nidl) field can be any length which could potentially lead to unaligned access.

Signed-off-by: Thieu Le <thieule@google.com>

### DIFF
--- a/src/nvme/types.h
+++ b/src/nvme/types.h
@@ -2955,7 +2955,7 @@ struct nvme_ns_id_desc {
 	__u8	nidl;
 	__le16	rsvd;
 	__u8	nid[];
-};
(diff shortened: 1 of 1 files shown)


thieule742:
The namespace identifiers are returned in one contiguous buffer. Even if the main buffer is aligned, if there is one odd length nvme_ns_id_desc within that buffer, it will throw off the alignment of the subsequent nvme_ns_id_desc structures. An example of this is the QEMU NVMe device which returns a CSI identifier that is of length 1.

igaw:
Thanks!

PR #1027, opened 2025-06-13
mi:     nvme_mi_mi_xfer() is overly restrictive

chorkin:
MI commands may be unaligned and also have payloads that are being sent and recieved at the same command.  One case of this is AEM Sync. This wasn't seen before because that particular path does not use nvme_mi_mi_xfer().

### DIFF
--- a/src/nvme/mi.c
+++ b/src/nvme/mi.c
@@ -1742,19 +1742,6 @@ int nvme_mi_mi_xfer(nvme_mi_ep_t ep,
 		return -1;
 	}
 
-	/* request and response lengths & offset must be aligned */
-	if ((req_data_size & 0x3) ||
-	    (*resp_data_size & 0x3)) {
-		errno = EINVAL;
-		return -1;
-	}
-
-	/* bidirectional not permitted */
-	if (req_data_size && *resp_data_size) {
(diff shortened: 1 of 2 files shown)


chorkin:
@jk-ozlabs , I realized that these checks should have been removed when we introduced the AEMs since we found out that they are not required (and actually prevent AEM flow from working right).  This is just to clean up the pass through path.

jk-ozlabs:
Looks good, but just to clarify - how does the AEM path need to use passthrough?

chorkin:
> Looks good, but just to clarify - how does the AEM path need to use passthrough?

It doesn't, but when I put the AEM SYNC path in (an MI config set command), we removed those (or didn't include these) restrictions.  I didn't think about other paths that are also technically MI commands, which is why this wasn't removed earlier.

jk-ozlabs:
Right, gotchya! Thanks for the explanation, all looks good.

jk-ozlabs (approved):
LGTM, thanks.

Also confirming that the docs for `nvme_mi_mi_xfer` do not need updating, as they do not prescribe these restrictions.

igaw:
Thanks!

PR #1028, opened 2025-06-18
tree: fix mem leaks in multiple nvme ctrl routines

martin-gpy:
Valgrind revealed memory leaks in multiple nvme ctrl routines including nvme_configure_ctrl(), nvme_ctrl_alloc() & nvme_scan_ctrl(). This was because previously allocated memory was not freed before assigning new attr values to the respective nvme_ctrl struct members in these routines. Fix the same.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -1220,6 +1220,126 @@ void nvme_ctrl_set_tls_key(nvme_ctrl_t c, const char *key)
 		c->tls_key = strdup(key);
 }
 
+void nvme_ctrl_set_addr(nvme_ctrl_t c, const char *addr)
+{
+	if (c->address) {
+		free(c->address);
+		c->address = NULL;
+	}
+	if (addr)
+		c->address = strdup(addr);
+}
+
+void nvme_ctrl_set_name(nvme_ctrl_t c, const char *name)
+{
+	if (c->name) {
+		free(c->name);
+		c->name = NULL;
+	}
+	if (name)
+		c->name = strdup(name);
+}
+
+void nvme_ctrl_set_sysfs_dir(nvme_ctrl_t c, const char *sysfs_dir)
+{
+	if (c->sysfs_dir) {
+		free(c->sysfs_dir);
+		c->sysfs_dir = NULL;
+	}
+	if (sysfs_dir)
+		c->sysfs_dir = strdup(sysfs_dir);
+}
+
+void nvme_ctrl_set_firmware(nvme_ctrl_t c)
+{
+	if (c->firmware) {
+		free(c->firmware);
+		c->firmware = NULL;
+	}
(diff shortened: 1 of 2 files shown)


src/nvme/tree.c
```diff
@@ -1220,6 +1220,126 @@ void nvme_ctrl_set_tls_key(nvme_ctrl_t c, const char *key)
 		c->tls_key = strdup(key);
 }
 
+void nvme_ctrl_set_addr(nvme_ctrl_t c, const char *addr)
+{
+	if (c->address) {
+		free(c->address);
```
martin-belanger, line 1226: `free()` is NULL-safe. So you could just call it w/o first cheking that the pointer in non-NULL.

https://man7.org/linux/man-pages/man3/free.3p.html

igaw:
`nvme_configure_ctrl` is called either in `nvme_init_ctrl` or `nvme_ctrl_alloc`. Hmm, I thought these function should only initialize the `ctrl` object once.

src/nvme/tree.h
```diff
@@ -1168,6 +1168,87 @@ const char *nvme_ctrl_get_tls_key(nvme_ctrl_t c);
  */
 void nvme_ctrl_set_tls_key(nvme_ctrl_t c, const char *key);
 
+/**
+ * nvme_ctrl_set_addr() - Set controller address
+ * @c:		Controller whose address to set
+ * @addr:	Address
+ */
+void nvme_ctrl_set_addr(nvme_ctrl_t c, const char *addr);
```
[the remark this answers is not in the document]
martin-gpy: Are you saying these pointers should be freed in nvme_configure_ctrl() & nvme_ctrl_alloc() itself?

igaw:
> So how should we proceed here? Do you want me to resubmit after addressing @martin-belanger's comment above? Or avoid exposing these new ctrl functions altogether and free the individual ctrl poiners within `nvme_configure_ctrl()` & `nvme_ctrl_alloc()` itself. Please clarify.

I haven't had enough time to look into it yet. The issue I have is that all those function have been written with the assumption that we only initialize a ctrl object once. But doesn't seem the case. Thus I want to figure out if there is a fundamental problem which we need to address. I don't want to paper over design issues.

src/nvme/tree.c
```diff
@@ -2263,7 +2382,6 @@ nvme_ctrl_t nvme_scan_ctrl(nvme_root_t r, const char *name)
 	if (!c)
 		return NULL;
 
-	path = NULL;
```
igaw, line 2266: unrelated.

igaw:
We already have `nvme_deconfigure_ctrl` which does all the freeing. Let's just use this one in `nvme_configure_ctrl` with a comment explaining why it is necessary to free stuff first.  

In `nvme_ctrl_alloc` use the `FREE_CTRL_ATTR` to stay consistent.

PR #1033, opened 2025-07-01
tree: free ctrl attributes when (re)configure ctrl

igaw:
The trl object can be reused, thus all resources should be freed first.

Suggested-by: Martin George <marting@netapp.com>

Fixes: #1028

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -2064,10 +2064,29 @@ static void nvme_read_sysfs_tls(nvme_root_t r, nvme_ctrl_t c)
 	free(key);
 }
 
-static int nvme_configure_ctrl(nvme_root_t r, nvme_ctrl_t c, const char *path,
-			       const char *name)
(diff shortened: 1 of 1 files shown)


src/nvme/tree.c
```diff
@@ -2240,22 +2253,23 @@ static nvme_ctrl_t nvme_ctrl_alloc(nvme_root_t r, nvme_subsystem_t s,
 		errno = ENODEV;
 		return NULL;
 	}
+	FREE_CTRL_ATTR(c->address);
 	c->address = addr;
 	addr = NULL;
```
martin-gpy, line 2258: Similar to assigning ```c->sysfs_dir``` to ```strdup(path)``` above, please assign ```c->address``` to ```strdup(addr)``` as well and remove the ```addr = NULL``` statement:

```
-       c->address = addr;
-       addr = NULL;
+       c->address = strdup(addr);
```

That makes it more intuitive and easier to read.

PR #1035, opened 2025-07-08
types: Add Create Queue Specific (CQS) shift and mask for LM CDQ command

NateThornton:
While the NVMe standard defines the Controller Data Queue command's CNTLID field as bits 15:00, this field is offset into CDW11 by the Create Queue Specific (CQS) field in bits 31:16, making the existing CDW format incorrect

Define the `NVME_LM_CQS_SHIFT` and `NVME_LM_CQS_MASK` fields and use the in the `nvme_lm_cdq` ioctl.


igaw:
Thanks!

PR #1038, opened 2025-07-18
tree: sanitize free() handling

martin-gpy:
As per the POSIX manual, free() is NULL-safe i.e. we could call it without first checking whether the respective pointer is non-NULL. So apply that here. And while at it, reorder some of the _cleanup_free_ items in the declaration section.

Suggested-by: Martin Belanger <martin.belanger@dell.com>

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -421,10 +421,9 @@ const char *nvme_root_get_application(nvme_root_t r)
 
 void nvme_root_set_application(nvme_root_t r, const char *a)
 {
-	if (r->application) {
-		free(r->application);
-		r->application = NULL;
-	}
+	free(r->application);
+	r->application = NULL;
+
 	if (a)
 		r->application = strdup(a);
 }
@@ -466,10 +465,9 @@ const char *nvme_host_get_hostsymname(nvme_host_t h)
 
 void nvme_host_set_hostsymname(nvme_host_t h, const char *hostsymname)
 {
-	if (h->hostsymname) {
-		free(h->hostsymname);
-		h->hostsymname = NULL;
-	}
+	free(h->hostsymname);
+	h->hostsymname = NULL;
+
 	if (hostsymname)
 		h->hostsymname = strdup(hostsymname);
 }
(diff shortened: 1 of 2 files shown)


src/nvme/fabrics.c
```diff
@@ -1841,10 +1841,8 @@ static char *unescape_uri(const char *str, int len)
 struct nvme_fabrics_uri *nvme_parse_uri(const char *str)
 {
 	struct nvme_fabrics_uri *uri;
-	_cleanup_free_ char *scheme = NULL;
-	_cleanup_free_ char *authority = NULL;
-	_cleanup_free_ char *path = NULL;
-	_cleanup_free_ char *h = NULL;
+	_cleanup_free_ char *scheme = NULL, *authority = NULL;
+	_cleanup_free_ char *path = NULL, *h = NULL;
 	const char *host;
```
igaw, line 1846: Unnecessary code churn.

src/nvme/log.c
```diff
@@ -51,8 +51,7 @@ __nvme_msg(nvme_root_t r, int level,
 		"[%s] <%s>%s ",
 		"[%s] <%s> %s: ",
 	};
-	_cleanup_free_ char *header = NULL;
-	_cleanup_free_ char *message = NULL;
+	_cleanup_free_ char *header = NULL, *message = NULL;
```
igaw, line 54: Same here.

src/nvme/tree.c
```diff
@@ -130,9 +130,7 @@ int nvme_host_get_ids(nvme_root_t r,
 		      char *hostnqn_arg, char *hostid_arg,
 		      char **hostnqn, char **hostid)
 {
-	_cleanup_free_ char *nqn = NULL;
-	_cleanup_free_ char *hid = NULL;
-	_cleanup_free_ char *hnqn = NULL;
+	_cleanup_free_ char *nqn = NULL, *hid = NULL, *hnqn = NULL;
```
igaw, line 133: again nothing changes.

src/nvme/tree.c
```diff
@@ -202,8 +200,7 @@ int nvme_host_get_ids(nvme_root_t r,
 
 nvme_host_t nvme_default_host(nvme_root_t r)
 {
-	_cleanup_free_ char *hostnqn = NULL;
-	_cleanup_free_ char *hostid = NULL;
+	_cleanup_free_ char *hostnqn = NULL, *hostid = NULL;
```
igaw, line 203: another one of code churn, nothign gained, makes the review more harder, makes backporting harder.

igaw:
Thanks!

PR #1039, opened 2025-07-23
mi: correct mi admin feature functions map file

ikegami-t:
Since incorrectly added the definitions in libnvme.map file.


igaw:
Thanks!

PR #1040, opened 2025-07-24
linux: TLS PSK derivation fixes

cleech:
I was attempting to debug connecting the Linux driver / libnvme /
ktls-utils host stack to the SPDK nvmf_tgt over TLS, and ran into some
issues.

The TLS connection fails to complete a handshake because the TLS PSKs
are different. The NVMe/TCP specified key derivation steps from the
configured interchange format, to a retained PSK and finally the TLS
PSK, is implemented incompatibly in libnvme and SPDK.  After some
investigation, I believe the SPDK implementation to be correct and am
providing a libnvme patch to match it. With libnvme modified, I see the
TLS handshake complete in tlshd.


igaw:
Thanks!

PR #1042, opened 2025-07-25
linux: fix 'TLS PSK derivation fixes'

hreinecke:
The patch 'TLS PSK derivation fixes' was a bit too eager in adding the 'length' field to the HKDF label.
The NVMe TCP spec defines

Retained PSK = HKDF-Expand-Label(PRK, “HostNQN”, NQNh, Length(Configured PSK))

which according to RFC 8446 translates to:

HKDF-Expand(PRK, HkdfLabel, Length(Configured PSK))

and

HkdfLabel {
  (u16) Length(Configured PSK)
  "tls13 HostNQN"
  NQNh
}

The previous patch would tread both the 'label' and the 'context' field as an HkdfLabel, which is wrong.

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -619,7 +619,7 @@ static DEFINE_CLEANUP_FUNC(
 #define _cleanup_evp_pkey_ctx_ __cleanup__(cleanup_evp_pkey_ctx)
 
 /*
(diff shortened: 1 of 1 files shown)


hreinecke:
Original patch had been reverted, so close.

PR #1044, opened 2025-07-29
linux:  Add functions to set and clear etdas bit

jeff-lien-sndk:
To retrieve telemetry log data area 4, the Extended Telemetry Data Area 4 Supported in the Host Behavior Support feature needs to be set.  The bit will be cleared after the log has been retrieved.  This commit will add functions to set and clear that bit.

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -2,6 +2,12 @@
 LIBNVME_UNRELEASED {
 };
 
+LIBNVME_1_15 {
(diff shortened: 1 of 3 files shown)


jeff-lien-sndk:
@igaw Do you handle libnvme maintainer duties as well?  Just wondering if there's something more needed for this PR.  There's been no activity since I created it.

Issue #1045, opened 2025-07-30
segfaults with the loop transport


tbzatek:
FYI, just spotted a bunch of segfaults coming from the fact the `loop` transport's `address` sysfs value is empty:

https://github.com/linux-nvme/libnvme/blob/6e6154a20b87a19eef019efd9b0c639b63493ccc/src/nvme/tree.c#L2232-L2233

Introduced in #1033

```
Program received signal SIGSEGV, Segmentation fault.
Downloading source file /usr/src/debug/glibc-2.41.9000-24.fc43.x86_64/string/../sysdeps/x86_64/multiarch/strlen-avx2.S
__strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:76                                                                                                                                                                                                                                                             
76		VPCMPEQ	(%rdi), %ymm0, %ymm1
(gdb) t a a bt

Thread 1 (Thread 0x7ffff7b85840 (LWP 159833) "nvme"):
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:76
#1  0x00007ffff7dc12e3 in __GI___strdup (s=s@entry=0x0) at strdup.c:41
#2  0x00007ffff7f90e93 in nvme_ctrl_alloc (r=0x5555556da6e0, s=0x5555556dc130, path=0x5555556d84f0 "/sys/class/nvme/nvme3", name=0x5555556da233 "nvme3") at ../src/nvme/tree.c:2233
#3  nvme_scan_ctrl (r=r@entry=0x5555556da6e0, name=0x5555556da233 "nvme3") at ../src/nvme/tree.c:2296
#4  0x00007ffff7f9150a in nvme_scan_topology (r=r@entry=0x5555556da6e0, f=f@entry=0x0, f_args=f_args@entry=0x0) at ../src/nvme/tree.c:298
#5  0x0000555555572e38 in list (argc=<optimized out>, argv=<optimized out>, cmd=<optimized out>, plugin=<optimized out>) at ../nvme.c:3598
#6  0x00005555555b8b91 in handle_plugin (argc=2, argv=0x7fffffffe180, plugin=0x5555556c8f00 <builtin>) at ../plugin.c:188
#7  0x000055555555b3f1 in main (argc=3, argv=0x7fffffffe178) at ../nvme.c:11068
```

PR #1046, opened 2025-07-31
tree: do not try to strdup NULL pointer

igaw:
blktests nvme/003 using the loop transport fails because nvme_ctrl_alloc tries to strdup NULL pointers (address or sysfs_dir).

Introduce a 'safe' strdup version and start this version.

Reported-by: Tomáš Bžatek <tbzatek@redhat.com>
Reported-by: Yi Zhang <yi.zhang@redhat.com>
Fixes: e64249888521 ("tree: free ctrl attributes when (re)configure ctrl")

Fixes: #1045 #https://github.com/linux-nvme/nvme-cli/issues/2885

### DIFF
--- a/src/nvme/private.h
+++ b/src/nvme/private.h
@@ -335,4 +335,11 @@ void __nvme_mi_mctp_set_ops(const struct __mi_mctp_socket_ops *newops);
 int __nvme_import_keys_from_config(nvme_host_t h, nvme_ctrl_t c,
 				   long *keyring_id, long *key_id);
 
+static inline char *xstrdup(const char *s)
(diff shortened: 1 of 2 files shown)


MaisenbacherD:
> I am running blktests tr=loop with this version. Looks good so far. @MaisenbacherD do you think it would be simple to get a nightly build with blktests setup? I think you are finished with the heavy lifting for the in tree tests. Now it would be awesome to have blktests also run with the latest nvme-cli.

Yes, this should be fairly easy to set up now.
I will post a PR for this soon. :)

PR #1049, opened 2025-08-14
libnvme: propagate the error to the caller if nvme_get_log_page() fails

maurizio-lombardi:
The nvme_get_log_page() function may return an NVMe status code on failure, leaving the errno number set to zero.
The nvme_discovery_log() function wasn't
propagating this error information to its caller; it would just return NULL without setting a meaningful errno value.

the NULL pointer is interpreted as an error by nvme-cli, it then tries to print the string associated to the errno value, resulting in a funny output:

failed to get discovery log: Success

Fix this by capturing the status code returned by
nvme_get_log_page(), converting it to an errno value using nvme_status_to_errno(), and setting errno before returning.

Additionally, the error message has been improved to include the NVMe status code and the log level has been raised from LOG_INFO to LOG_ERR.

@igaw Daniel, do you have any plan to fix the error handling in libnvme/nvme-cli ? I remember we had a small chat about this some time ago, maybe at ALPSS, I don't remember.

I think this needs to be fixed, even if it results in API breakages.
In general, I think that the usage of errno should be restricted to only those functions that directly call ioctl(); IMO libnvme users like nvme-cli should never use errno.

I also think that prototypes like this:
`struct nvmf_discovery_log *nvmf_get_discovery_wargs(struct nvme_get_discovery_args *args)`
were a mistake.

I think it would have been better if it was like this:
`int nvmf_get_discovery_wargs(struct nvme_get_discovery_args *args, struct nvme_discovery_log **log)`
So it could return an integer < 0 for errors like ENOMEM, EIO etc...
and integers > 0 for nvme commands status code
0 for success.


igaw:
Ah sorry, I only skimmed over the patch. Let me look closer now.

src/nvme/fabrics.c
```diff
@@ -1161,10 +1162,11 @@ static struct nvmf_discovery_log *nvme_discovery_log(
 		 name, retries, args->max_retries);
 	log_args.log = log;
 	log_args.len = DISCOVERY_HEADER_LEN;
-	if (nvme_get_log_page(fd, NVME_LOG_PAGE_PDU_SIZE, &log_args)) {
-		nvme_msg(r, LOG_INFO,
-			 "%s: discover try %d/%d failed, error %d\n",
-			 name, retries, args->max_retries, errno);
+	status = nvme_get_log_page(fd, NVME_LOG_PAGE_PDU_SIZE, &log_args);
+	if (status) {
+		nvme_msg(r, LOG_ERR,
```
igaw, line 1167: This is debugging code, not error.

src/nvme/fabrics.c
```diff
@@ -1134,6 +1134,7 @@ static struct nvmf_discovery_log *nvme_discovery_log(
 	nvme_root_t r = root_from_ctrl(args->c);
 	struct nvmf_discovery_log *log;
 	int retries = 0;
+	int status = 0;
```
igaw, line 1137: `err` to stay consistent with the rest and no need to initialize it.

PR #1051, opened 2025-08-19
Fixup HKDF-Expand-Label algorithm

hreinecke:
Hi all,

here's my attempt to fixup the HKDF-Expand-Label algorithm. The first patch (from Chris Leech) implements the RFC-conformant algorithm for HKDF-Expand-Label, and the second patch adds back the original implementation via a new function nvme_import_tls_key_compat()

### DIFF
--- a/src/libnvme.map
+++ b/src/libnvme.map
@@ -3,6 +3,8 @@ LIBNVME_UNRELEASED {
 	global:
 		nvme_set_etdas;
 		nvme_clear_etdas;
+		nvme_insert_tls_key_compat;
+		nvme_generate_tls_key_identity_compat;
 };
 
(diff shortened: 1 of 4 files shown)


igaw:
strange one build on mainline failed in the psk tests:

https://github.com/linux-nvme/libnvme/actions/runs/17155560911/job/48671470694

I can also reproduce it locally with
```
./scripts/build.sh fallback
```

igaw:
The fallback uses  OpenSSL-3.0.8 which is not the latest in the LTS branch but still supported in upstream.

So we either update the dependency on OpenSSL correctly or make the code backwards compatible with OpenSSL 3.0.x LTS version.

PR #1052, opened 2025-08-21
docs: add missing API types documentation

Arno500:
The api-types.h file was not included during documentation generation, thus not being available in the online doc. It was containing some important structures used as arguments for some function calls.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -9,6 +9,7 @@
 #
 
 api_files = [
+  'api-types.h',
   'fabrics.h',
   'filters.h',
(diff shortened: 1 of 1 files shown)


Arno500:
Fixed!

igaw:
Ah checkpatch will still complain, because it expects a 'Signed-off-by: [email address]'

If you don't want to add your email just add your github handle that is also okay I suppose :)

Arno500:
Woops my bad, missed the second warning :/
Should be good now!

Issue #1053, opened 2025-08-22
TLS key generation not comptiable with OpenSSL 3.0.x LTS.


igaw:
https://github.com/linux-nvme/libnvme/pull/1051 updated the PSK and it turns out that there is a compatibilty issue with OpenSSL 3.0.x LTS. With newer version of OpenSSL the test passes.

```
$ ./scripts/build.sh fallback
$ meson test -C .build-ci psk  -v
ninja: Entering directory `/home/wagi/work/libnvme-upstream/.build-ci'
ninja: no work to do.
1/1 libnvme / psk RUNNING
>>> MALLOC_PERTURB_=198 MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 LD_LIBRARY_PATH=/home/wagi/work/libnvme-upstream/.build-ci/subprojects/openssl-3.0.8:/home/wagi/work/libnvme-upstream/.build-ci/src:/home/wagi/work/libnvme-upstream/.build-ci/subprojects/json-c-0.18 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MESON_TEST_ITERATION=1 /home/wagi/work/libnvme-upstream/.build-ci/test/test-psk
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
test nvme_export_tls_key hmac 1 NVMeTLSkey-1:01:VRLbtnN9AQb2WXW3c9+wEf/DRLz0QuLdbYvEhwtdWwNf9LrZ:
test nvme_export_tls_key hmac 2 NVMeTLSkey-1:02:VRLbtnN9AQb2WXW3c9+wEf/DRLz0QuLdbYvEhwtdWwP/w0S89ELi3W2LxIcLXVsDn8kXZQ==:
test nvme_import_tls_key hmac 1 NVMeTLSkey-1:01:VRLbtnN9AQb2WXW3c9+wEf/DRLz0QuLdbYvEhwtdWwNf9LrZ:
[...]

PR #1055, opened 2025-08-26
linux: fix HKDF TLS key derivation back to OpenSSL 3.0.8

cleech:
Attempted fix for Issue #1053 

I'm not crazy about rewriting things this way, but it passes the new tests when building with OpenSSL 3.0.8 and still works on my SPDK connection interoperability test.

OpenSSL prior to 3.3.1 had an issue with EVP_PKEY_CTX_add1_hkdf_info() where it acted like a 'set1' function instead of an 'add1' as documented.  Work around that by building the entire info vector outside of the OpenSSL API and only calling this function once.

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -686,45 +686,10 @@ static DEFINE_CLEANUP_FUNC(
 	cleanup_evp_pkey_ctx, EVP_PKEY_CTX *, EVP_PKEY_CTX_free)
 #define _cleanup_evp_pkey_ctx_ __cleanup__(cleanup_evp_pkey_ctx)
 
-/*
- * hkdf_info_printf()
- *
- * Helper function to append variable length label and context to an HkdfLabel
- *
- * RFC 8446 (TLS 1.3) Section 7.1 defines the HKDF-Expand-Label function as a
- * specialization of the HKDF-Expand function (RFC 5869), where the info
- * parameter is structured as an HkdfLabel.
- *
- * An HkdfLabel structure includes two variable length vectors (label and
- * context) which must be preceded by their content length as per RFC 8446
- * Section 3.4 (and not NUL terminated as per Section 7.1). Additionally,
- * HkdfLabel.label must begin with "tls13 "
- *
- * Returns the number of bytes appended to the HKDF info buffer, or -1 on an
- * error.
- */
-__attribute__((format(printf, 2, 3)))
-static int hkdf_info_printf(EVP_PKEY_CTX *ctx, char *fmt, ...)
-{
-	_cleanup_free_ char *str;
-	uint8_t len;
-	int ret;
-	va_list myargs;
-
-	va_start(myargs, fmt);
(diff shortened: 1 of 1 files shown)


igaw:
> I'm not crazy about rewriting things this way, but it passes the new tests when building with OpenSSL 3.0.8

Me neither. I think the current code is nicer to read. Maybe we could add a fallback implementation which is only selected for older version of OpenSSL. So when <= 3.3 get retired we can just rip out the the fallback implementation.

src/nvme/linux.c
```diff
@@ -793,20 +801,20 @@ static int derive_retained_key(int hmac, const char *hostnqn,
 		errno = ENOKEY;
 		return -1;
 	}
-	if (EVP_PKEY_CTX_add1_hkdf_info(ctx,
-			(const unsigned char *)&length, 2) <= 0) {
+	*(uint16_t *)&hkdf_info[0] = length;
```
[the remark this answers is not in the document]
cleech: length is already converted here, but I can move things around to put the htons call closer to the assignment to make it clearer

cleech:
Ok, how about a slightly different take.  I think this is cleaner, and I ended up removing the hkdf_info_printf() function in favor of snprintf() calls because it just wasn't helping anymore.

I removed all the changes to the compat functions, because the entire point of them is to not change.  But that does mean that the compat tests will fail in CI.  Given that the old code is not only out-of-spec, but inconsistent across big/little endian systems and across openssl versions, I'm still not in favor of trying to maintain it.

igaw:
> I removed all the changes to the compat functions, because the entire point of them is to not change. But that does mean that the compat tests will fail in CI. Given that the old code is not only out-of-spec, but inconsistent across big/little endian systems and across openssl versions, I'm still not in favor of trying to maintain it.

Well, do we want to break existing users? And I am sure there are users out there. Would it possible to issue a warning that a identifier has the wrong format?

src/nvme/linux.c
```diff
@@ -793,16 +768,34 @@ static int derive_retained_key(int hmac, const char *hostnqn,
...
+	ret = snprintf(pos, HKDF_INFO_LABEL_MAX + 1, "%c%s",
+		       (int)strlen(hkdf_label), hkdf_label);
+	if (ret <= 0 || ret > HKDF_INFO_LABEL_MAX) {
 		errno = ENOKEY;
 		return -1;
 	}
-	if (hkdf_info_printf(ctx, "tls13 HostNQN") <= 0) {
+	pos += ret;
+
+	ret = snprintf(pos, HKDF_INFO_CONTEXT_MAX + 1, "%c%s",
```
igaw, line 789: The size argument for the second `snprintf` needs to take into account that the previous `snprintf` already used up a bit of the buffer, right?

igaw: okay, but I let you fix the static analyzer reports :)

igaw:
Thanks!

PR #1059, opened 2025-09-03
test/psk: add compat vector for OpenSSL < 3.2

hreinecke:
OpenSSL versions earlier than 3.2 have a bug where EVP_PKEY_CTX_add1_hkdf_info() will always overwrite the existing 'info' value, and thus calculate a different identity hash. This issue has been uncovered by the PSK testcases, and has always been present.
We have fixed this with eff0ffef ("linux: fix HKDF TLS key derivation back to OpenSSL 3.0.8"), but the PSK testcases will still fail. So add the resulting hash values for the 'compat' test, and select the correct test vector based on the OpenSSL version.

Fixes: https://github.com/linux-nvme/libnvme/issues/1053

### DIFF
--- a/test/psk.c
+++ b/test/psk.c
@@ -9,6 +9,8 @@
 #include <stdlib.h>
 #include <errno.h>
 
+#include <openssl/opensslv.h>
+
 #include <ccan/array_size/array_size.h>
 
 #include <libnvme.h>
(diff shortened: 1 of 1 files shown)


test/psk.c
```diff
@@ -99,6 +101,42 @@ static struct test_data_identity test_data_identity[] = {
 	  "NVMe1R02 nqn.psk-test-host nqn.psk-test-subsys QhW2+Rp6RzHlNtCslyRxMnwJ11tKKhz8JCAQpQ+XUD8f9td1VeH5h53yz2wKJG1a" },
 };
 
+/*
+ * OpenSSL prior to 3.2 has a bug where
+ * EVP_PKEY_CTX_add1_hkdf_info() will always overwrite
+ * existing 'info' string. So add the resulting 'compat'
+ * identity hash vector here to make the tests succeed.
+ */
+#ifdef OPENSSL_VERSION_MINOR <= 1
```
cleech, line 110: this doesn't compile, should be an #if

cleech:
I haven't tested at all these boundaries, but I think the conditional should be
```
#if ((OPENSSL_VERSION_MINOR == 0 && OPENSSL_VERSION_PATCH < 14) \
       || (OPENSSL_VERSION_MINOR == 1 && OPENSSL_VERSION_PATCH < 6) \
       || (OPENSSL_VERSION_MINOR == 2 && OPENSSL_VERSION_PATCH < 2) \
       || (OPENSSL_VERSION_MINOR == 3 && OPENSSL_VERSION_PATCH < 2))
```
after looking through all the openssl git release branches to find the back-ports

Issue #951, opened 2025-01-28
nvme_disconnect_ctrl followed by a lookup will crash


igaw:
As discovered in https://github.com/linux-nvme/nvme-cli/issues/2674 something is amiss when a controller is disconnected and then a lookup happens (tree iteration), this will trigger a crash in `strcmp`

```
nvme_disconnect_ctrl() 
    nvme_deconfigure_ctrl() 
lookup_nvme_ctrl() 
   strcmp(nvme_ctrl_get_name(c), name)
```

igaw:
probably not worth spending more time on this. closing it.

PR #1060, opened 2025-09-03
Rework controller reconfiguation

hreinecke:
Calling nvme_reconfigure_ctrl() requires some attributes to be cleared and other to remain constant. This patchset documents the scope of the controller attributes, and also ensures that the code does not trip over an empty 'name' attribute (which is volatile and only set after 'connect' succeeded).

Fixes: https://github.com/linux-nvme/libnvme/issues/951

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -1311,7 +1311,7 @@ nvme_path_t nvme_ctrl_next_path(nvme_ctrl_t c, nvme_path_t p)
 
 #define FREE_CTRL_ATTR(a) \
 	do { free(a); (a) = NULL; } while (0)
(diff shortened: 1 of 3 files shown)


src/nvme/tree.c
```diff
@@ -2102,16 +2102,17 @@ int nvme_init_ctrl(nvme_host_t h, nvme_ctrl_t c, int instance)
 		return ret;
 	}
 
-	ret = nvme_reconfigure_ctrl(h->r, c, path, name);
-	if (ret < 0)
-		return ret;
-
+	FREE_CTRL_ATTR(c->address);
```
igaw: Isn't `nvme_init_ctrl` called only once and thus `c->address` should be NULL at this point anyway?

PR #1061, opened 2025-09-03
Fixes for python discovery

hreinecke:
I found a rather stupid typo in nvme:discover, and to ensure we catch similar issues going forward I've also added a python discovery script.


igaw:
Shouldn't the example go into the `examples` directory then?

hreinecke:
> Shouldn't the example go into the `examples` directory then?

You are right. Fixed now.

igaw:
Thanks!

PR #1062, opened 2025-09-04
python: Don't exit immediately on discovery fail.

martin-belanger:
This is just a minor fix to the example code.

In the example, when ctrl.discover() fails, we should not invoke sys.exit() immediately, but instead allow the ctrl.disconnect() to be invoked.


hreinecke (approved):
Yeah, noticed it myself. Was about to send a PR, but you beat me to it.

igaw:
Thanks!

Issue #1064, opened 2025-09-04
Python iterators fail in empty lists


hreinecke:
The python bindings have implemented iterators for 'hosts', 'subsystems', and 'controllers':

```
import sys
from libnvme import nvme

root = nvme.root()
host = nvme.host(root)

for s in host.subsystems():
    for c in s.controllers():
        print(f'{c}')

```
but when running that on a machine with no nvme devices an error is returned:
```
Traceback (most recent call last):
  File "./empty-iterators.py", line 9, in <module>
    for s in host.subsystems():
TypeError: 'NoneType' object is not iterable

```
requiring us to use:
```
for s in host.subsystems() or []:
    for c in s.controllers() or []:
        print(f'{c}')

```
which really is a bit naff. We should implement the iterators properly in SWIG, only I have no idea how to do it.

martin-belanger:
@hreinecke - Not sure how to do it either, but I can have a stab at it.

PR #1067, opened 2025-09-04
[WIP] python: fix iterators failing on empty lists

martin-belanger:
@hreinecke - this is my attempt to fix the issue: https://github.com/linux-nvme/libnvme/issues/1064
@igaw - this is still WIP. I tested it and it works, but I want to hear back from Hannes before merging it.

Hannes, a while ago you added several `iter` objects (e.g. `host_iter`, `subsystem_iter`, etc). I was wondering if this was done to allow methods such as `subsystems()`, `controllers()`, etc. to work properly as iterators. Or was there another reason?

In any case, with this PR, `subsystems()`, `controllers()`, etc, now work as native Python iterators w/o requiring any of the `iter` objects. So, we could potentially do some clean up (i.e. remove `host_iter`, `subsystem_iter`, etc).


martin-belanger:
> Have you checked whether you can remove the **iter** elements? They should not be needed now.

I will try removing the `iter` elements and I'll run some tests. I just needed your confirmation that they were not needed for something else.

hreinecke:
No, they are not. I've removed them on my local branch and everything worked. Would've pushed them but that would have created a new PR I guess, so I didn't.

martin-belanger:
@hreinecke @igaw - Done with the changes. Removed the WIP label on this PR.

hreinecke (approved):
That looks good to me.

libnvme/nvme.i
```diff
@@ -395,6 +395,17 @@ PyObject *hostid_from_file();
...
+struct nvme_host * nvme_first_host(struct nvme_root * r);
+struct nvme_host * nvme_next_host(struct nvme_root * r, struct nvme_host * h);
+struct nvme_subsystem * nvme_first_subsystem(struct nvme_host * h);
+struct nvme_subsystem * nvme_next_subsystem(struct nvme_host * h, struct nvme_subsystem * s);
+struct nvme_ctrl * nvme_subsystem_first_ctrl(struct nvme_subsystem * s);
+struct nvme_ctrl * nvme_subsystem_next_ctrl(struct nvme_subsystem * s, struct nvme_ctrl * c);
+struct nvme_ns * nvme_subsystem_first_ns(struct nvme_subsystem * s);
+struct nvme_ns * nvme_subsystem_next_ns(struct nvme_subsystem * s, struct nvme_ns * n);
+struct nvme_ns * nvme_ctrl_first_ns(struct nvme_ctrl * c);
+struct nvme_ns * nvme_ctrl_next_ns(struct nvme_ctrl * c, struct nvme_ns * n);
```
igaw, line 344: Why are these deceleration necessary? I see we include `tree.h` right above.

PR #1068, opened 2025-09-05
test/psk: add compat vector for OpenSSL < 3.2

hreinecke:
Older OpenSSL versions have a bug where
EVP_PKEY_CTX_add1_hkdf_info() will always overwrite the existing 'info' value, and thus calculate a different identity hash. This issue has been uncovered by the PSK testcases, and has always been present.
We have fixed this with eff0ffef ("linux: fix HKDF TLS key derivation back to OpenSSL 3.0.8"), but the PSK testcases will still fail. So add the resulting hash values for the 'compat' test, and select the correct test vector based on the OpenSSL version.

### DIFF
--- a/test/psk.c
+++ b/test/psk.c
@@ -99,6 +99,41 @@ static struct test_data_identity test_data_identity[] = {
 	  "NVMe1R02 nqn.psk-test-host nqn.psk-test-subsys QhW2+Rp6RzHlNtCslyRxMnwJ11tKKhz8JCAQpQ+XUD8f9td1VeH5h53yz2wKJG1a" },
 };
 
+/*
+ * Older OpenSSL versions have a bug where
+ * EVP_PKEY_CTX_add1_hkdf_info() will always overwrite
+ * existing 'info' string. So add the resulting 'compat'
+ * identity hash vector here to make the tests succeed.
+ */
+static struct test_data_identity test_data_identity_compat_openssl_bug[] = {
+	{ { 0x55, 0x12, 0xDB, 0xB6,
+	    0x73, 0x7D, 0x01, 0x06,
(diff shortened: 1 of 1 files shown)


igaw:
Currently in the code base we have two test vectors. The old one which everyone seems to use (vendors of storage arrays etc)  and the SPDK implementation:

- `test_data_identity_compat`: Non RFC 8446 compliant, works with OpenSSL 1.1x, OpenSSL >= 3.2 
- `test_data_identity`: RFC 8446 compliant works with all OpenSSL 3.x versions

With OpenSSL version between 3.0 to roughly 3.2 libnvme will create a identifier which creates a third version no one is expecting. This is of the `EVP_PKEY_CTX_add1_hkdf_info` bug, correct?

In this case we should update the existing code which uses several `EVP_PKEY_CTX_add1_hkdf_info`, replace it with one single call of this function; the same approach we have in place for RFC 8446 compliant version.

I really don't want to have different outcomes of the identifier string depending which OpenSSL is used, this is going to be nightmare to explain to users.

hreinecke:
But even if we go with your solution, we will be incompatible with _some_ older versions, depending on the openssl version used. So the 'compat' flag will generate strings which might be incompatible with existing implementations, ie the very thing the 'compat' flag should protect against.

igaw:
My points is,  that some version of nvme-cli will create an identifier which will not work with storage vendors implementations. My understanding is that buggy versions of `EVP_PKEY_CTX_add1_hkdf_info` will never generate an usable identifier.

FWIW, it would be possible to detect buggy versions `EVP_PKEY_CTX_add1_hkdf_info` during the `meson configure` step.

PR #1069, opened 2025-09-09
Implement discovery loop in python

hreinecke:
Now that we have fixed up the python bindings we can implement the command 'discovery-all' in python, serving as an example how 'connect-all' can be implemented.


igaw:
Nice! Thanks!

PR #1071, opened 2025-09-15
tree: support apple-nvme transport

craftyguy:
Apple silicon support on Linux uses a separate platform driver called
'apple-nvme', and libnvme doesn't know about this transport. This fixes
a "No transport address for 'apple-nvme'" error when running on Apple
silicon, where libnvme was unable to enumerate nvme storage attached to
this platform.

```
$ nvme list
Node                  Generic               SN                   Model                                    Namespace  Usage                      Format           FW Rev
--------------------- --------------------- -------------------- ---------------------------------------- ---------- -------------------------- ---------------- --------
/dev/nvme0n1          /dev/ng0n1            xxxxxxxxxxxxxxxx     APPLE SSD AP2048Z                        0x1          2.00  TB /   2.00  TB      4 KiB +  0 B   532.140.
...
```

Signed-off-by: Clayton Craft <clayton@craftyguy.net>


igaw:
This change seems to be incomplete. There are a few more places where the transport is checked and it really looks like these places should also be covered. The documentation should also be updated accordingly.

I suppose it would be a good idea to collect the sysfs entries `scripts/collect-sysfs` and have a test for this transport. This would all it get it tested.

igaw:
FWIW, just reformatted the if conditions. Looks great! Thanks a lot!

PR #1073, opened 2025-09-16
linux: use EVP_PKEY_CTX_add1_hkdf_info only once in compat function

igaw:
linux: use EVP_PKEY_CTX_add1_hkdf_info only once in compat function

OpenSSL prior to 3.3.1 had an issue with EVP_PKEY_CTX_add1_hkdf_info()
where it acted like a 'set1' function instead of an 'add1' as
documented.  Work around that by building the entire info vector outside
of the OpenSSL API and only calling this function once.

This is the same workaround used in commit eff0ffef0273 ("linux: fix
HKDF TLS key derivation back to OpenSSL 3.0.8").

Signed-off-by: Daniel Wagner <wagi@kernel.org>

Fixes: #1053

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -252,6 +252,22 @@ else
   conf.set('fallthrough', 'do {} while (0) /* fallthrough */')
 endif
 
+if openssl_dep.found()
+  if openssl_dep.type_name() != 'internal'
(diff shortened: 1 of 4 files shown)


hreinecke:
Thanks for doing this!

PR #1075, opened 2025-09-26
mi-mctp: Fix compile error for test/mi-mctp.c when !HAVE_LINUX_MCTP_H

chorkin:
This PR fixes an issue when building the project on systems that don't support the <linux/mctp.h> header file.


chorkin:
@jk-ozlabs can you have a peek at this?

jk-ozlabs:
The change looks good, I'd just suggest a different naming for the new header file. The top-level `src/` directory just has the exported headers, so probably is not suitable for this compat header.

Perhaps `src/nvme/mi-mctp-compat.h` instead?

igaw:
Thanks!

Issue #1089, opened 2025-11-05
Cross-compiling is broken


bkuhls:
https://github.com/linux-nvme/libnvme/pull/1073 causes errors during configure, for example
```
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: aarch64
Host machine cpu: cortex-a75
Target machine cpu family: aarch64
Target machine cpu: cortex-a75
[...]
../output-1/build/libnvme-1.16/meson.build:258:13: ERROR: Can not run test applications in this cross environment.
```
Found by buildroot autobuilders:
https://autobuild.buildroot.net/?reason=libnvme-1.16

igaw:
Thanks for the report.

Ah, this test will only work for normal builds. While we might be able to cross compile the test, it certainly going to be difficult to run it. Given it is giving only a warning for older OpenSSL version, we could just disable it for in this situation.

igaw:
@bkuhls could you give the #1090 a test? Getting the cross compile CI build with OpenSSL configured is a bit tricky. And I don't have time for this right now. Something for later.

PR #1090, opened 2025-11-06
build: avoid breaking cross builds with OpenSSL enabled

igaw:
The OpenSSL sanity check wants to be executed, thus we should only try this if it's possible. Since this is just a warning, we can just skip it if if it is not possible.

Fixes: #1089


bkuhls:
Thanks, this PR fixes the bug.
Tested-by: Bernd Kuhls <bernd@kuhls.net>

Issue #1111, opened 2026-06-27
NBFT Security Profile descriptors are silently dropped by read_security()


flaviens:
## Summary

NBFT Security Profile descriptors appear to be exposed in the public NBFT model, and SSNS/Discovery descriptors can reference them, but the parser currently drops them. In `src/nvme/nbft.c`, `read_security()` immediately returns `-EINVAL`, and `struct nbft_info_security` / `struct libnbft_security` only carries an index plus `/* TODO add fields */`.

That means a valid NBFT table can say that an SSNS requires in-band authentication / secure channel policy, while consumers of libnvme's parsed NBFT object do not receive that policy. In the `nvme connect-all --nbft` path, the SSNS/HFI address fields are used to build the connect request, but the Security Profile descriptor is not represented as mandatory connection policy.

## Why this matters

For NVMe-oF boot, NBFT is the handoff from firmware to the runtime OS. If firmware booted a root namespace with a required security policy, the OS reconnect path needs to preserve that policy or fail closed. Today, the parsed object can preserve the connection hints while dropping the security profile. This makes it easy for higher-level boot code to treat NBFT as a root reconnect policy while silently losing the security part of that policy.

## Code points

Current source points from the audited tree:

- `src/nvme/nbft.c`: `read_security()` returns `-EINVAL` unconditionally.
- `src/nvme/nbft.h`: `struct nbft_security` defines the raw descriptor fields.
- `src/nvme/nbft.h`: `struct nbft_info_security` contains only `index` and a TODO.
- `src/nvme/nbft.c`: `read_ssns()` and `read_discovery()` look up a referenced security descriptor, but the list is empty when `read_security()` drops every descriptor.
- nvme-cli's `connect-all --nbft` path consumes SSNS/HFI fields to create controllers.

## Minimal reproducer

A synthetic NBFT with one Host, one HFI, one SSNS, and one Security Profile descriptor can be generated with:

```sh
python3 harness/nvmeof_boot_lab/scripts/make_synthetic_nbft.py \
  --out-dir /tmp/nvmboot-nbft \
  --target-ip 127.0.0.1 \
  --target-nqn nqn.2026-06.example:synthetic-root \
  --namespace-uuid aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaa0001 \
  --require-digests
```

[...]

PR #1112, opened 2026-06-30
nbft: parse Security Profile descriptors

flaviens:
This implements parsing for NBFT Security Profile descriptors instead of dropping them via the current `read_security()` `-EINVAL` stub.

Changes:
- preserve Security Profile flags, secret type, optional list objects, and optional keypath in `struct nbft_info_security`;
- keep SSNS `security` references non-null when a valid Security Profile descriptor is present;
- extend `nbft-dump` output for those fields;
- add `NBFT-synthetic-security` as a regression fixture.

This is just the parser/visibility step. As discussed in #1111, connect-side policy should still skip or reject NBFT records that reference Security Descriptors until nvme-cli/libnvme can enforce the described auth/TLS/digest policy.

Fixes/Refs: #1111


igaw:
The libnvme repo ships only the 1.x version. The libnvme v3.x code is now in the nvme-cli repo. I don't have problem merging this here into libnvme 1.x as long we have a similar fix for libnvme v3.x. But first let's get @tbzatek to review it :)

tbzatek:
I'm not sure whether we want to have this in the legacy libnvme, due to risk of it not being tested against variety of firmware implementations and the risk of regressions.

FWIW, my take on this has been open for three years now: https://github.com/timberland-sig/libnvme/pull/10
Back then it was decided not to support that due to variety of concerns by the Timberland SIG members.

flaviens:
Up to your judgement, feel free to close if preferred!

tbzatek:
> Up to your judgement, feel free to close if preferred!

Let's agree to keep it with 3.0 only.

igaw:
okay fine with me. 

Thanks @flaviens for your work! Highly appreciated!

PR #1115, opened 2026-07-15
fix: fix store uuid/nguid from sysfs

lignumqt:
In sysfs, the UUID is stored as a string with hyphens. You don't need to save the hyphens in the binary representation of the UUID, otherwise you'll get truncated output.


igaw:
Could you add your Signed-off-by tag please? Thanks

igaw:
updated commit title prefix to 'tree:'

igaw:
Thanks!

PR #1116, opened 2026-07-15
add missing sym

zhangjian3032:
Noticed a few exported symbols were missing from the public headers. 
Add a generated symbol/header consistency check following
systemd's approach to automatically verify that public headers and version scripts stay in sync.

### DIFF
--- a/scripts/generate-sym-test.py
+++ b/scripts/generate-sym-test.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import re
+import sys
+from pathlib import Path
+from typing import IO
+
+
+def process_sym_file(file: IO[str]) -> None:
+    for line in file:
+        m = re.search(r'^\s+([a-zA-Z0-9_]+);', line)
+        if m:
+            print(f'        {{ "{m[1]}", {m[1]} }},')
+
+
+def include_name(header: str) -> str:
+    path = Path(header)
+
+    if path.name.startswith("libnvme"):
+        return path.name
+
+    parts = path.parts
+    if "nvme" in parts:
+        idx = parts.index("nvme")
+        return "/".join(parts[idx:])
+
+    return path.name
+
+
+def iter_header_statements(file: IO[str]) -> list[str]:
+    text = re.sub(r'/\*.*?\*/', '', file.read(), flags=re.S)
+    statements = []
+    current = []
+    brace_depth = 0
+
+    for line in text.splitlines():
+        stripped = line.strip()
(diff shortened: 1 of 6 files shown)


igaw:
Sorry about that. We didn't have any automatic checks in place for a long time. In libnvme3 we do have some but I don't think we have what you do in the second patch. Would you mind to check what could be done in libnvme3?

https://github.com/linux-nvme/nvme-cli/tree/master/libnvme

Thanks!

zhangjian3032:
@igaw Hi,

Thanks, sorry I had only looked at the standalone libnvme repository before, and did not
notice that nvme-cli already carries the libnvme3 implementation and
some ABI consistency checks.

I took a closer look at the current libnvme3 checks. As I understand it,
libnvme3 already has two checks:

  1. __libnvme_public definitions <-> .ld version-script entries
     tools/check-public-symbols.py

  2. .ld exported symbols -> installed header declarations
     tools/check-public-headers.py
     
I tried adding the reverse check:
   installed header function prototype -> .ld exported symbol

With the current tree this reports the following symbols:

  libnvme_basename
  libnvme_get_lba_status_log
  libnvme_mi_aem_open
  libnvme_mi_ep_set_mprt_max
  libnvme_mi_first_transport_handle
  libnvme_mi_mi_config_get_async_event
  libnvme_mi_mi_config_set_async_event
  libnvme_mi_next_transport_handle
  libnvme_mi_submit_entry
  libnvme_mi_submit_exit
  libnvme_ns_head_get_sysfs_dir
  libnvme_ns_identify_descs
  libnvme_release_fds
  
I am not fully sure whether all of these should be exported ABI. For
example, libnvme_basename() is declared in an installed header, but it
does not look like something that is necessarily intended to be a public
shared-library symbol.

On the other hand, this check does seem to catch real inconsistencies.
For example, libnvme_mi_submit_entry() and libnvme_mi_submit_exit() are
still declared in mi.h, but the current implementation appears to have
moved to the per-endpoint callback API:

  libnvme_mi_ep_set_submit_entry()
  libnvme_mi_ep_set_submit_exit()

So the old submit_entry/submit_exit prototypes may now be stale header
declarations rather than missing exports.

So I can prepare a patch to check:

  installed .h prototype -> .c definition with __libnvme_public -> .ld entry

Does that direction sound reasonable?

igaw:
Thanks for looking into this. It sounds we got some more holes to fix :)

Yes, please send patches for this! Highly appreciated!

The submit_entry/submit_exit stuff has changed, now instead using weak symbols, it's possible to install callbacks on the handle. The weak symbol approach is a bit hidden and libnvme is targeting for administration tasks not throughput. That means the old declarations etc can be removed.

PR #1120, opened 2026-08-20
linux: fix nvme_fw_download_seq chunk advance for partial transfers

prabhakarpujeri:
## Summary

`nvme_fw_download_seq()` advances `args.data`, `size` and `offset` by the **full chunk size `xfer`**, while only `MIN(xfer, size)` bytes are transferred per iteration. When the image size is not a multiple of `xfer`:

1. the last (short) chunk is downloaded correctly, but
2. the subsequent update walks the state past the end of the caller's buffer and **underflows `size`** (unsigned), so the loop keeps issuing firmware-download commands - roughly 2**20 more - reading (and sending to the controller) up to 4 GB past the end of the image buffer. Depending on controller behavior this can also corrupt the firmware being updated.
3. Separately, `xfer == 0` (either passed directly, or indirectly when `nvme_fw_download` is left with nothing to send and reports success) never terminates.

## Fix

Advance `args.data`, `size` and `args.offset` by the actual transferred length (`args.data_len`) instead of `xfer`, and reject a zero `xfer` up front with `EINVAL`.

## Testing

New unit test `test_fw_download_seq` (mock-ioctl based) covers:

- an image size that is **not a multiple of `xfer`** (4112 bytes with 4096-byte chunks). With the old code this issues further commands reading past the end of the buffer, which the mock catches - the test aborts against unfixed code and passes with the fix.
- the zero-`xfer` rejection.

The rest of the test suite passes (`meson test -C .build`: 29 OK, 2 expected-fail which predate this change).

### DIFF
--- a/src/nvme/linux.c
+++ b/src/nvme/linux.c
@@ -107,15 +107,20 @@ int nvme_fw_download_seq(int fd, __u32 size, __u32 xfer, __u32 offset,
 		.result = NULL,
 	};
 
+	if (!xfer) {
+		errno = EINVAL;
+		return -1;
+	}
+
 	while (size > 0) {
 		args.data_len = MIN(xfer, size);
 		err = nvme_fw_download(&args);
 		if (err)
 			break;
 
-		args.data += xfer;
-		size -= xfer;
-		args.offset += xfer;
+		args.data += args.data_len;
+		size -= args.data_len;
+		args.offset += args.data_len;
 	}
 
 	return err;
--- a/test/ioctl/misc.c
+++ b/test/ioctl/misc.c
@@ -4,6 +4,7 @@
 
 #include "mock.h"
 #include "util.h"
+#include <errno.h>
 #include <nvme/api-types.h>
 #include <nvme/ioctl.h>
 #include <nvme/types.h>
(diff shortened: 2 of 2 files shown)


igaw:
I haven't had chance to look into your submissions for libnvme 1.x yet. Sorry about that. 

Anyway, I am sure you are aware that libnvme3 lives now in the nvme-cli repo, so in case we have the same problem there, would you mind to send patches there as well? Thanks.

prabhakarpujeri:
Thanks for the heads-up — no worries about the 1.x timing at all; I know you all are busy.

Yes, I'm aware libnvme3 lives in the nvme-cli repo now. I'll audit the same code paths there and send the applicable fixes against the nvme-cli repo as well. I'll reference the two sets of PRs to each other where they overlap so nothing gets lost.

prabhakarpujeri:
Checked nvme-cli master (libnvme 3 era, currently vendored in `libnvme/`, verified at `f51976ec`): the bug does not exist there.

- `nvme_fw_download_seq()` is a libnvme 1.x API only — `git grep fw_download_seq origin/master` finds nothing, and the vendored `ioctl.c` has no equivalent sequencer
- the bundled firmware flow lives in `plugins/fw/fw-plugin.c` (`fw_download()` -> `fw_download_single()`) and is not affected by the partial-transfer advance: its loop is `for (pos = 0; pos < fw_size; pos += cfg.xfer)` with `cfg.xfer = min(cfg.xfer, fw_size - pos)` re-clamped every iteration and the image buffer indexed directly as `fw_buf + pos`, so a short final chunk advances `pos` by exactly the transferred length and the loop terminates; no commands are issued past the end of the buffer
- a zero `xfer` cannot reach the loop either: `cfg.xfer` is normalized beforehand (0 or a non-4096 multiple is replaced by the FWUG-derived size or 4096)

So no equivalent patch is needed on the nvme-cli side.

igaw:
Understood, thanks for double checking. Let's role with this then, thanks!

PR #1121, opened 2026-08-20
fabrics: reject oversized discovery log record counts

prabhakarpujeri:
## Problem

`nvme_discovery_log()` (fabrics.c) reads `numrec` from the discovery log page header returned by an **unauthenticated** discovery controller and uses it directly to size the records buffer:

```c
numrec = le64_to_cpu(log->numrec);
...
entries_size = sizeof(*log->entries) * numrec;   /* wraps for numrec >= 2^54 */
log = __nvme_alloc(sizeof(*log) + entries_size);
```

`struct nvmf_disc_log_entry` is 1024 bytes, so `entries_size` wraps for `numrec >= 2^54`. For example with `numrec = 2^54 + 2` the allocation covers only 2 entries, but afterward the fetched log is considered to hold all `2^54 + 2` entries (`sanitize_discovery_log_entry()` in `nvmf_get_discovery_wargs()` walks `log->entries[i]` and traddr/tsas parsing reads + writes via `strchomp`), causing massive out-of-bounds heap reads and writes.

A memory-corruption primitive reachable by a malicious or buggy discovery controller.

## Fix

Reject record counts whose derived buffer size would not fit into a `size_t` (the multiplication/addition would wrap), returning `EINVAL` before allocating or issuing further Get Log Page commands. Legitimate responses are unaffected: wrap requires numrec ≈ 2^54, while the maximal real response is a few thousand entries.

## Test

`test_numrec_too_large` (mock-ioctl) mocks a header with `numrec = UINT64_MAX / sizeof(entry)` and expects `EINVAL` with no further commands issued. Aborts (SIGABRT via unconsumed mocks / OOB access) against unfixed code; passes with the patch. Full `meson test` suite green.

### DIFF
--- a/src/nvme/fabrics.c
+++ b/src/nvme/fabrics.c
@@ -1216,6 +1216,21 @@ static struct nvmf_discovery_log *nvme_discovery_log(
 		if (numrec == 0)
 			break;
 
+		/*
+		 * numrec comes from the (unauthenticated) discovery
+		 * controller and is used to compute the log page buffer
+		 * size. Reject counts so large that the size computation
+		 * would wrap, which would result in a buffer smaller than
+		 * the number of entries that get processed below.
+		 */
+		if (numrec > (SIZE_MAX - sizeof(*log)) / sizeof(*log->entries)) {
+			nvme_msg(r, LOG_ERR,
+				 "%s: invalid number of discovery log records %"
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #1125, opened 2026-08-25
nbft: reject heap objects too small for the structures they decode

prabhakarpujeri:
## Problem

Descriptors reference their payload through heap objects ({offset, length}), but the parser only checks with `in_heap()` that the region lies within the NBFT heap — it never checks that the object is large enough for what follows. Non-string objects are subsequently dereferenced as fixed-size structures:

- `read_hfi()` casts the transport-info heap object of a TCP HFI to `struct nbft_hfi_info_tcp *` (128 bytes) and copies it whole;
- `read_ssns()` reads a `struct nbft_ssns_ext_info` out of the SSNS extended-info heap object;
- `read_ssns()` passes the subsystem transport-address object to `format_ip_addr()`, which unconditionally `memcpy()`s a full 16-byte `struct in6_addr`.

A malformed table declaring e.g. a 2-byte transport info object therefore makes the parser read structure-sized data from the declared location. If the object sits near the end of the table buffer the read runs past the heap allocation entirely.

I confirmed this with valgrind using a mutated copy of the `NBFT-static-ipv4` test fixture: with the HFI transport info object shrunk to 2 bytes and its offset moved to just before the end of the buffer, `test/nbft/nbft-dump` reports

```
Invalid read of size 1
Address ... is 4 bytes after a block of size 725 alloc'd
```

i.e. the record is accepted and the parser reads TCP transport info outside the allocation.

## Fix

`__get_heap_obj()` gains a `min_len` parameter; non-string heap objects shorter than that are rejected with `-EINVAL`, which drops the offending descriptor record gracefully (record-level parse failures are already non-fatal by design). The minimum is derived from the output pointer's type with `_Generic` inside the existing `get_heap_obj()` macro, so call sites need no explicit size argument:

- `struct nbft_hfi_info_tcp **` and `struct nbft_ssns_ext_info **` require the full structure size;
- `char **` (strings) and `__u8 **` (opaque byte arrays) keep no minimum, as before.

The `(char **)` casts at the four affected call sites are removed so the generic selection sees the true pointer types, and `read_ssns()` additionally verifies the transport-address object covers a 16-byte IPv6 address before calling `format_ip_addr()` (the object's length field doesn't express the address family — the flags field does).

## Testing

- Full meson test suite passes: the existing `test/nbft` fixture tables still parse identically (no false rejections), and the `tables_bad` fixtures still fail as expected.
- Re-ran `nbft-dump` on the mutated fixture under valgrind: the malformed HFI record is now rejected and dropped (absent from the dump), the rest of the table still parses, and valgrind reports no invalid reads.
- `git stash`'d the change and rebuilt to confirm the unpatched parser both accepts the malformed record and trips the invalid read shown above.

### DIFF
--- a/src/nvme/nbft.c
+++ b/src/nvme/nbft.c
@@ -84,7 +84,7 @@ static char *trtype_to_string(__u8 transport_type)
 static int __get_heap_obj(struct nbft_header *header, const char *filename,
 			  const char *descriptorname, const char *fieldname,
 			  struct nbft_heap_obj obj, bool is_string,
-			  char **output)
+			  size_t min_len, char **output)
 {
 	if (le16_to_cpu(obj.length) == 0)
 		return -ENOENT;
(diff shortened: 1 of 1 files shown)


igaw:
For nvme-cli 3 I don't have a problem to change the min compiler dependency. Though libnvme 1.x is basically in maintenance mode thus not sure about upgrading from C99 to C11 is okay. But then C11 is out there for a while...

prabhakarpujeri:
Reworked the patch so it no longer depends on `_Generic`: the minimum length is now derived with `sizeof(**(output))` in the `get_heap_obj()` macro, which is plain C99. Since `__get_heap_obj()` already rejects empty objects before consulting the minimum, the one-byte minimum this produces for strings and plain byte arrays is a no-op, and the protection for the two structure types is unchanged. Verified that `src/nvme/nbft.c` compiles cleanly under `-std=c99 -Wall -Werror`, and the checkpatch complaints about the type-selector formatting are gone (0 errors, 0 warnings on the new diff).

Pushed as 92f35baf, so this no longer touches the C language baseline.

igaw:
Very cool! Thanks a lot!

PR #1128, opened 2026-08-25
tree: hex-decode the sysfs eui attribute into the EUI-64

prabhakarpujeri:
## Problem

`nvme_strtoeuid()` parses the namespace `eui` sysfs attribute by `memcpy()`ing its first eight characters into `ns->eui64`:

```c
static int nvme_strtoeuid(const char *str, void *res)
{
	memcpy(res, str, 8);
	return 0;
}
```

But the kernel emits the namespace EUI-64 as a 16-digit hex string (`%8ph`), so the "EUI-64" stored in the tree ends up being the ASCII of the first eight hex digits, not the EUI-64 itself. Any consumer comparing `nvme_ns_get_eui64()` against an identify-namespace EUI64 sees a mismatch.

This is the same class of mistake commit 03297d8b ("tree: fix store uuid/nguid from sysfs") fixed for the `uuid` and `nguid` attributes — the `eui` parser was left behind with the raw memcpy.

## Fix

Hex-decode the attribute into the 8 EUI-64 bytes, accepting upper- and lowercase digits, and reject malformed values with `-EINVAL`.

## Testing

- Verified the old vs. new conversion on the kernel's exact output format (`"453a1122c0ffee01\n"`):
  - old: stores `34 35 33 61 31 31 32 32` (ASCII) — wrong;
  - new: stores `45 3a 11 22 c0 ff ee 01` — correct binary EUI-64;
  - malformed input (non-hex characters) is now rejected instead of silently stored.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -2696,9 +2696,30 @@ static int nvme_strtoi(const char *str, void *res)
 	return 0;
 }
 
+static int hex_digit(char c)
+{
+	if (c >= '0' && c <= '9')
+		return c - '0';
+	if (c >= 'a' && c <= 'f')
+		return c - 'a' + 10;
+	if (c >= 'A' && c <= 'F')
+		return c - 'A' + 10;
+	return -1;
+}
+
 static int nvme_strtoeuid(const char *str, void *res)
 {
-	memcpy(res, str, 8);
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #1129, opened 2026-08-25
ioctl: honor the save parameter in nvme_set_features_host_behavior()

prabhakarpujeri:
## Problem

`nvme_set_features_host_behavior(fd, save, data)` accepts a `save` argument like every other set-features wrapper but passes a literal `false` to `nvme_set_features_data()`:

```c
return nvme_set_features_data(fd, NVME_FEAT_FID_HOST_BEHAVIOR,
	NVME_NSID_NONE, 0, false, sizeof(*data), data, NULL);
```

The host-behavior feature is therefore never saved across controller resets even when the caller asks for it, silently violating the API contract. The existing unit test even carries a comment documenting the bug (`/* nvme_set_features_host_behavior() ignores SAVE */`) instead of asserting the expected behavior.

## Fix

Pass the `save` parameter through, and update `test_set_host_behavior` to expect the SAVE bit (cdw10 bit 31) set, consistent with the other feature tests.

## Testing

- With only the test change (library unpatched): `test-features` fails with `got cdw10 0x16, expected 0x80000016` and aborts, confirming the test catches the dropped SAVE bit.
- With the fix: suite green.

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -934,7 +934,7 @@ int nvme_set_features_host_behavior(int fd, bool save,
 	struct nvme_feat_host_behavior *data)
 {
 	return nvme_set_features_data(fd, NVME_FEAT_FID_HOST_BEHAVIOR,
-		NVME_NSID_NONE, 0, false, sizeof(*data), data, NULL);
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #1130, opened 2026-08-25
ioctl: return -1 with errno intact from nvme_verify_chr()

prabhakarpujeri:
## Problem

`nvme_verify_chr()` is the shared sanity check for the reset/rescan wrappers (`nvme_subsystem_reset`, `nvme_ctrl_reset`, `nvme_ns_rescan`). Its fstat() failure path does `return errno;` — propagating a *positive* errno through the public API, breaking the library convention of returning `< 0` with errno set. All three callers do `if (ret) return ret;`, so e.g. `nvme_ctrl_reset(badfd)` returns `+EBADF` to the application instead of `-1` with `errno == EBADF`.

(Also addressed here while touching the function: it uses a `static struct stat`, so concurrent resets on two fds race writing into shared storage. There's no shared state worth keeping — make it a local.)

## Fix

- `return -1;` on fstat() failure, preserving the fstat errno (matches the ENOTBLK path's existing style);
- make the `stat` buffer an automatic.

Add `test_reset_bad_fd` to the ioctl/misc suite asserting the `-1` + `errno == EBADF` contract for all three wrappers.

## Testing

- With only the test added (library unpatched) the test aborts immediately (`got 9, errno Bad file descriptor`), confirming it catches the positive-errno path; with the fix the suite is green.

### DIFF
--- a/src/nvme/ioctl.c
+++ b/src/nvme/ioctl.c
@@ -32,11 +32,11 @@
 
 static int nvme_verify_chr(int fd)
 {
-	static struct stat nvme_stat;
+	struct stat nvme_stat;
 	int err = fstat(fd, &nvme_stat);
 
 	if (err < 0)
-		return errno;
+		return -1;
 
 	if (!S_ISCHR(nvme_stat.st_mode)) {
 		errno = ENOTBLK;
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #1132, opened 2026-08-25
base64: harden base64_decode() padding and residual-bit validation

prabhakarpujeri:
## Problem

The end-of-data validity check in `base64_decode()` reads:

```c
if (ac && ((1 << bits) - 1))
	return -EAGAIN;
```

That's `ac && mask`, not `ac & mask`. Two consequences:

1. When the input carries padding, the loop always ends with `bits == 0`, so `(1 << bits) - 1` is 0 and the check can never fire: non-canonical encodings whose padding hides non-zero bits (`"Zh=="`, `"QUJ="`) are silently accepted.
2. `'='` characters are accepted anywhere — in the middle of the data or more than two of them — by shifting an extra 6 zero bits into the accumulator per pad, so strings like `"Zm=g"`, `"=Zm9"`, `"Zg==Zg=="` decode to garbage instead of failing.

Both gaps are latent for the in-tree caller (TLS PSK import) because it feeds multiples-of-4, padded strings and verifies a digest afterwards — but the decoder itself doesn't enforce its contract.

## Fix

- `'='` may only terminate the final quantum: at most twice, and only while a partial quantum is pending (`bits >= 2`); data after padding is EINVAL.
- An unpadded partial quantum at end of input is EINVAL.
- The low `2 * pads` bits the padding stands for must be zero, so non-canonical encodings with garbage in the dropped bits are rejected.

## Testing

New unit test `test/base64.c` (plugged into meson; it compiles `base64.c` directly since the decoder is internal and not in `libnvme.map`):

- all RFC 4648 vectors (`"Zg=="` … `"Zm9vYmFy"`) still decode and re-encode to their canonical form;
- malformed variants (`"Zh=="`, `"QUJ="`, `"Z==="`, `"Zg==Zg=="`, `"=Zm9"`, `"Zm=g"`, `"Zm9vYg="`, truncated/padded-missing forms) are rejected;
- round-trip sweep of every length 0..253 bytes.

The six `expected rejection` vectors fail against the unpatched decoder and pass with the fix; full meson suite green.

### DIFF
--- a/src/nvme/base64.c
+++ b/src/nvme/base64.c
@@ -66,19 +66,24 @@ int base64_encode(const unsigned char *src, int srclen, char *dst)
 int base64_decode(const char *src, int srclen, unsigned char *dst)
 {
 	uint32_t ac = 0;
-	int i, bits = 0;
+	int i, bits = 0, pads = 0;
 	unsigned char *bp = dst;
 
 	for (i = 0; i < srclen; i++) {
-		const char *p = strchr(base64_table, src[i]);
+		const char *p;
 
 		if (src[i] == '=') {
-			ac = (ac << 6);
-			bits += 6;
-			if (bits >= 8)
-				bits -= 8;
+			if (bits < 2 || pads >= 2)
+				/* padding only terminates a 4-char quantum */
+				return -EINVAL;
+			pads++;
+			bits -= 2;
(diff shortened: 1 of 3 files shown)


igaw:
Thanks!

PR #1135, opened 2026-08-25
tree: actually free filtered-out subsystems

prabhakarpujeri:
## Problem

The scan filters drop rejected objects via the public free wrappers:

```c
nvme_free_ns(n);          /* real free */
nvme_free_ctrl(c);        /* real free */
nvme_free_subsystem(s);   /* empty SWIG stub! */
```

`nvme_filter_subsystem()` therefore never removes a subsystem rejected by the scan filter — it stays linked in the tree, visible to subsequent operations, and the subsystem (plus its now-parentless ctrls/namespaces) leaks. Only the subsystem-level path is broken; namespaces and controllers are freed properly.

## Fix

Call the internal `__nvme_free_subsystem()` — the same routine the host teardown path uses — with a forward declaration since the filter sits above the definition in tree.c.

## Testing

- Full build and meson test suite pass. No test currently supplies a non-NULL scan filter, so this path is compile- and review-validated: the change is a one-line swap to the internal deleter the sibling filters effectively use.

### DIFF
--- a/src/nvme/tree.c
+++ b/src/nvme/tree.c
@@ -36,6 +36,8 @@
 #include "log.h"
 #include "private.h"
 
+static void __nvme_free_subsystem(struct nvme_subsystem *s);
+
 /**
  * struct candidate_args - Used to look for a controller matching these parameters
  * @transport:		Transport type: loop, fc, rdma, tcp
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #1140, opened 2026-08-25
examples: bound device-supplied lengths in mi-mctp, fix ssid print

prabhakarpujeri:
## Problem

Two device-supplied length fields in `examples/mi-mctp.c` are trusted without checking them against the local buffers:

1. `do_controllers()` iterates `le16_to_cpu(ctrl_list.num)` entries of `ctrl_list.identifier[]`. `num` is 16-bit; the array holds `NVME_ID_CTRL_LIST_MAX` (2047) entries. A non-conformant device reporting more than that makes the loop read ~63 KB past the array and issues thousands of spurious `show_ctrl` MI transactions.
2. `do_security_info()` validates `n_proto` against the response's `data_len`, but `data_len` covers the whole 264-byte buffer (6-byte header + `protocols[256]`), so a full-length response can pass with `n_proto` = 257/258 → the print loop reads 1-2 bytes past `protocols[]`.

Plus a small copy-paste error in `show_ctrl()`: the "PCI subsys device" line prints `ctrl.ssvid` (vendor ID) again instead of `ctrl.ssid`.

## Fix

- Clamp the controller count to `NVME_ID_CTRL_LIST_MAX` with a warning.
- Reject `n_proto > sizeof(proto_info.protocols)` before the print loop.
- Print `ctrl.ssid` for the subsystem device ID.

## Testing

- `ninja -C build` builds the examples; full meson test suite passes (the example's MI paths need real MCTP hardware and are build/review-validated only).

### DIFF
--- a/examples/mi-mctp.c
+++ b/examples/mi-mctp.c
@@ -141,7 +141,7 @@ static int show_ctrl(nvme_mi_ep_t ep, uint16_t ctrl_id)
 	printf("    PCI vendor: %04x\n", le16_to_cpu(ctrl.vid));
 	printf("    PCI device: %04x\n", le16_to_cpu(ctrl.did));
 	printf("    PCI subsys vendor: %04x\n", le16_to_cpu(ctrl.ssvid));
-	printf("    PCI subsys device: %04x\n", le16_to_cpu(ctrl.ssvid));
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #1143, opened 2026-08-27
python: support TLS and remaining fabrics config options

sithglan:
## Summary

Extend the `nvme_fabrics_config` SWIG typemap so `ctrl.connect()` accepts `queue_size`, `fast_io_fail_tmo`, `keyring`, `tls_key`, `tls_configured_key`, `tls`, and `concat` from the Python config dict, covering every field of `nvme_fabrics_config`.

The TLS-related options (`tls`, `keyring`, `tls_key`, `tls_configured_key`) are needed by nvme-stas to support NVMe/TCP connections protected by TLS Pre-Shared Keys, passing key serials from the kernel keyring.

### DIFF
--- a/libnvme/nvme.i
+++ b/libnvme/nvme.i
@@ -154,6 +154,34 @@ PyObject *hostid_from_file();
 			temp.data_digest = PyObject_IsTrue(value) ? true : false;
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


sithglan:
nvme-stas moved to libnvme3.

PR #1144, opened 2026-09-02
mi: add nvme_mi_mi_controller_health_status_poll

drakedog2008:
Implement Controller Health Status Poll (Opcode 02h) defined in NVMe-MI 2.0 (Section 5.3 & 5.3.1).

- Add struct nvme_mi_ctrl_health_poll_args to allow comprehensive query configuration (start_ctrl_id, clear, all, inc_pci, inc_sriov_pf, inc_sriov_vf, and filter_* flags).
- Add nvme_mi_mi_controller_health_status_poll() with auto-paging multi-submit loop when num_entries > 255 (up to 255 entries per submit, advancing SCTLID = last_ctlid + 1).
- Add convenience inline helper nvme_mi_mi_controller_health_status_poll_all() for querying all controllers starting from ID 0.
- Update struct nvme_mi_ctrl_health_status (CHDS) to include the 16-bit chsc[2] (Controller Health Status Changed Flags) at offset 9 and rsvd11[5].
- Add enum nvme_mi_chsc and update enum nvme_mi_csts with TCIDA.
- Export nvme_mi_mi_controller_health_status_poll in libnvme-mi.map.
- Add unit tests in test/mi.c covering single-entry query, selective flags, auto-paging across chunk boundaries, and error handling.


drakedog2008:
Hi @jk-ozlabs, 

Can you help review this PR?

igaw:
Would you mind to rebase it, there is a conflict due to changes I merged before this one?

And could you check if the libnvme3 has also all bits? I think I added this change but better to double check. Thanks!

drakedog2008:
Hi @igaw 

I rebased the code and created a similar PR for the libnvme3

https://github.com/linux-nvme/nvme-cli/pull/4006

src/nvme/mi.h
```diff
@@ -1122,6 +1124,93 @@ int nvme_mi_mi_read_mi_data_ctrl(nvme_mi_ep_t ep, __u16 ctrl_id,
 int nvme_mi_mi_subsystem_health_status_poll(nvme_mi_ep_t ep, bool clear,
 					    struct nvme_mi_nvm_ss_health_status *nshds);
 
+/**
+ * struct nvme_mi_ctrl_health_poll_args - Arguments for nvme_mi_mi_controller_health_status_poll()
```
igaw, line 1128: overlong lines

src/nvme/types.h
```diff
@@ -7483,6 +7486,39 @@ enum nvme_mi_csts {
...
+	NVME_MI_CHSC_NSSRO	= 1 << 4,
+	NVME_MI_CHSC_CECO	= 1 << 5,
+	NVME_MI_CHSC_NAC	= 1 << 6,
+	NVME_MI_CHSC_FA		= 1 << 7,
+	NVME_MI_CHSC_CSTS	= 1 << 8,
+	NVME_MI_CHSC_CTEMP	= 1 << 9,
+	NVME_MI_CHSC_PDLU	= 1 << 10,
+	NVME_MI_CHSC_SPARE	= 1 << 11,
+	NVME_MI_CHSC_CWARN	= 1 << 12,
+	NVME_MI_CHSC_TCIDA	= 1 << 13,
```
igaw, line 7521: Should these have also `CHSCF` in the name?

igaw: well, it followes the establish pattern, but still, I do not like that we deviate from the spec and start interpreting/renaming things. It makes it so way harder to find the matching part in the spec.

src/nvme/mi.c
```diff
@@ -1951,6 +1951,117 @@ int nvme_mi_mi_subsystem_health_status_poll(nvme_mi_ep_t ep, bool clear,
...
+	if (args->filter_cwarn)
+		dw1_flags |= (1U << 4);
+	if (args->filter_spare)
+		dw1_flags |= (1U << 3);
+	if (args->filter_pdlu)
+		dw1_flags |= (1U << 2);
+	if (args->filter_ctemp)
+		dw1_flags |= (1U << 1);
+	if (args->filter_csts)
+		dw1_flags |= (1U << 0);
```
igaw, line 1998: Why are these values not added as define? 

Also, the current API exposes the NVMe types and data structs and the caller has set such things. My understanding is that `nvme_mi_ctrl_health_poll_args` is a helper to avoid exposing the low level stuff. But as I said in a different place, the types and data struct the we provide in the library should follow the standard closely. Yes, we can also provide helper function (as we do) but these are not mixed with the types.

igaw:
> Hi @igaw
> 
> I rebased the code and created a similar PR for the libnvme3

Thanks!

The v3 has been released, do you really need this for v1? I'd rather not add new features to the libnvme v1 library at this piont.

src/nvme/mi.h
drakedog2008: Thank for pointing this out. I will fix on the next version.

src/nvme/types.h
drakedog2008: Thanks for pointing this out, Daniel!

To clarify where the naming came from: it wasn't an intentional deviation or reinterpretation, but rather following a slight inconsistency in the NVMe-MI 2.0 spec itself:

* In **Figure 97** (*Controller Health Data Structure (CHDS)*), the field at bytes 10:09 is explicitly titled:  
  `Controller Health Status Changed (CHSC)`
* In **Figure 98**, the corresponding bitfield table is titled:  
  `Controller Health Status Changed Flags (CHSCF)`

Because the struct field in Figure 97 was named `CHSC`, that was used for the field and enum name. However, I completely agree with you—using **`CHSCF`** (`enum nvme_mi_chscf`, `NVME_MI_CHSCF_*`, and `chscf[2]`) is much neater, avoids ambiguity, and directly matches the flags table in Figure 98.

I will update the enum to `enum nvme_mi_chscf` / `NVME_MI_CHSCF_*` and the struct field to `chscf[2]` in the next revision.

src/nvme/mi.c
drakedog2008: Hi @igaw 

Good point on adding explicit defines for the command Dword bitfields into `types.h`. I will add `NVME_MI_CHSP_DW0_*` (`ALL`, `INCVF`, `INCPF`, `INCF`) and `NVME_MI_CHSP_DW1_*` (`CCF`, `CWARN`, `SPARE`, etc.) matching Figures 94 and 95 of the spec.

Regarding `nvme_mi_ctrl_health_poll_args`: I agree that having 9 individual boolean fields in the struct creates an unnecessary abstraction layer over the spec bitfields. At the same time, we shouldn't pass raw `cdw0` directly, because `cdw0` contains `MAXRENT` (which is chunked by the function up to 255 per submit) and `SCTLID` (which is dynamically incremented across iterations during auto-paging).

Between the following two options, which do you prefer for `libnvme-mi`?

1. **Direct parameters (no `_args` struct)**:
   ```c
   int nvme_mi_mi_controller_health_status_poll(nvme_mi_ep_t ep,
                                               __u16 start_ctrl_id,
                                               bool clear,
                                               __u32 dw0_flags,
                                               __u32 dw1_flags,
                                               struct nvme_mi_ctrl_health_status *entries,
                                               unsigned int *num_entries);
    ```
2. Keep `_args` struct, but replace the 9 booleans with `dw0_flags` and `dw1_flags`:

    ```c
    struct nvme_mi_ctrl_health_poll_args {
        __u16 start_ctrl_id;
        bool clear;
        __u32 dw0_flags; /* NVME_MI_CHSP_DW0_* bitmask */
        __u32 dw1_flags; /* NVME_MI_CHSP_DW1_* bitmask */
        struct nvme_mi_ctrl_health_status *entries;
        unsigned int *num_entries;
    };
    ```
  I'm happy to go with either approach!

drakedog2008:
> > Hi @igaw
> > I rebased the code and created a similar PR for the libnvme3
> 
> Thanks!
> 
> The v3 has been released, do you really need this for v1? I'd rather not add new features to the libnvme v1 library at this piont.

Currently our [nvme (mi) daemon](https://gbmc.googlesource.com/dbus-sensors/+/refs/heads/google-dev) is built upon the libnvme. We haven't upgraded to the libnvme3 yet and we are adding the nvme event (by health status polling) support for some NVMe-MI 1.2 devices.

src/nvme/mi.c
igaw: Personally, I would go with the no `_args` approach, because `_args` is just a second level of abstraction and whenever we want to change it we have to deal with it somehow (which we tried in libnvme1 and miserably failed :))

Thus I'd say go with the function and if we need to modify if later, we get a version 2. This is about choosing your poison.

igaw:
> Currently our [nvme (mi) daemon](https://gbmc.googlesource.com/dbus-sensors/+/refs/heads/google-dev) is built upon the libnvme. We haven't upgraded to the libnvme3 yet and we are adding the nvme event (by health status polling) support for some NVMe-MI 1.2 devices.

Alright, I don't want to make your life harder for no reason. I am fine taking this for libnvme 1. This gives you time for the migration and you don't have to deal with two things at the same time. Just don't expect that libnvme 1 will be actively maintainer for ever.

src/nvme/mi.c
drakedog2008: Update the code based on option 1 (no `_args`)

src/nvme/types.h
```diff
@@ -7462,18 +7463,21 @@ struct nvme_mi_ctrl_health_status {
 	__u8	pdlu;
 	__u8	spare;
 	__u8	cwarn;
-	__u8	rsvd9[7];
+	__u8	chscf[2];
```
jk-ozlabs, line 7466: This is a bit odd representing it as an array of `u8`, whereas the spec defines a bit-set over a 16-bit field (which you have replicated in `nvme_mi_chscf`) meaning that users of the field need to split the bitmask over the array entries.

src/nvme/mi.c
```diff
@@ -1951,6 +1951,100 @@ int nvme_mi_mi_subsystem_health_status_poll(nvme_mi_ep_t ep, bool clear,
 	return 0;
 }
 
+int nvme_mi_mi_controller_health_status_poll(
+	nvme_mi_ep_t ep, __u16 start_ctrl_id, bool clear, __u32 dw0_flags,
+	__u32 dw1_flags, struct nvme_mi_ctrl_health_status *entries,
+	unsigned int *num_entries)
+{
```
jk-ozlabs, line 1958: Minor, but some inconsistent indentation here, and a couple of sites below for similar function declarations.

src/nvme/mi.c
```diff
@@ -1951,6 +1951,100 @@ int nvme_mi_mi_subsystem_health_status_poll(nvme_mi_ep_t ep, bool clear,
...
+		memset(&resp, 0, sizeof(resp));
+		resp.hdr = &resp_hdr.hdr;
+		resp.hdr_len = sizeof(resp_hdr);
+		resp.data = &entries[total];
+		resp.data_len = chunk * sizeof(*entries);
+
+		rc = nvme_mi_submit(ep, &req, &resp);
+		if (rc) {
+			*num_entries = total;
+			return rc;
```
jk-ozlabs, line 2013: The function semantics regarding partial results (and hence a truncated `num_entries`) isn't super clear. Are you intending to always set `num_entries` (and return a partial result) including on error return values?

This would be good to clarify on the function doc.

src/nvme/mi.c
```diff
@@ -1951,6 +1951,100 @@ int nvme_mi_mi_subsystem_health_status_poll(nvme_mi_ep_t ep, bool clear,
...
+		if (chunk > 255)
+			chunk = 255;
+
+		maxrent = (__u8)(chunk - 1);
+		cdw0 = (current_sctlid & 0xffff) |
+		       ((__u32)maxrent << 16) |
+		       (dw0_flags & (NVME_MI_CHSP_DW0_ALL |
+				     NVME_MI_CHSP_DW0_INCVF |
+				     NVME_MI_CHSP_DW0_INCPF |
+				     NVME_MI_CHSP_DW0_INCF));
```
jk-ozlabs, line 1991: Minor, but might be neater to mask the `dw0_flags` and `dw1_flags` before constructing the `cdw0` / `cdw1` values, and you would no longer need to do that on every iteration.

src/nvme/mi.c
```diff
@@ -1951,6 +1951,100 @@ int nvme_mi_mi_subsystem_health_status_poll(nvme_mi_ep_t ep, bool clear,
...
+			return -1;
+		}
+
+		total += rent;
+
+		if (rent < chunk)
+			break;
+
+		if (le16_to_cpu(entries[total - 1].ctlid) == 0xffff)
+			break;
```
jk-ozlabs, line 2039: I am not clear on what this is indicating, as there is no specific mention of this case in the spec. Can you add a comment?

jk-ozlabs (changes requested):
Mostly look good, just a few comments inline.

Main concern is the caller-visible representation of chscf, but other things are pretty trivial.

